PR #7, opened 2022-02-06
Block-fetch mini-protocol

agaffney:
* add node-to-node mode to test program
* implement NtN chain-sync
* implement block-fetch
* support multiple messages in same muxer segment
* block hash calculation
* common functions for decoding blocks and block headers


cmd/go-ouroboros-network/chainsync.go
```diff
@@ -0,0 +1,225 @@
...
+	oConn                 *ouroboros.Ouroboros
+	nodeToNode            bool
+	readyForNextBlockChan chan bool
+	byronEpochBaseSlot    uint64
+	byronEpochSlot        uint64
+}
+
+var syncState chainSyncState
+
+// Intersect points (last block of previous era) for each era on testnet/mainnet
```
jwitko, line 24: Do you forsee this being a hassle to keep up to date?

cmd/go-ouroboros-network/chainsync.go
```diff
@@ -0,0 +1,225 @@
...
+	os.Exit(1)
+	return nil
+}
+
+func chainSyncDoneHandler() error {
+	return nil
+}
+
+func blockFetchStartBatchHandler() error {
+	return nil
```
jwitko, line 187: I've seen a few of these now.  Are they just placeholders for future development?

jwitko (approved):

cmd/go-ouroboros-network/chainsync.go
agaffney: The library will (purposely) blow up if you don't define a callback function and receive the associated message, but we don't really care about the message here, so the callback does nothing.

cmd/go-ouroboros-network/chainsync.go
agaffney: A little bit, but they are really just here for convenience when testing the library. I can easily update it for a new era (for testnet at least) when I'm adding support for the new era's block format, assuming that the new block format is already in use on the blockchain.

PR #31, opened 2022-03-02
Basic server support

### LINKED ISSUE #30: Implement server side of handshake
This will also involve adding some general way to specify whether a connection is acting as a client or a server.

agaffney:
* basic server test program
* implement basic support for handshake version negotation
* move common mini-protocol options to separate struct
* track mode (NtN/NtC) and role (server/client) for protocol

Fixes #30


wolf31o2 (changes requested):
Let's not create a TODO without an appropriate ticket to go with it. Create the ticket and link it in the comment.

wolf31o2 (approved):

PR #101, opened 2022-10-02
feat: make diffusion mode configurable

### LINKED ISSUE #41: Support for specifying protocol diffusion mode
We currently only support the initiator-only diffusion mode, but we should at least make it explicit rather than implicit.

We should also add checks for messages that are marked as being from the initiator (not a response) when acting as a client in initiator-only mode. The `cardano-node` will produce an error like this when receiving a message marked as from the initiator while in this mode:

```
MuxError MuxInitiatorOnly "id = MiniProtocolNum 7"
```

agaffney:
Fixes #41

### DIFF
--- a/cmd/go-ouroboros-network/main.go
+++ b/cmd/go-ouroboros-network/main.go
@@ -11,6 +11,8 @@ import (
 const (
 	TESTNET_MAGIC = 1097911063
 	MAINNET_MAGIC = 764824073
+	PREPROD_MAGIC = 1
+	PREVIEW_MAGIC = 2
 )
 
 type globalFlags struct {
@@ -22,7 +24,8 @@ type globalFlags struct {
 	networkMagic int
 	testnet      bool
 	mainnet      bool
-	syncEra      string
+	preprod      bool
+	preview      bool
 }
 
 func newGlobalFlags() *globalFlags {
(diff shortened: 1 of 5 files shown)


cmd/go-ouroboros-network/main.go
```diff
@@ -53,6 +54,8 @@ func main() {
 			f.networkMagic = TESTNET_MAGIC
 		} else if f.mainnet {
 			f.networkMagic = MAINNET_MAGIC
+		} else if f.preview {
+			f.networkMagic = PREVIEW_MAGIC
 		} else {
 			fmt.Printf("You must specify one of -testnet, -mainnet, or -network-magic\n\n")
```
wolf31o2, line 60: Update this, too

cmd/go-ouroboros-network/main.go
```diff
@@ -11,6 +11,7 @@ import (
 const (
 	TESTNET_MAGIC = 1097911063
 	MAINNET_MAGIC = 764824073
+	PREVIEW_MAGIC = 2
```
wolf31o2, line 14: Add PREPROD magic 1

wolf31o2 (changes requested):

wolf31o2 (approved):

PR #144, opened 2022-12-21
feat: close the error channel on connection close

### LINKED ISSUE #143: Close error channel on connection close
When calling `.Close()` or throwing an error which shuts down the connection, close the error channel so the consumer knows to stop listening on it.

agaffney:
Fixes #143


agaffney:
This is replaced by #147

PR #183, opened 2023-01-24
feat: cleanup protocol resources on shutdown

### LINKED ISSUE #148: Cleanup per-protocol resources when the connection shuts down

agaffney:
Fixes #148


ouroboros.go
```diff
@@ -132,20 +132,24 @@ func (o *Ouroboros) setupConnection() error {
...
+			// Break out of goroutine if muxer's error channel is closed
+			if !ok {
+				return
+			}
+			if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
+				// Return a bare io.EOF error if error is EOF/ErrUnexpectedEOF
+				o.ErrorChan <- io.EOF
+			} else {
+				// Wrap error message to denote it comes from the muxer
+				o.ErrorChan <- fmt.Errorf("muxer error: %s", err)
```
wolf31o2, line 148: It's an error to an error channel, skip error in the output.

wolf31o2 (approved):
One minor comment... feel free to ignore it, though

ouroboros.go
agaffney: That's not new code (just moved around), but I purposely prefix `muxer error: ` on there so you can actually know where the error came from instead of just getting a random I/O related error without context

Issue #257, opened 2023-05-12
Ouroboros Close don't return because deadlock


safanaj:
With the introduction of waitGroup now there are 2 deadlock during the Close method.

The first encountered is because the muxer.Stop don't return until the connection still open, the muxer has the readLoop goroutine blocked on [binary.Read](https://github.com/blinklabs-io/gouroboros/blob/ea94cd02e649dc22e2f0dd957c05149dfbbcaf2a/muxer/muxer.go#L190) waiting for readable data on the connection, to avoid to send undesired error the muxer done channel have to be closed before the connection.

The second deadlock is caused by the [mini-protocol error forwarder goroutine](https://github.com/blinklabs-io/gouroboros/blob/ea94cd02e649dc22e2f0dd957c05149dfbbcaf2a/ouroboros.go#L299) , this prevent the waitGroup to complete, so the [protoErrorChan](https://github.com/blinklabs-io/gouroboros/blob/ea94cd02e649dc22e2f0dd957c05149dfbbcaf2a/ouroboros.go#L159) has to be closed before to wait for the waitGroup.

Proposed patch:
```
diff --git a/muxer/muxer.go b/muxer/muxer.go
index 75c8f12..d18911a 100644
--- a/muxer/muxer.go
+++ b/muxer/muxer.go
@@ -46,6 +46,7 @@ const (
 type Muxer struct {
 	errorChan         chan error
 	conn              net.Conn
+	connp             *net.Conn
 	sendMutex         sync.Mutex
 	startChan         chan bool
 	doneChan          chan bool
@@ -57,9 +58,10 @@ type Muxer struct {
 }
 
 // New creates a new Muxer object and starts the read loop
-func New(conn net.Conn) *Muxer {
+func New(connp *net.Conn) *Muxer {
 	m := &Muxer{
-		conn:              conn,
+		conn:              *connp,
+		connp:             connp,
 		startChan:         make(chan bool, 1),
 		doneChan:          make(chan bool),
 		errorChan:         make(chan error, 10),
@@ -81,7 +83,7 @@ func (m *Muxer) Start() {
 }
 
 // Stop shuts down the muxer
-func (m *Muxer) Stop() {
+func (m *Muxer) Stop() error {
 	// We use a mutex to prevent this function from being called multiple times
 	// concurrently, which would cause a race condition
 	m.stopMutex.Lock()
@@ -89,11 +91,24 @@ func (m *Muxer) Stop() {
[...]

PR #258, opened 2023-05-13
fix: prevent deadlocks during Close. closes: #257

safanaj:


muxer/muxer.go
```diff
@@ -57,9 +58,10 @@ type Muxer struct {
 }
 
 // New creates a new Muxer object and starts the read loop
-func New(conn net.Conn) *Muxer {
+func New(connp *net.Conn) *Muxer {
 	m := &Muxer{
-		conn:              conn,
+		conn:              *connp,
+		connp:             connp,
```
agaffney, line 64: I don't really understand what you're trying to do here with 2 copies of the net.Conn object, one of which is a pointer. If the point is just to see if we've been provided an actual `net.Conn` object at some point, there are probably "better" ways to do that.

safanaj: the idea here is just to set the conn to nil so to prevent the Ouroboros.Close to double close it, probably the best way is just remove the net.Conn Close in the Ouroboros.Close as the "network thing" is a muxer responsability.
Basically I did this ugly thing to avoid to remove [this](https://github.com/blinklabs-io/gouroboros/blob/ea94cd02e649dc22e2f0dd957c05149dfbbcaf2a/ouroboros.go#L151)

agaffney:
This was fixed elsewhere

Issue #259, opened 2023-05-13
Double start of protocol causing panic on close 


safanaj:
Double start of protocol should be prevented or be idempotent.

A started bool in the Protocol struct can be used to skip multiple Start invocation.
In case of multiple Start the panic on the muxer stop (called by ouroboros struct Close) is caused by [this](https://github.com/blinklabs-io/gouroboros/blob/ea94cd02e649dc22e2f0dd957c05149dfbbcaf2a/protocol/protocol.go#L298), on protocol Start goroutine accessing/closing the same channel are started

agaffney:
@safanaj if you're a Discord user, we'd love for you to join us in the Blink Labs Discord (https://discord.gg/5fPRZnX4qW) to discuss your use cases, testing strategies, etc.

agaffney:
This is probably fixed by #269

PR #260, opened 2023-05-13
fix: prevent protocol multiple start

safanaj:
fixes: #259


agaffney:
This was fixed elsewhere

PR #269, opened 2023-05-17
fix: protect stop/start/close functions from being called again

### LINKED ISSUE #268: Prevent multiple calls of Start/Stop/Close functions
The various Start/Stop/Close functions can currently be called multiple times, which causes problems when channels and such are already closed. These functions should have some guard to prevent anything happening when they are called again.

NOTE: this issue is broken out from #257

agaffney:
Fixes #268


wolf31o2 (approved):

safanaj:
just for ref: https://github.com/blinklabs-io/gouroboros/issues/257#issuecomment-1551740946

PR #355, opened 2023-06-30
feat: generate JSON for arbitrary CBOR

### LINKED ISSUE #305: extend cbor.Value to store structure info of CBOR value
This structure should be marshalable to JSON for snek's use. The idea is to use this for TX metadata and possibly other arbitrary datum.

agaffney:
Fixes #305

### DIFF
--- a/cbor/value.go
+++ b/cbor/value.go
@@ -15,88 +15,94 @@
 package cbor
 
 import (
+	"encoding/hex"
+	"encoding/json"
 	"fmt"
+	"sort"
+	"strings"
 )
 
 // Helpful wrapper for parsing arbitrary CBOR data which may contain types that
 // cannot be easily represented in Go (such as maps with bytestring keys)
 type Value struct {
-	Value interface{}
+	value interface{}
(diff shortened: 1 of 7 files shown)


cbor/value.go
```diff
@@ -15,88 +15,88 @@
...
+			return v.processArray(tmpTag.Content)
+		case CborTagMap:
+			return v.processMap(tmpTag.Content)
+		default:
+			// Parse the tag value via our custom Value object to handle problem types
+			tmpValue := Value{}
+			if _, err := Decode(tmpTag.Content, &tmpValue); err != nil {
+				return err
+			}
+			if tmpTag.Number >= 121 && tmpTag.Number <= 127 {
```
wolf31o2, line 72: Document these magic numbers

wolf31o2 (changes requested):
Also, create a test case and guard against the case mentioned by the GitHub code scanner (or an issue for a later PR is fine, too)

cbor/value.go
agaffney: Comments added

wolf31o2 (approved):

PR #493, opened 2024-02-12
Shutdown finished channel

igorcrevar:
This pull request introduces a wait mechanism to accurately determine when the connection close process is completed.
Connection close is initiated as it was before with `connecion.Close()` and after `<-oConn.ShutdownFinishedCh()` we can be sure that the whole process of closing has been finished

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -60,6 +60,7 @@ type Connection struct {
 	protoErrorChan        chan error
 	handshakeFinishedChan chan interface{}
 	doneChan              chan interface{}
+	shutdownFinishedChan  chan struct{}
 	waitGroup             sync.WaitGroup
 	onceClose             sync.Once
 	sendKeepAlives        bool
@@ -93,6 +94,7 @@ func NewConnection(options ...ConnectionOptionFunc) (*Connection, error) {
 		protoErrorChan:        make(chan error, 10),
 		handshakeFinishedChan: make(chan interface{}),
 		doneChan:              make(chan interface{}),
+		shutdownFinishedChan:  make(chan struct{}),
 	}
 	// Apply provided options functions
 	for _, option := range options {
@@ -122,6 +124,10 @@ func (c *Connection) Muxer() *muxer.Muxer {
 // ErrorChan returns the channel for asynchronous errors
 func (c *Connection) ErrorChan() chan error {
 	return c.errorChan
+}
+
+func (c *Connection) ShutdownFinishedChan() <-chan struct{} {
+	return c.shutdownFinishedChan
 }
 
 // Dial will establish a connection using the specified protocol and address. It works the same as [DialTimeout],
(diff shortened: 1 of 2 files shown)


connection.go
```diff
@@ -60,6 +60,7 @@ type Connection struct {
 	protoErrorChan        chan error
 	handshakeFinishedChan chan interface{}
 	doneChan              chan interface{}
+	shutdownFinishedCh    chan struct{}
```
agaffney, line 63: For consistency with other similar signaling channels here, please name this `shutdownFinishedChan` and use `chan interface{}` (or `chan any`)

connection.go
```diff
@@ -124,6 +126,10 @@ func (c *Connection) ErrorChan() chan error {
 	return c.errorChan
 }
 
+func (c *Connection) ShutdownFinishedCh() <-chan struct{} {
```
agaffney, line 129: This should be `ShutdownFinishedChan()` for consistency with `ErrorChan()` right above it

agaffney (commented):
I'm curious what your use case is for this. We could probably make the close operation sync instead of async now that most/all of the goroutine hang issues have been addressed.

igorcrevar:
I have just saw latest changes in code-base
```
func (c *Connection) shutdown() {
	// Gracefully stop the muxer
	if c.muxer != nil {
		c.muxer.Stop()
	}
	// Wait for other goroutines to finish
	c.waitGroup.Wait()
	// Close consumer error channel to signify connection shutdown
	close(c.errorChan)
}
```

In our use case, we can now do something like 
```
if oldConn := bs.connection; oldConn != nil {
    if err := oldConn.Close(); err != nil { // close previous connection
	    bs.logger.Warn("Error while closing previous connection", "err", err)
    } else {
	    <-oldConn.ErrorChan() // error channel will be closed after connection closing is done!
    }
}
```

We required the ability to wait until the connection closing process is completed before initiating a new connection. I believe it is now safe to close this pull request. What do you think?

connection.go
igorcrevar: https://github.com/blinklabs-io/gouroboros/pull/493/commits/26d75c059ec39d53ec351a8b767c2b4d66a7b4e6

connection.go
igorcrevar: https://github.com/blinklabs-io/gouroboros/commit/26d75c059ec39d53ec351a8b767c2b4d66a7b4e6
https://github.com/blinklabs-io/gouroboros/pull/493/commits/25da92c22220431ae2dad904b09c03d08a569ae7

agaffney:
With the `errorChan` closing being the last thing being done on connection shutdown and it already serving as a shutdown signal, we probably don't need this PR's functionality

igorcrevar:
> With the `errorChan` closing being the last thing being done on connection shutdown and it already serving as a shutdown signal, we probably don't need this PR's functionality
I agree. I made this pr before those changes were included

PR #496, opened 2024-02-13
Minor fixes to gourobous command.

rakshasa:
Fixed segfault when `gouroboros server` is called without address or socket args.

Changed os.RemoveAll to os.Remove as the former is dangerous.

Added proper help output for gouroboros command.

### DIFF
--- a/cmd/gouroboros/main.go
+++ b/cmd/gouroboros/main.go
@@ -21,7 +21,7 @@ import (
 	"net"
 	"os"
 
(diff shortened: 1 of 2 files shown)


wolf31o2:
Can you update the commit message to conform to conventional commits? Something like:
```
fix: segfault in example server without address or socket args
```

rakshasa:
Fixed in new PRs.

Issue #501, opened 2024-02-15
ouroboros_mock panic on data race when reusing pre-defined conversation entries


agaffney:
In the connection manager tests, we create 3 connections in a loop and run the same mocked conversation in parallel on all of them. When the connection manager compares an inbound message to the next expected message, it sets the CBOR on the message, which can happen at the same time for multiple connections, causing a panic.

```
WARNING: DATA RACE
Write at 0x00c00010ab10 by goroutine 49:
  github.com/blinklabs-io/gouroboros/protocol.(*MessageBase).SetCbor()
      /home/agaffney/dev/blinklabs-io/gouroboros/protocol/message.go:34 +0x64
  github.com/blinklabs-io/gouroboros/protocol/keepalive.(*MsgKeepAlive).SetCbor()
      <autogenerated>:1 +0x1f
  github.com/blinklabs-io/gouroboros/internal/test/ouroboros_mock.(*Connection).processInputEntry()
      /home/agaffney/dev/blinklabs-io/gouroboros/internal/test/ouroboros_mock/connection.go:191 +0x581
  github.com/blinklabs-io/gouroboros/internal/test/ouroboros_mock.(*Connection).asyncLoop()
      /home/agaffney/dev/blinklabs-io/gouroboros/internal/test/ouroboros_mock/connection.go:133 +0x1c4
  github.com/blinklabs-io/gouroboros/internal/test/ouroboros_mock.NewConnection.func3()
      /home/agaffney/dev/blinklabs-io/gouroboros/internal/test/ouroboros_mock/connection.go:78 +0x33

Previous write at 0x00c00010ab10 by goroutine 12:
  github.com/blinklabs-io/gouroboros/protocol.(*MessageBase).SetCbor()
      /home/agaffney/dev/blinklabs-io/gouroboros/protocol/message.go:34 +0x64
  github.com/blinklabs-io/gouroboros/protocol/keepalive.(*MsgKeepAlive).SetCbor()
      <autogenerated>:1 +0x1f
  github.com/blinklabs-io/gouroboros/internal/test/ouroboros_mock.(*Connection).processInputEntry()
      /home/agaffney/dev/blinklabs-io/gouroboros/internal/test/ouroboros_mock/connection.go:191 +0x581
  github.com/blinklabs-io/gouroboros/internal/test/ouroboros_mock.(*Connection).asyncLoop()
[...]

PR #504, opened 2024-02-19
fix: changing entry conversations in mock connection manager caused r…

rakshasa:
…ace conditions

### DIFF
--- a/internal/test/ouroboros_mock/connection.go
+++ b/internal/test/ouroboros_mock/connection.go
@@ -187,9 +187,12 @@ func (c *Connection) processInputEntry(entry ConversationEntry) error {
 		if msg == nil {
 			return fmt.Errorf("received unknown message type: %d", msgType)
 		}
-		// Set CBOR for expected message to match received to make comparison easier
-		entry.InputMessage.SetCbor(msg.Cbor())
-		// Compare received message to expected message
+
+		// Compare received message to expected message, excluding the cbor content
+		//
+		// As changing the CBOR of the expected message is not thread-safe, we instead clear the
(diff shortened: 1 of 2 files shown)


protocol/message.go
```diff
@@ -29,6 +30,11 @@ type MessageBase struct {
 	MessageType uint8
 }
 
+// ClearCbor removes the original CBOR that was parsed
+func (m *MessageBase) ClearCbor() {
+	m.rawCbor = nil
+}
```
agaffney, line 36: Instead of adding a new function for this, it would be cleaner to have `SetCbor()` properly handle `nil` instead of creating an empty slice.

internal/test/ouroboros_mock/connection.go
```diff
@@ -187,16 +187,21 @@ func (c *Connection) processInputEntry(entry ConversationEntry) error {
...
+		msgCbor := msg.Cbor()
+		msg.ClearCbor()
 		if !reflect.DeepEqual(msg, entry.InputMessage) {
 			return fmt.Errorf(
 				"parsed message does not match expected value: got %#v, expected %#v",
 				msg,
 				entry.InputMessage,
 			)
 		}
+		msg.SetCbor(msgCbor)
```
agaffney, line 204: Is this needed? We don't use `msg` again in this function

agaffney:
Just to note, this PR should address #501

internal/test/ouroboros_mock/connection.go
rakshasa: removed

agaffney (approved):

PR #505, opened 2024-02-19
feat: replace mutex with state loop and channels in protocol

rakshasa:


agaffney:
This looks interesting, but how does it help? The stated purpose of this PR is to get rid of the state mutex, but was it causing a problem?

rakshasa:
The state of the protocol is conceptually separate from of the recv and send loops, and by separating it out it makes the control flow much clearer.

This makes the code easier to read and future changes to state flow easier to implement.

In addition the send loop was locking the state mutex for the full duration of the send loop block, including where it writes to the `muxerSendChan`.

agaffney:
I'm not ignoring this. I just need some time to really look this over and understand it. In the meantime, could you rebase it against current `main` for the test race fixes?

agaffney (approved):
This looks great. It took me a little while to wrap my head around what was happening in some places, but everything (now) makes sense.

I'm not a super fan of the named anonymous functions embedded in larger functions, partly for function size reasons, partly because it reminds me a bit too much of Python 😅, but it's good as it is.

agaffney:
@rakshasa thank you for your contribution!

PR #512, opened 2024-03-05
fix: fix nil pointers identified by nilaway (part 1)

wolf31o2:


bech32/bech32.go
```diff
@@ -27,7 +27,7 @@ func toBytes(chars string) ([]byte, error) {
 	for i := 0; i < len(chars); i++ {
 		index := strings.IndexByte(charset, chars[i])
 		if index < 0 {
-			return nil, ErrNonCharsetChar(chars[i])
+			return []byte{}, ErrNonCharsetChar(chars[i])
```
agaffney, line 30: This is upstream vendored code. We should probably add a comment here

cmd/common/cmdline.go
```diff
@@ -33,8 +33,14 @@ type GlobalFlags struct {
 }
 
 func NewGlobalFlags() *GlobalFlags {
+	var name string
+	if os.Args == nil {
+		name = "gouorobos"
```
agaffney, line 38: `gouroboros`

cmd/gouroboros/main.go
```diff
@@ -35,8 +35,14 @@ type globalFlags struct {
 }
 
 func newGlobalFlags() *globalFlags {
+	var name string
+	if os.Args == nil {
+		name = "gouorobos"
```
agaffney, line 40: `gouroboros`

cmd/common/cmdline.go
```diff
@@ -33,8 +33,14 @@ type GlobalFlags struct {
 }
 
 func NewGlobalFlags() *GlobalFlags {
+	var name string
+	if os.Args == nil {
```
agaffney, line 37: Is it *actually* possible for `os.Args` to be `nil`? It can technically be `nil` because it's a byte slice, but the docs are fairly definitive: `Args hold the command-line arguments, starting with the program name.` This doesn't hurt anything, though.

cmd/gouroboros/main.go
wolf31o2: It's not that I misspelled it. It's that I did it twice.

bech32/bech32.go
wolf31o2: Why is this vendored instead of being imported?

agaffney: I think it was to avoid the flood of deps from importing `https://github.com/btcsuite/btcd`

agaffney (approved):

PR #517, opened 2024-03-06
fix: nil pointers identified by nilaway (part 2)

wolf31o2:


ledger/era.go
```diff
@@ -53,7 +53,7 @@ var eras = map[uint8]Era{
 func GetEraById(eraId uint8) *Era {
 	era, ok := eras[eraId]
 	if !ok {
-		return nil
+		return &Era{}
 	}
 	return &era
 }
```
agaffney, line 59: This only returns `*Era` so that we could return `nil`. With this change we need to make sure that downstream consumers know how to check for an invalid era.

wolf31o2: I switched to returning an Era and returning an explicit EraInvalid variable, like we do with NetworkInvalid. This does put it on the user to check, but now it'll not cause a nil pointer exception.

PR #520, opened 2024-03-09
fix: handling for long addresses

### LINKED ISSUE #519: Add bounds checking for Address
When investigating how we are effected by https://github.com/IntersectMBO/cardano-ledger/issues/2729, I was surprised to find that we handled everything properly. This is because we're not checking any lengths and copying into unbounded byte slices.

We need to make sure that we're doing some sanity checks around sizes/lengths AND that we can also handle the case described in the other issue. This will probably mean silently storing the extra bytes from the address, but they will only be used when generating the bech32 representation of the address.

agaffney:
Fixes #519

### DIFF
--- a/ledger/common.go
+++ b/ledger/common.go
@@ -213,6 +213,7 @@ type Address struct {
 	networkId      uint8
 	paymentAddress []byte
 	stakingAddress []byte
+	extraData      []byte
 }
 
 // NewAddress returns an Address based on the provided bech32 address string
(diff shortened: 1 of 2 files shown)


agaffney:
I was also going to use `[AddressHashSize]byte` internally for the address parts, but that made a lot of things messier because of the conversions back and forth, and we've got explicit bounds checks in place.

wolf31o2 (approved):

PR #521, opened 2024-03-10
feat: add keepalive test

verbotenj:


internal/test/ouroboros_mock/entry.go
```diff
@@ -109,6 +110,16 @@ var ConversationEntryKeepAliveResponse = ConversationEntry{
...
 
+// ConversationEntryKeepAliveResponse is a pre-defined conversation entry for a keep-alive response
+var ConversationEntryKeepAliveResponseWrongResponse = ConversationEntry{
+	Type:       EntryTypeOutput,
+	ProtocolId: keepalive.ProtocolId,
+	IsResponse: true,
+	OutputMessages: []protocol.Message{
+		keepalive.NewMsgKeepAliveResponse(MockKeepAliveWrongCookie),
+	},
+}
```
agaffney, line 121: We should add only common conversation entries here. If they are only used in the tests for another package, they should be defined in those tests.

internal/test/ouroboros_mock/entry.go
```diff
@@ -124,6 +135,15 @@ var ConversationKeepAlive = []ConversationEntry{
...
 }
 
+// ConversationKeepAlive is a pre-defined conversation with a NtN handshake and repeated keep-alive requests
+// and responses
+var ConversationKeepAliveWrongResponse = []ConversationEntry{
+	ConversationEntryHandshakeRequestGeneric,
+	ConversationEntryHandshakeNtNResponse,
+	ConversationEntryKeepAliveRequest,
+	ConversationEntryKeepAliveResponseWrongResponse,
+}
```
agaffney, line 145: Same as above

internal/test/ouroboros_mock/entry.go
```diff
@@ -23,10 +23,11 @@ import (
...
 const (
-	MockNetworkMagic       uint32 = 999999
-	MockProtocolVersionNtC uint16 = (14 + protocol.ProtocolVersionNtCOffset)
-	MockProtocolVersionNtN uint16 = 13
-	MockKeepAliveCookie    uint16 = 999
+	MockNetworkMagic         uint32 = 999999
+	MockProtocolVersionNtC   uint16 = (14 + protocol.ProtocolVersionNtCOffset)
+	MockProtocolVersionNtN   uint16 = 13
+	MockKeepAliveCookie      uint16 = 999
+	MockKeepAliveWrongCookie uint16 = 1000
```
agaffney, line 30: Move this to keepalive tests

protocol/keepalive/serverclient_test.go
```diff
@@ -0,0 +1,151 @@
...
+	oConn, err := ouroboros.New(
+		ouroboros.WithConnection(mockConn),
+		ouroboros.WithNetworkMagic(ouroboros_mock.MockNetworkMagic),
+		ouroboros.WithNodeToNode(true),
+		ouroboros.WithKeepAlive(true),
+		ouroboros.WithKeepAliveConfig(keepalive.NewConfig(
+			keepalive.WithKeepAliveFunc(func(protocolID uint16) error {
+				t.Log("keep alive beat")
+				return nil
+			}),
```
agaffney, line 62: `keepalive.WithKeepAliveFunc` doesn't do anything for the client side

protocol/keepalive/serverclient_test.go
```diff
@@ -0,0 +1,151 @@
...
+			keepalive.WithPeriod(time.Millisecond*100),
+			// Set correct cookie
+			keepalive.WithCookie(ouroboros_mock.MockKeepAliveCookie),
+			keepalive.WithDoneFunc(func() error {
+				t.Log("keep alive done")
+				return nil
+			}),
```
agaffney, line 69: `keepalive.WithDoneFunc` doesn't do anything on the client side.

protocol/keepalive/serverclient_test.go
agaffney, line 1: There are no server tests here, so this should probably just be called `client_test.go`

protocol/keepalive/serverclient_test.go
```diff
@@ -0,0 +1,151 @@
...
+// ConversationKeepAlive is a pre-defined conversation with a NtN handshake and repeated keep-alive requests
+// and responses
+var ConversationKeepAliveWrongResponse = []ouroboros_mock.ConversationEntry{
+	ouroboros_mock.ConversationEntryHandshakeRequestGeneric,
+	ouroboros_mock.ConversationEntryHandshakeNtNResponse,
+	ouroboros_mock.ConversationEntryKeepAliveRequest,
+	ConversationEntryKeepAliveResponseWrongResponse,
+}
+
+func TestServerKeepaliveHandling(t *testing.T) {
```
agaffney, line 37: We should add a `defer` with goleak at the top of this test function to make sure all goroutines are cleaned up properly at the end

protocol/keepalive/serverclient_test.go
```diff
@@ -0,0 +1,151 @@
...
+			t.Fatal(err.Error())
+		}
+	case <-time.After(2 * time.Second):
+		t.Fatalf("did not complete within timeout")
+	}
+
+	// Close connection
+	if err := oConn.Close(); err != nil {
+		t.Fatalf("unexpected error when closing Connection object: %s", err)
+	}
```
agaffney, line 89: There should be another wait for shutdown like the above with `oConn.ErrorChan()` instead of `asyncErrChan`

protocol/keepalive/serverclient_test.go
```diff
@@ -0,0 +1,151 @@
...
+	oConn, err := ouroboros.New(
+		ouroboros.WithConnection(mockConn),
+		ouroboros.WithNetworkMagic(ouroboros_mock.MockNetworkMagic),
+		ouroboros.WithNodeToNode(true),
+		ouroboros.WithKeepAlive(true),
+		ouroboros.WithKeepAliveConfig(keepalive.NewConfig(
+			keepalive.WithKeepAliveFunc(func(protocolID uint16) error {
+				t.Log("keep alive beat")
+				return nil
+			}),
```
agaffney, line 125: `keepalive.WithKeepAliveFunc` isn't useful with a client

protocol/keepalive/serverclient_test.go
```diff
@@ -0,0 +1,151 @@
...
+			keepalive.WithPeriod(time.Millisecond*10),
+			keepalive.WithCookie(MockKeepAliveWrongCookie),
+			keepalive.WithDoneFunc(func() error {
+				t.Log("keep alive done")
+				return nil
+			}),
```
agaffney, line 131: `keepalive.WithDoneFunc` isn't useful with a client

protocol/keepalive/serverclient_test.go
```diff
@@ -0,0 +1,151 @@
...
+		if ok {
+			t.Fatal(err.Error())
+		}
+	case <-time.After(2 * time.Second):
+		t.Fatalf("did not complete within timeout")
+	}
+	// Close connection
+	if err := oConn.Close(); err != nil {
+		t.Fatalf("unexpected error when closing Connection object: %s", err)
+	}
```
agaffney, line 150: Wait for the connection shutdown here

protocol/keepalive/serverclient_test.go
```diff
@@ -0,0 +1,151 @@
...
+		t.Fatalf("did not complete within timeout")
+	}
+
+	// Close connection
+	if err := oConn.Close(); err != nil {
+		t.Fatalf("unexpected error when closing Connection object: %s", err)
+	}
+}
+
+func TestServerKeepaliveHandlingWithWrongResponse(t *testing.T) {
```
agaffney, line 92: Add a `defer` with goleak

agaffney (commented):
I'm not actually sure how the second test works. You use the "wrong" keep-alive cookie value both in the mock conversation and the configured `ouroboros.Connection` client, but you still apparently get the expected error.

verbotenj:
> I'm not actually sure how the second test works. You use the "wrong" keep-alive cookie value both in the mock conversation and the configured `ouroboros.Connection` client, but you still apparently get the expected error.

I think ConversationKeepAliveWrongResponse is not used properly (...) since `keepalive.WithCookie(MockKeepAliveWrongCookie),` defines what it should return. It should match and no errors should be thrown.

agaffney (commented):
We don't want to merge tests that don't actually work properly. Try to figure out why the "right" value is being used even when we explicitly give it the "wrong" value so that we're actually testing what we think we're testing.

verbotenj:
> We don't want to merge tests that don't actually work properly. Try to figure out why the "right" value is being used even when we explicitly give it the "wrong" value so that we're actually testing what we think we're testing.

Updated tests with ConverstionEntries that make sense for the tests. The issue was using different cookies in conversation entries and mixing them with the "oConn" cookie config

agaffney (approved):

PR #527, opened 2024-03-11
perf: a little optimization for Transactions() methods

igorcrevar:
Changes inside `Transactions() []Transaction` methods:
- in advance allocate memory for returning slice of Transactions (size of slice is known)
- use map to mark invalid transactions replacing the previous loop through an invalid slice.

Additionally, a `cachedTransactions` slice can be introduced within each struct, allowing for immediate return from the method when the slice is not nil, if you ok with that?

### DIFF
--- a/ledger/allegra.go
+++ b/ledger/allegra.go
@@ -71,14 +71,13 @@ func (b *AllegraBlock) Era() Era {
 }
 
 func (b *AllegraBlock) Transactions() []Transaction {
-	ret := []Transaction{}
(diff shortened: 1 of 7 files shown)


agaffney (approved):
This looks great, and I even learned a few things (such as `t.Run()`). Thanks for your contribution!

agaffney:
@igorcrevar it's probably not that helpful to cache the transactions when it's only avoiding building a small array of known size

PR #589, opened 2024-04-18
feat:add Mary protocol parameters support

### LINKED ISSUE #326: Mary protocol params support

verbotenj:
```
era = Mary, slot = 39916796, block_no = 6236059, id = e72579ff89dc9ed325b723a33624b596c08141c7bd573ecfff56a1f7229e4d09
era = Alonzo, slot = 39916975, block_no = 6236060, id = 8959c0323b94cc670afe44222ab8b4e72cfcad3b5ab665f334bbe642dc6e9ef4
```

Closes #326

### DIFF
--- a/ledger/shelley.go
+++ b/ledger/shelley.go
@@ -432,13 +432,46 @@ type ShelleyProtocolParameterUpdate struct {
 	Rho                *cbor.Rat `cbor:"10,keyasint"`
 	Tau                *cbor.Rat `cbor:"11,keyasint"`
 	Decentralization   *cbor.Rat `cbor:"12,keyasint"`
-	Nonce              *cbor.Rat `cbor:"13,keyasint"`
+	Nonce              *Nonce    `cbor:"13,keyasint"`
 	ProtocolVersion    struct {
 		cbor.StructAsArray
 		Major uint
 		Minor uint
 	} `cbor:"14,keyasint"`
 	MinUtxoValue uint `cbor:"15,keyasint"`
+}
+
+const (
+	NonceType0 = 0
+	NonceType1 = 1
+)
+
+type Nonce struct {
+	cbor.StructAsArray
+	Type  uint
+	Value [32]byte
+}
+
+func (n *Nonce) UnmarshalCBOR(data []byte) error {
+	nonceType, err := cbor.DecodeIdFromList(data)
+	if err != nil {
+		return err
+	}
+
+	n.Type = uint(nonceType)
+
+	switch nonceType {
+	case NonceType0:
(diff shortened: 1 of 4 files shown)


ledger/shelley.go
```diff
@@ -441,6 +441,40 @@ type ShelleyProtocolParameterUpdate struct {
...
 
+const (
+	NonceType0 = 0
+	NonceType1 = 1
+)
+
+type Nonce struct {
+	cbor.StructAsArray
+	Type  uint `cbor:"0,keyasint"`
+	Value any  `cbor:"1,keyasint,omitempty"`
```
agaffney, line 452: Remove these `cbor` tags. They don't do anything when using `cbor.StructAsArray`. The `keyasint` and `omitempty` tags only work when dealing with a map (this is a list).

agaffney, line 452: The nonce value is `bytes .size 32` according to the CDDL, so we should use `[32]byte` instead of `any`

ledger/shelley_test.go
```diff
@@ -0,0 +1,35 @@
...
+			data: []byte{0x81, 0x00},
+		},
+		{
+			name: "NonceType1",
+			data: []byte{0x82, 0x01, 0x01},
+		},
+		{
+			name: "UnsupportedNonceType",
+			data: []byte{0x83, 0x00, 0x01, 0x02},
+		},
```
agaffney, line 23: This doesn't actually fail. Your unmarshal function is checking the first value in the list (`0` in this case), which is a supported value. Instead I'd use something like `8102` (`[2]`), which will cause an error.

agaffney (changes requested):

ledger/shelley.go
verbotenj: When I set to [32]byte, it cannot unmarshal positive integer into Go struct field struct { StructAsArray cbor.StructAsArray; Type uint; Value [32]uint8 }.Value of type [32]uint8

verbotenj: Then I tried to use temp struct

```
		tmp := struct {
			cbor.StructAsArray
			Type  uint
			Value []byte
		}{}

```
And got
cbor: cannot unmarshal positive integer into Go struct field struct { StructAsArray cbor.StructAsArray; Type uint; Value []uint8 }.Value of type []uint8

verbotenj: I tried also

```
		tmp := struct {
			cbor.StructAsArray
			Type  uint
			Value []uint8
		}{}
```
I am not sure what does error want to tell me

agaffney (approved):

PR #599, opened 2024-04-26
feat: add support for TX attribute: protocol parameters update

### LINKED ISSUE #333: Support for TX attribute: protocol parameter updates
Looks like this is being covered here:

- [x] #324 
- [x] #325 
- [x] #326 
- [x] #327 
- [x] #328

verbotenj:
Closes #333

### DIFF
--- a/ledger/allegra.go
+++ b/ledger/allegra.go
@@ -170,6 +170,14 @@ func (t AllegraTransaction) IsValid() bool {
 	return true
 }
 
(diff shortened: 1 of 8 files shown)


wolf31o2 (changes requested):
You need to add the function to the Transaction interface and add it to any remaining TransactionBody types. The computer will let you know if you miss anything once you add it to the interface.

wolf31o2 (approved):

PR #607, opened 2024-05-02
feat: TX consumed/produced helper functions

### LINKED ISSUE #606: Create helper functions for ledger.Transaction to returned consumed inputs and produced outputs
This is mostly needed to handle different behavior when a transaction is marked as "valid" or not. If it's valid, we want to return the standard TX inputs and outputs. If it's marked as not valid, we want to return the collateral inputs and/or collateral return value(s). We only want this on `ledger.Transaction` and not `ledger.TransactionBody` because we need information outside of the TX body to make the determination on validity.

agaffney:
Fixes #606

### DIFF
--- a/ledger/babbage.go
+++ b/ledger/babbage.go
@@ -209,6 +209,12 @@ func (b *BabbageTransactionBody) ReferenceInputs() []TransactionInput {
 }
 
 func (b *BabbageTransactionBody) CollateralReturn() TransactionOutput {
+	// Return an actual nil if we have no value. If we return our nil pointer,
+	// we get a non-nil interface containing a nil value, which is harder to
+	// compare against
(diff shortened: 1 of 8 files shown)


ledger/babbage.go
```diff
@@ -209,6 +209,9 @@ func (b *BabbageTransactionBody) ReferenceInputs() []TransactionInput {
 }
 
 func (b *BabbageTransactionBody) CollateralReturn() TransactionOutput {
+	if b.TxCollateralReturn == nil {
+		return nil
+	}
```
verbotenj, line 214: This seems redundant, because you would return nil anyway

agaffney: I added a comment in the code about why we're doing that

verbotenj (approved):

PR #651, opened 2024-06-05
feat: address hash function

wolf31o2:


agaffney (commented):
I'm not sure that's going to give a useful value. IIRC, the AddrKeyHash for the certs refers to a 28-byte address key hash, which is not the entire address. it should just be the payment or staking portion of the address

agaffney (approved):

PR #668, opened 2024-07-11
fix: add nil handling for BabbageTransactionOutput Utxorpc()

verbotenj:
include unit tests


ledger/babbage_test.go
```diff
@@ -83,3 +86,77 @@ func TestBabbageBlockTransactions(t *testing.T) {
...
+	}
+}
+
+type TxOutput struct {
+	Address   []byte
+	Coin      uint64
+	DatumHash []byte
+}
+
+func TestUtxorpc_DatumOptionNil(t *testing.T) {
```
agaffney, line 133: We should keep the test names prefixed with the object type, since all tests under `ledger/` are in the same namespace. This should be something like `TestBabbageTransactionOutput_Utxorpc_DatumOptionNil`

ledger/babbage_test.go
```diff
@@ -83,3 +86,77 @@ func TestBabbageBlockTransactions(t *testing.T) {
...
+
+	txOutput := output.Utxorpc()
+
+	assert.NotNil(t, txOutput)
+	assert.Equal(t, []byte{}, txOutput.DatumHash)
+	assert.Equal(t, []byte{0x0}, txOutput.Address)
+	assert.Equal(t, uint64(1000), txOutput.Coin)
+}
+
+func TestDatumHashReturnsNil(t *testing.T) {
```
agaffney, line 150: Same as above

agaffney (approved):

PR #707, opened 2024-09-10
chore: replace retired iohk.io nodes in docs/tests

wolf31o2:


wolf31o2:
This is the same as #706 so closing.

PR #760, opened 2024-10-15
feat(address): improved NewAddress to support both bech32/base58

verbotenj:


ledger/common/address.go
```diff
@@ -53,15 +54,25 @@ type Address struct {
...
-	if err != nil {
-		return Address{}, err
-	}
-	decoded, err := bech32.ConvertBits(data, 5, 8, false)
-	if err != nil {
-		return Address{}, err
+	var decoded []byte
+	var err error
+
+	if isMixedCase(addr) {
```
agaffney, line 64: We can probably drop the helper function and just do something like `if strings.ToLower(addr) != addr {`. This will effectively detect upper-case characters in the string, which we should never see for bech32.

verbotenj: That's a smart move, thanks

agaffney (approved):

PR #770, opened 2024-10-28
fix: Acquire nil point

igorcrevar:
If `localtxmonitor` => `Acquire` is called with a `nil` point (indicating the latest point), a bug occurs because the logger attempts to print the point's `Hash` and `slot`, even though point is `nil`


wolf31o2 (approved):
This is probably not the only place I did this, unfortunately. Thanks for pointing it out and fixing this one. I'll make a pass over all of the recent logging changes to make sure this doesn't happen elsewhere.

wolf31o2:
Somehow, that was the only one. Nice find.

igorcrevar:
When will be a new release with this fix? we have one app that depends on it :)

Issue #774, opened 2024-10-29
Deadlock with connection.BlockFetch().Client.GetBlock


igorcrevar:
If `connection.BlockFetch().Client.GetBlock` is executed on a closed connection, the next call will block indefinitely. I believe the issue is caused by a mutex that is not being unlocked. See the code below for reproduction:

```
func bugTest() error {
	connection, err := ouroboros.NewConnection(
		ouroboros.WithNetworkMagic(uint32(1)),
		ouroboros.WithNodeToNode(true),
	)
	if err != nil {
		return err
	}

	// dial some known node -> connect to node
	if err := connection.Dial("tcp", "preprod-node.play.dev.cardano.org:3001"); err != nil {
		return err
	}

	blockHash, err := hex.DecodeString("f7e99e7d597ef0b49ac4b442addb208b09ef928ee729487bc2d0a4e65dd11a66")
	if err != nil {
		return err
	}
        // some point that exists on testnet
	bp := common.Point{
		Slot: 74526873,
		Hash: blockHash,
	}

	b, err := connection.BlockFetch().Client.GetBlock(bp) // first call will pass
	if err != nil {
		return err
	}

	fmt.Println(b.SlotNumber(), b.Hash(), b.Transactions()) // will print some data

	connection.Close()

	_, err1 := connection.BlockFetch().Client.GetBlock(bp) // will retrieve error
	fmt.Println(err1) // this will print error "protocol is shutting down"

	_, err2 := connection.BlockFetch().Client.GetBlock(bp) // this will block
	fmt.Println(err2)

	return errors.Join(err1, err2)
}
```

agaffney:
Fixed in #787

PR #787, opened 2024-11-11
fix: connection.BlockFetch().Client.GetBlock potential deadlock

Newt6611:
issue #774 

Unlock busyMutex when returning `ProtocolShuttingDownError`


wolf31o2 (approved):
Thanks for the fix!

PR #811, opened 2024-12-27
feat: support for acquire immutable tip in local-state-query

### LINKED ISSUE #766: Add support for additional local-state-query Re/Acquire messages
Another variant of the Re/Acquire messages was added to the spec. In addition to the specific point and NoPoint (now called VolatileTip) variants, there's now an ImmutableTip with message IDs 10 (Acquire) and 11 (ReAcquire).

agaffney:
* rename (Re)AcquireNoPoint to (Re)AcquireVolatileTip
* add new messages for ImmutableTip
* add client functions for acquiring volatile/immutable tip
* change Acquire callback signature for immutable trip

Fixes #766

### DIFF
--- a/protocol/localstatequery/localstatequery.go
+++ b/protocol/localstatequery/localstatequery.go
@@ -47,7 +47,11 @@ var StateMap = protocol.StateMap{
 				NewState: stateAcquiring,
 			},
 			{
-				MsgType:  MessageTypeAcquireNoPoint,
+				MsgType:  MessageTypeAcquireVolatileTip,
+				NewState: stateAcquiring,
+			},
+			{
+				MsgType:  MessageTypeAcquireImmutableTip,
 				NewState: stateAcquiring,
 			},
 			{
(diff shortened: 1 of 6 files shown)


protocol/localstatequery/localstatequery.go
```diff
@@ -121,6 +129,13 @@ type Config struct {
 	QueryTimeout   time.Duration
 }
 
+// Helper types
+type AcquirePoint struct {
+	Point     *common.Point
+	Volatile  bool
+	Immutable bool
```
wolf31o2, line 136: Can it ever be neither or both? Are these completely independent from each other?

wolf31o2 (approved):

protocol/localstatequery/localstatequery.go
agaffney: It can only be one of them at a time. It might make a little more sense to have a "type" field to make it a bit more explicit which one it is, or even separate types for each with an interface type

agaffney: I switch this to 4 separate types, one for each acquire target type, and an interface that covers all of them for convenience.

wolf31o2 (approved):

PR #951, opened 2025-03-13
feat(ledger): Decode block header as proper era type

### LINKED ISSUE #844: Decode block header as proper era type
We hand-wave away the differences between Shelley/Allegra/Mary/Alonzo and Babbage/Conway when decoding block headers (in `NewBlockHeaderFromCbor`) because they're technically the same shape. We should break out the cases and use the proper block header type for the era.

arepala-uml:
1. Added changes for decoding specific block header type and written separate cases for Shelley/Allegra/Mary/Alonzo and Babbage/Conway in NewBlockHeaderFromCbor.
2. Added unit test cases for each case to validate whether the correct corresponding function is executed in NewBlockHeaderFromCbor.

Closes #844


ledger/verify_block_test.go
wolf31o2, line 1: Move the tests you created here to a `block_test.go` file, since we're testing functionality in `block.go`

ledger/block.go
```diff
@@ -54,12 +54,21 @@ func NewBlockHeaderFromCbor(blockType uint, data []byte) (BlockHeader, error) {
 	case BlockTypeByronMain:
 		return NewByronMainBlockHeaderFromCbor(data)
 	// TODO: break into separate cases and parse as specific block header types (#844)
```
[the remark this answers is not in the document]
arepala-uml: I will remove this comment

ledger/verify_block_test.go
arepala-uml: I will make this change

wolf31o2 (approved):

PR #960, opened 2025-03-23
Add JSON labels to Genesis structs

### LINKED ISSUE #926: Add JSON labels to Genesis structs
Some of the structs are missing them, which causes the golangci-lint `musttag` linter to fail. I think it's a good idea to be explicit on there. Fix the issue and remove the `//nolint:musttag` comments to test.

arepala-uml:
1. Added missing `json` struct tags to multiple genesis structs.
2. Remove nolint comments from reader constructors
3. Add unit tests to verify JSON decoding

Closes #926


wolf31o2 (approved):
LGTM... when you squash and merge, update the commit message to something like:

`fix(ledger): json labels on genesis objects`

PR #962, opened 2025-03-23
fix(ledger): use a named type for credential hash

wolf31o2:
This makes it more like other hashes where we have defined which hash type it should be underneath.

### DIFF
--- a/internal/test/ledger/ledger.go
+++ b/internal/test/ledger/ledger.go
@@ -51,7 +51,11 @@ func (ls MockLedgerState) StakeRegistration(
 ) ([]common.StakeRegistrationCertificate, error) {
 	ret := []common.StakeRegistrationCertificate{}
 	for _, cert := range ls.MockStakeRegistration {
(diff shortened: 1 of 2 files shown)


internal/test/ledger/ledger.go
```diff
@@ -51,7 +51,11 @@ func (ls MockLedgerState) StakeRegistration(
 ) ([]common.StakeRegistrationCertificate, error) {
 	ret := []common.StakeRegistrationCertificate{}
 	for _, cert := range ls.MockStakeRegistration {
-		if string(cert.StakeRegistration.Credential) == string(stakingKey) {
+		if string(
+			common.Blake2b224(cert.StakeRegistration.Credential).Bytes(),
```
agaffney, line 55: You can probably avoid this whole dance with `[]byte(cert.StakeRegistration.Credential)`

agaffney (approved):

internal/test/ledger/ledger.go
wolf31o2: I kept this because it's what's being done in the other test.

PR #964, opened 2025-03-24
fix(genesis): Added json labels to all genesis structs

### LINKED ISSUE #926: Add JSON labels to Genesis structs

arepala-uml:
1. Added json tags to all the genesis structs.

Closes #926

### DIFF
--- a/ledger/common/genesis.go
+++ b/ledger/common/genesis.go
@@ -21,7 +21,7 @@ import (
 
 // GenesisRat is a wrapper to big.Rat that allows for unmarshaling from a bare float from JSON
 type GenesisRat struct {
-	*big.Rat `json:"string"`
+	*big.Rat
 }
 
 func (r *GenesisRat) UnmarshalJSON(data []byte) error {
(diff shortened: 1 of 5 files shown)


ledger/common/genesis.go
```diff
@@ -21,7 +21,7 @@ import (
 
 // GenesisRat is a wrapper to big.Rat that allows for unmarshaling from a bare float from JSON
 type GenesisRat struct {
-	*big.Rat
+	*big.Rat `json:"string"`
```
agaffney, line 24: This is "wrong" and probably not effective in any meaningful way. This field is an embedded struct, and the `GenesisRat` type has custom `UnmarshalJSON` and `MarshalJSON` functions

arepala-uml: I have removed it and tested again. Please review it once

agaffney (approved):

PR #971, opened 2025-03-30
feat(ledger): Add a method to get the Certificate type to the interface

### LINKED ISSUE #636: Add a method to get the Certificate type to the interface
A common use case for Certificate data will be to only operate on specific certificate types. It would greatly simplify downstream if we had a way to access the type from the Certificate interface.

arepala-uml:
1. Added a new Type() uint method to the Certificate interface.
2. Added this method for all 19 certificate types using the existing CertType field. 
3. Added a test function to validate the method Type() returns the expected values.

Closes #636

### DIFF
--- a/ledger/common/common_test.go
+++ b/ledger/common/common_test.go
@@ -243,3 +243,51 @@ func TestBlake2b224_String(t *testing.T) {
 		t.Errorf("expected %s but got %s", expected, hash.String())
 	}
 }
+
+func TestCertificateTypeMethods(t *testing.T) {
+	tests := []struct {
+		name     string
+		cert     Certificate
+		expected uint
+	}{
+		{"StakeRegistration", &StakeRegistrationCertificate{CertType: CertificateTypeStakeRegistration}, CertificateTypeStakeRegistration},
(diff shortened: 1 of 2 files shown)


ledger/common/common_test.go
```diff
@@ -243,3 +243,51 @@ func TestBlake2b224_String(t *testing.T) {
...
+				t.Logf("PASS: %s -> Type() = %d, expected = %d \n", tt.name, got, tt.expected)
+			}
+		})
+	}
+
+	if allPassed {
+		t.Logf("\n ALL CERTIFICATE TYPE TESTS PASSED SUCCESSFULLY")
+	} else {
+		t.Logf("\n SOME CERTIFICATE TYPE TESTS FAILED ")
+	}
```
wolf31o2, line 292: Get rid of this. We're already showing pass or fail on each.

wolf31o2 (changes requested):
Add a negative test case, too. It doesn't need to fit into the existing loop. Let's ensure we're throwing an error correctly when they do not match.

ledger/common/common_test.go
arepala-uml: Sure, I will modify it

wolf31o2 (approved):

PR #979, opened 2025-04-12
refactor: remove embeds in per-era TransactionBody types

### LINKED ISSUE #978: Un-embed TransactionBody types
We build most eras' TransactionBody type by embedding the previous era's type inside it. This reduces code duplication but makes other things awkward, such as inline initialization of an instance of these types. Remove the embedding and instead duplicate the struct fields and helper functions.

agaffney:
Fixes #978

### DIFF
--- a/ledger/common/tx.go
+++ b/ledger/common/tx.go
@@ -93,3 +93,134 @@ type Utxo struct {
 	Id     TransactionInput
 	Output TransactionOutput
 }
+
+// TransactionBodyBase provides a set of functions that return empty values to satisfy the
+// TransactionBody interface. It also provides functionality for generating a transaction hash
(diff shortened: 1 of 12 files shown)


ledger/common/tx.go
```diff
@@ -93,3 +93,134 @@ type Utxo struct {
 	Id     TransactionInput
 	Output TransactionOutput
 }
+
+// TransactionBodyBase provides a set of functions that return empty values to satisfy the
+// TransactionBody interface. It also provides functionality for generating a transaction hash
+// and storing/retrieving the original CiBOR
```
wolf31o2, line 99: CBOR

wolf31o2 (approved):
Typo in a comment, but otherwise looks good

PR #984, opened 2025-04-14
feat: changes to support cost models while updating Alonzo protocols through genesis config

### LINKED ISSUE #852: Support for cost models when updating Alonzo protocol params from genesis config
We currently ignore the cost models, because it's 150+ string keys in the JSON that we need to map to numeric indexes. Implement this properly in `alonzo.UpdateFromGenesis()`

### LINKED ISSUE #963: CostModels can be arrays or maps
It was reported in blinklabs-io/dingo#518 that the CostModels in the genesis files can be either an array or a map. We ship the Hydra devnet genesis files with Dingo, which uses arrays. Both are valid. Update the cost and write tests to validate parsing of both `map[string]int` and `[]int`.

jkawan:
1. Changes added to support cost models when updating Alonzo protocol params from genesis config - mapped string keys to numeric indexes
2. Changes to support both array and map from genesis files
3. Added unit tests for the changes

Closes #852 
Closes #963

### DIFF
--- a/ledger/alonzo/pparams.go
+++ b/ledger/alonzo/pparams.go
@@ -17,26 +17,25 @@ package alonzo
 import (
 	"fmt"
 	"math"
-	"sort"
-	"strings"
 
 	"github.com/blinklabs-io/gouroboros/cbor"
 	"github.com/blinklabs-io/gouroboros/ledger/common"
+	"github.com/blinklabs-io/gouroboros/ledger/mary"
 	cardano "github.com/utxorpc/go-codegen/utxorpc/v1alpha/cardano"
 )
 
-type PlutusVersion string
-
+// Constants for Plutus version mapping
 const (
-	PlutusV1 PlutusVersion = "PlutusV1"
-	PlutusV2 PlutusVersion = "PlutusV2"
-	PlutusV3 PlutusVersion = "PlutusV3"
+	PlutusV1Key uint = 0
+	PlutusV2Key uint = 1
+	PlutusV3Key uint = 2
 )
 
-type CostModel struct {
-	Version    PlutusVersion
-	Parameters map[string]int64
-	Order      []string
+// Expected parameter counts for validation
+var plutusParamCounts = map[uint]int{
+	PlutusV1Key: 166,
+	PlutusV2Key: 175,
+	PlutusV3Key: 187,
 }
 
 type AlonzoProtocolParameters struct {
@@ -60,7 +59,7 @@ type AlonzoProtocolParameters struct {
 	MinUtxoValue         uint
 	MinPoolCost          uint64
 	AdaPerUtxoByte       uint64
-	CostModels           map[PlutusVersion]*CostModel
+	CostModels           map[uint][]int64
 	ExecutionCosts       common.ExUnitPrice
 	MaxTxExUnits         common.ExUnits
 	MaxBlockExUnits      common.ExUnits
@@ -128,7 +127,7 @@ func (p *AlonzoProtocolParameters) Update(
 		p.AdaPerUtxoByte = *paramUpdate.AdaPerUtxoByte
 	}
 	if paramUpdate.CostModels != nil {
-		p.convertLegacyCostModels(paramUpdate.CostModels)
+		p.CostModels = paramUpdate.CostModels
 	}
 	if paramUpdate.ExecutionCosts != nil {
 		p.ExecutionCosts = *paramUpdate.ExecutionCosts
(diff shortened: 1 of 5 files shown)


agaffney (changes requested):
While this approach will technically work for decoding the Alonzo genesis in most networks, we'll want to be more explicit with the mapping of these keys to numeric index. This will allow us to convert back and forth between the formats, as well as do things like validate the inputs. Each of the 3 Plutus versions will have their own mapping.

We also need to be able to support the simple list format as seen in the `devnet` genesis config:

https://github.com/blinklabs-io/docker-cardano-configs/blob/main/config/devnet/alonzo-genesis.json

wolf31o2:
> We also need to be able to support the simple list format as seen in the `devnet` genesis config:

We have a separate issue for that one.

ledger/alonzo/pparams.go
```diff
@@ -44,7 +60,7 @@ type AlonzoProtocolParameters struct {
 	MinUtxoValue         uint
 	MinPoolCost          uint64
 	AdaPerUtxoByte       uint64
-	CostModels           map[uint][]int64
+	CostModels           map[PlutusVersion]*CostModel
```
agaffney, line 63: This is (probably) going to break CBOR decoding of Alonzo protocol params. The previous struct field definition matches the structure of the CBOR data.

ledger/alonzo/pparams.go
```diff
@@ -286,27 +439,3 @@ func (p *AlonzoProtocolParameters) Utxorpc() *cardano.PParams {
...
-		A0:                 prevPParams.A0,
-		Rho:                prevPParams.Rho,
-		Tau:                prevPParams.Tau,
-		Decentralization:   prevPParams.Decentralization,
-		ExtraEntropy:       prevPParams.ExtraEntropy,
-		ProtocolMajor:      prevPParams.ProtocolMajor,
-		ProtocolMinor:      prevPParams.ProtocolMinor,
-		MinUtxoValue:       prevPParams.MinUtxoValue,
-	}
-}
```
agaffney, line 312: Why did you remove this function? We use this in dingo

ledger/alonzo/pparams.go
```diff
@@ -150,16 +168,115 @@ func (p *AlonzoProtocolParameters) UpdateFromGenesis(genesis *AlonzoGenesis) {
...
+}
+
+func toPlutusVersion(key string) (PlutusVersion, bool) {
+	switch strings.ToLower(key) {
+	case "plutus:v1", "plutusv1":
+		return PlutusV1, true
+	case "plutus:v2", "plutusv2":
+		return PlutusV2, true
+	case "plutus:v3", "plutusv3":
+		return PlutusV3, true
```
agaffney, line 249: Where are any of these forms used?

ledger/alonzo/pparams.go
jkawan: Apologies, I removed this function unintentionally while resolving the conflicts. fixing it

ledger/alonzo/pparams.go
jkawan: fixing the forms

ledger/alonzo/pparams.go
```diff
@@ -150,16 +167,78 @@ func (p *AlonzoProtocolParameters) UpdateFromGenesis(genesis *AlonzoGenesis) {
 		Memory: uint64(genesis.MaxBlockExUnits.Mem),
 		Steps:  uint64(genesis.MaxBlockExUnits.Steps),
 	}
-	if genesis.ExecutionPrices.Mem != nil &&
-		genesis.ExecutionPrices.Steps != nil {
+
+	if genesis.ExecutionPrices.Mem != nil && genesis.ExecutionPrices.Steps != nil {
```
agaffney, line 171: This is going to get split back to 2 lines the next time that we run `golines` on this repo

ledger/alonzo/pparams.go
```diff
@@ -150,16 +167,78 @@ func (p *AlonzoProtocolParameters) UpdateFromGenesis(genesis *AlonzoGenesis) {
...
+			switch v := model.(type) {
+			case map[string]interface{}:
+				maxIndex := 0
+				// Find maximum parameter index
+				for paramName := range v {
+					var index int
+					if _, err := fmt.Sscanf(paramName, "param%d", &index); err == nil && index > maxIndex {
+						maxIndex = index
+					}
+				}
```
agaffney, line 197: This seems to be expecting a structure like this, but I'm not seeing where that would come from outside of the tests. It seems that it would be easier to use `[]int64` or similar.

```
{ "param0": 123, "param1": 234, "param2": 345 }
```

ledger/alonzo/pparams.go
```diff
@@ -150,16 +167,78 @@ func (p *AlonzoProtocolParameters) UpdateFromGenesis(genesis *AlonzoGenesis) {
...
+						}
+					}
+				}
+
+			case []interface{}:
+				values = make([]int64, len(v))
+				for i, val := range v {
+					if intVal, ok := val.(float64); ok {
+						values[i] = int64(intVal)
+					}
```
agaffney, line 213: Will this only ever contain `float64` values? It looks like anything else would be ignored

agaffney (changes requested):

agaffney (approved):

PR #992, opened 2025-04-17
Cardano blueprint localstate query tests

ch1bo:
Explores how we could test the [LocalStatQuery API](https://github.com/cardano-scaling/cardano-blueprint/issues/25) that is being created in course of the blueprint initiative.

- This is adding a `cardano-blueprint` submodule from which local state `query` and `result` cbor examples are taken to test serialization roundtrips.

- Changes `SystemStartResult` to hold `big.Int` for `Year` and `Picoseconds` as the CDDL (and at least the Haskell server) are using unbounded integers

  ```go
  type SystemStartResult struct {
	  // Tells the CBOR decoder to convert to/from a struct and a CBOR array
	  _           struct{} `cbor:",toarray"`
	  Year        big.Int
	  Day         int
	  Picoseconds big.Int
  }
  ```
  - This required a few additional custom methods on `SystemStartResult` to make roundtrips work. Happy for some feedback in how to improve / avoid those changes.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "cardano-blueprint"]
+	path = cardano-blueprint
+	url = git@github.com:cardano-scaling/cardano-blueprint
--- a/cmd/gouroboros/query.go
+++ b/cmd/gouroboros/query.go
@@ -124,7 +124,9 @@ func testQuery(f *globalFlags) {
 			os.Exit(1)
 		}
 		fmt.Printf(
-			"system-start: year = %d, day = %d, picoseconds = %d\n",
+			// REVIEW: %d should work for big/Int, but warns and produces output
+			// like {%!d(bool=false) [2025]}
+			"system-start: year = %v, day = %d, picoseconds = %v\n",
 			systemStart.Year,
 			systemStart.Day,
 			systemStart.Picoseconds,
--- a/protocol/localstatequery/client_test.go
+++ b/protocol/localstatequery/client_test.go
@@ -17,6 +17,7 @@ package localstatequery_test
 import (
 	"encoding/json"
 	"fmt"
+	"math/big"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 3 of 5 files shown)


agaffney:
That failing test you noted is due to pointer vs. non-pointer in the DeepEqual() comparison

agaffney:
@ch1bo it took a bit of playing around, but I figured out how to fix the test failure. Reflection is fun!

```
diff --git a/protocol/localstatequery/messages_test.go b/protocol/localstatequery/messages_test.go
index 4156f5f..cbe1fba 100644
--- a/protocol/localstatequery/messages_test.go
+++ b/protocol/localstatequery/messages_test.go
@@ -135,10 +135,10 @@ func TestDecode(t *testing.T) {
                        if err != nil {
                                t.Fatalf("failed to decode result: %s", err)
                        }
-                       if !reflect.DeepEqual(decoded.Interface(), test.Result) {
+                       if !reflect.DeepEqual(reflect.Indirect(decoded).Interface(), test.Result) {
                                t.Fatalf(
                                        "MsgResult content did not decode to expected Result object\n  got:    %#v\n  wanted: %#v",
-                                       decoded.Interface(),
+                                       reflect.Indirect(decoded).Interface(),
                                        test.Result,
                                )
                        }
```

ch1bo:
Squashed into a single commit following https://www.conventionalcommits.org/en/v1.0.0/ as suggested by @wolf31o2. Only found these [contributing guidelines](https://github.com/blinklabs-io/.github/blob/01b39f20bbd57195af6c211c456a67fb2286e399/CONTRIBUTING.md) using the github UI. Originally I was looking for a `CONTRIBUTING.md` in the repo but did not find one.

wolf31o2:
Thanks for the contribution. We had per-repo contributing files, but they were redundant to maintain, especially remembering them in new repos. I'll look at adding them back to each repo.

wolf31o2:
I have no idea why this is still not processing on the GitHub side. Since it's acting so weird, I decided to make a new branch, cherry-pick your changes, rebase them on main, and add json struct tags to make golangci-lint happy.

The pull request in #1001 replaces this one.

PR #1024, opened 2025-06-02
feat(ledger/babbage): Integer underflow issue in babbage.UtxoValidateCollateralEqBalance

### LINKED ISSUE #1000: Integer underflow issue in babbage.UtxoValidateCollateralEqBalance
We (purposely) brush off errors looking up collateral UTxOs but then proceed to subtract the collateral return amount from zero, resulting in a very large number due to a `uint` underflow. We should check for the collateral inputs totaling to zero and either skip going through the collateral returns or immediately return our error.

arepala-uml:
1. Fixed an uint underflow in UtxoValidateCollateralEqBalance by skipping validation when no valid collateral UTxOs are found.  
2. Added a unit test to ensure collateral return subtraction is skipped when collBalance is zero, avoiding invalid balance validation.

Closes #1000

### DIFF
--- a/ledger/babbage/rules.go
+++ b/ledger/babbage/rules.go
@@ -184,10 +184,21 @@ func UtxoValidateCollateralEqBalance(
 		}
 		collBalance += utxo.Output.Amount()
 	}
-	// Subtract collateral return amount
+
+	// Skip validation if no valid collateral UTxOs were found
+	// This avoids subtracting from zero and prevents uint underflow
+	if collBalance == 0 {
+		return nil
+	}
+
+	// Subtract collateral return amount with underflow protection
 	collReturn := tx.CollateralReturn()
 	if collReturn != nil {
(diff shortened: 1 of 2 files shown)


ledger/babbage/rules.go
```diff
@@ -184,10 +184,21 @@ func UtxoValidateCollateralEqBalance(
...
+		return nil
+	}
+
+	// Subtract collateral return amount with underflow protection
 	collReturn := tx.CollateralReturn()
 	if collReturn != nil {
-		collBalance -= collReturn.Amount()
+		returnAmt := collReturn.Amount()
+		if collBalance < returnAmt {
+			return errors.New("collateral return amount exceeds collateral input balance")
```
agaffney, line 199: We only want to return the specific errors defined in the ledger for each era, and generally only at the end of the function

agaffney (changes requested):

PR #1038, opened 2025-06-12
fix(muxer): Return better error message when the peer closes the connection

### LINKED ISSUE #518: Return better error message when the peer closes the connection
We currently propagate the `EOF` returned by the underlying read call, but we could capture this and return a better error such as `the peer closed the connection` or `the connection was closed unexpectedly`

arepala-uml:
1. Replaced plain io.EOF with descriptive wrapped errors like peer closed the connection while reading header: EOF
in Muxer.readLoop(). 
2. Updated downstream handling in setupConnection() function and muxer error forwarding to preserve
the wrapped errors.

Closes #518

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -286,8 +286,8 @@ func (c *Connection) setupConnection() error {
 				return
 			}
 			if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
-				// Return a bare io.EOF error if error is EOF/ErrUnexpectedEOF
-				c.errorChan <- io.EOF
+				// Preserve the original wrapped error with friendly message
+				c.errorChan <- err
 			} else {
 				// Wrap error message to denote it comes from the muxer
 				c.errorChan <- fmt.Errorf("muxer error: %w", err)
@@ -354,7 +354,7 @@ func (c *Connection) setupConnection() error {
 	select {
 	case <-c.doneChan:
 		// Return an error if we're shutting down
-		return io.EOF
+		return fmt.Errorf("connection shutdown initiated: %w", io.EOF)
 	case err := <-c.protoErrorChan:
 		// Shutdown the connection and return the error
 		c.Close()
--- a/connection.go
+++ b/connection.go
@@ -285,8 +285,9 @@ func (c *Connection) setupConnection() error {
 			if !ok {
 				return
 			}
-			if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
-				// Preserve the original wrapped error with friendly message
+			var connErr *muxer.ConnectionClosedError
+			if errors.As(err, &connErr) {
+				// Pass through ConnectionClosedError from muxer
 				c.errorChan <- err
 			} else {
 				// Wrap error message to denote it comes from the muxer
--- a/muxer/muxer.go
+++ b/muxer/muxer.go
@@ -69,6 +69,19 @@ type Muxer struct {
 	onceStop               sync.Once
 }
 
+type ConnectionClosedError struct {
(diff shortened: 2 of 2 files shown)


muxer/muxer.go
```diff
@@ -297,10 +298,11 @@ func (m *Muxer) readLoop() {
 		// We use ReadFull because it guarantees to read the expected number of bytes or
 		// return an error
 		if _, err := io.ReadFull(m.conn, msg.Payload); err != nil {
-			if errors.Is(err, io.ErrClosedPipe) {
-				err = io.EOF
+			if errors.Is(err, io.ErrClosedPipe) || errors.Is(err, io.EOF) {
+				m.sendError(fmt.Errorf("peer closed the connection while reading payload: %w", err))
```
agaffney, line 302: Let's create a `ConnectionClosedError` (or similar) type that wraps the underlying error. This gives consumers something to match against without them having to know that it wraps an `io.EOF` or similar error

muxer/muxer.go
```diff
@@ -297,10 +298,11 @@ func (m *Muxer) readLoop() {
 		// We use ReadFull because it guarantees to read the expected number of bytes or
 		// return an error
 		if _, err := io.ReadFull(m.conn, msg.Payload); err != nil {
-			if errors.Is(err, io.ErrClosedPipe) {
-				err = io.EOF
```
agaffney, line 301: We still want to map any other EOF-like errors into `io.EOF` to keep existing code working

muxer/muxer.go
```diff
@@ -284,10 +284,11 @@ func (m *Muxer) readLoop() {
 		}
 		header := SegmentHeader{}
 		if err := binary.Read(m.conn, binary.BigEndian, &header); err != nil {
-			if errors.Is(err, io.ErrClosedPipe) {
-				err = io.EOF
+			if errors.Is(err, io.ErrClosedPipe) || errors.Is(err, io.EOF) {
+				m.sendError(fmt.Errorf("peer closed the connection while reading header: %w", err))
```
agaffney, line 288: We should use a new `ConnectionClosedError` type here like I mentioned below

connection.go
```diff
@@ -286,8 +286,8 @@ func (c *Connection) setupConnection() error {
 				return
 			}
 			if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
```
agaffney, line 288: We should change this check to look for the `muxer.ConnectionClosedError` type that we create

agaffney (changes requested):

connection.go
```diff
@@ -286,8 +286,8 @@ func (c *Connection) setupConnection() error {
...
-				// Return a bare io.EOF error if error is EOF/ErrUnexpectedEOF
-				c.errorChan <- io.EOF
+				// Preserve the original wrapped error with friendly message
```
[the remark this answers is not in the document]
arepala-uml: I have made the changes for this comment message and please review it once

muxer/muxer.go
arepala-uml: I have added a new ConnectionClosedError and please review it once

muxer/muxer.go
arepala-uml: I have made the changes for this and please review it once

muxer/muxer.go
arepala-uml: I have used a new ConnectionClosedError and please review it once

connection.go
arepala-uml: I have made the changes to look for ConnectionClosedError and please review it once

agaffney (approved):

PR #1049, opened 2025-06-14
feat: added changes to implement UTxOWholeResult type

jkawan:
1. Added changes in GetUTxOWhole() to print the result in json format.
2. Added query "utxo-whole-result" in query.go to test and find out the result in json format.

TODO: Implement UTxOWholeResult type once we can find the structure from the json format

### DIFF
--- a/protocol/localstatequery/client.go
+++ b/protocol/localstatequery/client.go
@@ -15,6 +15,7 @@
 package localstatequery
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"sync"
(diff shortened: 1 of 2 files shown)


agaffney (changes requested):
We also need to flesh out the `UTxOWholeResult` type as part of this. It should be similar to `UTxOByAddressResult`

protocol/localstatequery/client.go
```diff
@@ -499,6 +500,12 @@ func (c *Client) GetUTxOWhole() (*UTxOWholeResult, error) {
 	if err := c.runQuery(query, &result); err != nil {
 		return nil, err
 	}
+	jsonData, err := json.MarshalIndent(result, "", "  ")
+	if err != nil {
+		return nil, fmt.Errorf("error marshaling UTxOWhole result to JSON: %w", err)
+	}
+	fmt.Println("UTxOWhole Result:")
+	fmt.Println(string(jsonData))
```
wolf31o2, line 508: Remove this. We only want changes (currently) in the binary so we can determine the correct shape.

wolf31o2 (approved):

PR #1054, opened 2025-07-06
feat(shelley): created genesis pools from shelley genesis

### LINKED ISSUE #1042: Create genesis pools from Shelley Genesis
The Shelley genesis file has a `staking` key, which contains a `pools` map of pool IDs (in hex string form) with the pool parameters and a `stake` map which maps stake addresses (hex string) to pool IDs (also hex string).

```json
  "staking": {
    "pools": {
      "0aedc455785463235311c990f68742c9043cd79af09ab31c2ba5e195": {
        "cost": 340000000,
        "margin": 0.0,
        "metadata": null,
        "owners": [],
        "pledge": 0,
        "publicKey": "0aedc455785463235311c990f68742c9043cd79af09ab31c2ba5e195",
        "relays": [
          {
            "single host name": {
              "dnsName": "p4.example",
              "port": 3001
            }
          }
        ],
        "rewardAccount": {
          "credential": {
            "key hash": "6079cde665c2035b8d9ac8929307bdd7f20a51e678e9d4a5e39ace3a"
          },
          "network": "Testnet"
        },
        "vrf": "eb53a17fbad9b7ea0bcf1e1ea89355305600d593b426dfc3084a924d8877d47e"
      }
    },
    "stake": {
      "24632b71152f31516054075897d0d4ababc33204f8a8661136d49e36": "0aedc455785463235311c990f68742c9043cd79af09ab31c2ba5e195"
    }
  }
```

We need a function which can generate the list of initial pools.

jkawan:
1. Added function to generate initial pools
2. Added changes to convert pools to cbor format
3. Added function to return a specific pool by its ID
4. Added unit tests for the above changes

Closes #1042

### DIFF
--- a/ledger/common/address.go
+++ b/ledger/common/address.go
@@ -436,22 +436,14 @@ func (a Address) Bytes() ([]byte, error) {
 		}
 		return ret, nil
 	}
-
-	header := (a.addressType << 4) | (a.networkId & AddressHeaderNetworkMask)
-
-	if a.addressType == AddressTypeNoneScript && a.networkId == AddressNetworkTestnet {
-		header = (AddressTypeNoneScript << 4) | 1
-	}
-
 	ret := []byte{}
 	ret = append(
 		ret,
-		header,
+		(a.addressType<<4)|(a.networkId&AddressHeaderNetworkMask),
 	)
 	ret = append(ret, a.paymentAddress...)
 	ret = append(ret, a.stakingAddress...)
 	ret = append(ret, a.extraData...)
-
 	return ret, nil
 }
 
--- a/ledger/shelley/genesis_test.go
+++ b/ledger/shelley/genesis_test.go
@@ -330,28 +330,20 @@ func TestGenesisStaking(t *testing.T) {
 			t.Errorf("Expected 1 delegator, got %d", len(delegs))
 		} else {
 			// Verify address format
-			addrBytes, err := delegs[0].Bytes()
-			if err != nil {
-				t.Fatalf("Failed to get address bytes: %v", err)
+			// Verify address type and network
+			if delegs[0].NetworkId() != common.AddressNetworkTestnet {
+				t.Errorf("Expected testnet address, got network ID %d", delegs[0].NetworkId())
 			}
 
-			// Should be 29 bytes (1 header + 28 stake key)
-			if len(addrBytes) != 29 {
-				t.Errorf("Expected address length 29, got %d", len(addrBytes))
-			}
-
-			// Verify testnet script stake address header
-			// In TestInitialPools and TestPoolById, replace the header check with:
-			if addrBytes[0] != expectedTestnetScriptStakeHeader {
-				t.Errorf("Expected header byte %x, got %x",
-					expectedTestnetScriptStakeHeader, addrBytes[0])
+			if delegs[0].Type() != common.AddressTypeNoneScript {
+				t.Errorf("Expected script stake address type, got %d", delegs[0].Type())
 			}
 
 			// Verify stake key matches
-			stakeKey := hex.EncodeToString(addrBytes[1:])
-			if stakeKey != expectedStakeKey {
+			stakeKeyHash := delegs[0].StakeKeyHash()
+			if hex.EncodeToString(stakeKeyHash[:]) != expectedStakeKey {
 				t.Errorf("Delegator key mismatch:\nExpected: %s\nActual:   %s",
-					expectedStakeKey, stakeKey)
+					expectedStakeKey, hex.EncodeToString(stakeKeyHash[:]))
 			}
 		}
 	})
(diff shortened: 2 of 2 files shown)


ledger/shelley/genesis.go
```diff
@@ -128,6 +204,32 @@ func (g *ShelleyGenesis) GenesisUtxos() ([]common.Utxo, error) {
 	return ret, nil
 }
 
+// GetInitialPools returns all initial stake pools with their delegators
+func (g *ShelleyGenesis) GetInitialPools() (map[string]GenesisPool, map[string][]string, error) {
```
agaffney, line 208: Drop the `Get` prefix on the function. It's not idiomatic Go

ledger/shelley/genesis.go
```diff
@@ -128,6 +204,32 @@ func (g *ShelleyGenesis) GenesisUtxos() ([]common.Utxo, error) {
...
+	poolStake := make(map[string][]string)
+	for stakeAddr, poolId := range g.Staking.Stake {
+		poolStake[poolId] = append(poolStake[poolId], stakeAddr)
+	}
+	return g.Staking.Pools, poolStake, nil
+}
+
+// GetPoolById returns a specific pool by its ID along with its delegators
+func (g *ShelleyGenesis) GetPoolById(poolId string) (*GenesisPool, []string, error) {
```
agaffney, line 217: Drop the `Get` prefix on the function. It's not idiomatic Go

ledger/shelley/genesis.go
agaffney, line 208: We should return native ledger types here like we do above for `GenesisUtxos()`. At minimum, we should use `common.Address` where appropriate.

ledger/shelley/genesis.go
```diff
@@ -42,7 +43,40 @@ type ShelleyGenesis struct {
...
+	DNSName string `json:"dnsName"`
+	Port    int    `json:"port"`
+}
+
+type GenesisReward struct {
+	Credential struct {
+		KeyHash string `json:"key hash"`
+	} `json:"credential"`
+	Network string `json:"network"`
 }
```
agaffney, line 80: There's a lot of overlap between these structures and `common.PoolRegistrationCertificate`, `common.PoolRelay`, and related types. We could probably merge them and avoid the need to convert/copy between them later.

ledger/shelley/genesis.go
agaffney, line 217: This should return `[]common.Address` instead of `[]string`

agaffney (changes requested):

agaffney (changes requested):

ledger/shelley/genesis.go
```diff
@@ -128,6 +232,111 @@ func (g *ShelleyGenesis) GenesisUtxos() ([]common.Utxo, error) {
...
+			if len(stakeAddr) != 56 {
+				return nil, nil, errors.New("invalid stake address length")
+			}
+
+			stakeKeyBytes, err := hex.DecodeString(stakeAddr)
+			if err != nil {
+				return nil, nil, errors.New("failed to decode stake key")
+			}
+
+			stakeAddrBytes := append([]byte{0xE1}, stakeKeyBytes...)
```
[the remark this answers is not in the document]
jkawan: have addressed the comments, could you please review again ?

ledger/common/address.go
```diff
@@ -107,24 +107,47 @@ func NewAddressFromParts(
...
-		stakingAddress: stakingAddr[:],
-	}, nil
+
+	header := addrType<<4 | networkId
+	addrBytes := append([]byte{header}, paymentAddr...)
+	if len(stakingAddr) > 0 {
+		addrBytes = append(addrBytes, stakingAddr...)
+	}
+
+	return NewAddressFromBytes(addrBytes)
```
agaffney, line 150: There's no need to do the round-trip through byte. Use the method that we already had to build the new `Address` and return it.

ledger/common/address.go
```diff
@@ -107,24 +107,47 @@ func NewAddressFromParts(
...
+	if addrType == AddressTypeNoneKey || addrType == AddressTypeNoneScript {
+		if len(paymentAddr) > 0 {
+			return Address{}, errors.New("payment address must be empty for stake-only addresses")
+		}
+		if len(stakingAddr) != AddressHashSize {
+			return Address{}, fmt.Errorf("staking key must be exactly %d bytes", AddressHashSize)
+		}
+
+		if addrType == AddressTypeNoneScript && networkId == AddressNetworkTestnet {
+			header := byte(0xF1)
```
agaffney, line 125: We don't want to hard-code magic values like this

ledger/common/address.go
```diff
@@ -107,24 +107,47 @@ func NewAddressFromParts(
...
+			addrBytes := append([]byte{header}, stakingAddr...)
+			return NewAddressFromBytes(addrBytes)
+		}
+
+		header := addrType<<4 | networkId
+		addrBytes := append([]byte{header}, stakingAddr...)
+		return NewAddressFromBytes(addrBytes)
```
agaffney, line 132: Use the method that we already had for building a new `Address` rather than round-tripping it through bytes

agaffney (changes requested):

ledger/common/address.go
```diff
@@ -419,14 +436,22 @@ func (a Address) Bytes() ([]byte, error) {
 		}
 		return ret, nil
 	}
+
+	header := (a.addressType << 4) | (a.networkId & AddressHeaderNetworkMask)
+
+	if a.addressType == AddressTypeNoneScript && a.networkId == AddressNetworkTestnet {
+		header = (AddressTypeNoneScript << 4) | 1
+	}
```
agaffney, line 444: I'm not sure what's happening here. This seems like a very specific circumstance that we could probably account for in other ways

ledger/shelley/genesis_test.go
```diff
@@ -250,3 +254,161 @@ func TestGenesisUtxos(t *testing.T) {
...
+				t.Errorf("Expected header byte %x, got %x",
+					expectedTestnetScriptStakeHeader, addrBytes[0])
+			}
+
+			// Verify stake key matches
+			stakeKey := hex.EncodeToString(addrBytes[1:])
+			if stakeKey != expectedStakeKey {
+				t.Errorf("Delegator key mismatch:\nExpected: %s\nActual:   %s",
+					expectedStakeKey, stakeKey)
+			}
```
agaffney, line 355: Instead of operating on address bytes above, use the `.NetworkId()` and `.Type()` functions from the `Address` type to verify the correct header flags, and the `.StakeKeyHash()` function to get the stake key for comparison.

ledger/shelley/genesis_test.go
```diff
@@ -250,3 +254,161 @@ func TestGenesisUtxos(t *testing.T) {
...
```
agaffney, line 405: Same as above...don't use address bytes to inspect it

agaffney (changes requested):

ledger/common/address.go
jkawan: eliminated this , it was for a test only.

agaffney (approved):

PR #1056, opened 2025-07-07
test(ledger): added blocks required for testing for all the eras except Conway

### LINKED ISSUE #1025: Collect blocks from each era for tests
We have various needs in the tests for decoding blocks of various eras. Grab the CBOR for a "representative block" from each era from mainnet and add a basic test for each era for decoding the CBOR.

jkawan:
1. Collected blocks for all the eras (except Conway) using epoch, hash and slot number. And also used https://cexplorer.io/
2. Have added the blocks in params_test.go file (except byron) for now but can move it around to any file as required by tests later

Closes #1025


agaffney (changes requested):
The block hex strings should go in a new `block_test.go` for each era rather than in an existing unrelated test file.

wolf31o2 (approved):

Issue #1112, opened 2025-07-21
Don't return an error on connection close when all protocols are shut down


agaffney:
When an incoming client explicitly stops the chain-sync, block-fetch, and tx-submission protocols (by sending their "done" messages) before closing the connection, we should not return an error to the consumer. We should treat it much the same as if the consumer had called `.Close()`. This should also probably apply to when the protocols have been started but no messages have been sent yet.

PR #1141, opened 2025-08-17
Not returning on connection close for chainsync, block-fetch and tx-submission protocol

jkawan:
1. Added changes to not return error while closing the connection for chainsync, block-fetch and tx-submission protocol- made changes in client,server and respective protocol files.
2. Added unit tests to test the change.

### DIFF
--- a/protocol/blockfetch/client.go
+++ b/protocol/blockfetch/client.go
@@ -35,6 +35,8 @@ type Client struct {
 	blockUseCallback     bool
 	onceStart            sync.Once
 	onceStop             sync.Once
+	currentState         protocol.State
+	stateMutex           sync.Mutex
 }
 
 func NewClient(protoOptions protocol.ProtocolOptions, cfg *Config) *Client {
@@ -46,6 +48,7 @@ func NewClient(protoOptions protocol.ProtocolOptions, cfg *Config) *Client {
 		config:               cfg,
 		blockChan:            make(chan ledger.Block),
 		startBatchResultChan: make(chan error),
+		currentState:         StateIdle,
 	}
 	c.callbackContext = CallbackContext{
 		Client:       c,
@@ -82,6 +85,18 @@ func NewClient(protoOptions protocol.ProtocolOptions, cfg *Config) *Client {
 	return c
 }
 
+func (c *Client) IsDone() bool {
+	c.stateMutex.Lock()
+	defer c.stateMutex.Unlock()
+	return c.currentState.Id == StateDone.Id
+}
+
+func (c *Client) setState(newState protocol.State) {
+	c.stateMutex.Lock()
+	defer c.stateMutex.Unlock()
+	c.currentState = newState
+}
+
 func (c *Client) Start() {
 	c.onceStart.Do(func() {
 		c.Protocol.Logger().
@@ -110,7 +125,11 @@ func (c *Client) Stop() error {
 				"connection_id", c.callbackContext.ConnectionId.String(),
 			)
 		msg := NewMsgClientDone()
-		err = c.SendMessage(msg)
+		if sendErr := c.SendMessage(msg); sendErr != nil {
+			err = sendErr
+			return
+		}
+		c.setState(StateDone)
 	})
 	return err
 }
@@ -196,6 +215,8 @@ func (c *Client) messageHandler(msg protocol.Message) error {
 		err = c.handleBlock(msg)
 	case MessageTypeBatchDone:
 		err = c.handleBatchDone()
+	case MessageTypeClientDone:
+		c.setState(StateDone)
 	default:
 		err = fmt.Errorf(
 			"%s: received unexpected message type %d",
--- a/protocol/chainsync/client.go
+++ b/protocol/chainsync/client.go
@@ -45,6 +45,8 @@ type Client struct {
 	wantCurrentTipChan       chan chan<- Tip
 	wantFirstBlockChan       chan chan<- clientPointResult
 	wantIntersectFoundChan   chan chan<- clientPointResult
+	stateMutex               sync.Mutex
+	currentState             protocol.State
 }
 
 type clientPointResult struct {
@@ -79,6 +81,7 @@ func NewClient(
 		wantCurrentTipChan:       make(chan chan<- Tip, 1),
 		wantFirstBlockChan:       make(chan chan<- clientPointResult, 1),
 		wantIntersectFoundChan:   make(chan chan<- clientPointResult, 1),
+		currentState:             stateIdle,
 	}
 	c.callbackContext = CallbackContext{
 		Client:       c,
@@ -116,6 +119,19 @@ func NewClient(
 	}
 	c.Protocol = protocol.New(protoConfig)
 	return c
+}
+
+// Add state tracking methods
+func (c *Client) setState(newState protocol.State) {
+	c.stateMutex.Lock()
+	defer c.stateMutex.Unlock()
+	c.currentState = newState
+}
+
+func (c *Client) IsDone() bool {
+	c.stateMutex.Lock()
+	defer c.stateMutex.Unlock()
+	return c.currentState.Id == stateDone.Id
 }
 
 func (c *Client) Start() {
@@ -570,12 +586,13 @@ func (c *Client) messageHandler(msg protocol.Message) error {
 		err = c.handleIntersectFound(msg)
 	case MessageTypeIntersectNotFound:
 		err = c.handleIntersectNotFound(msg)
+	case MessageTypeDone:
+		c.setState(stateDone)
 	default:
 		err = fmt.Errorf(
 			"%s: received unexpected message type %d",
 			ProtocolName,
-			msg.Type(),
-		)
+			msg.Type())
 	}
 	return err
 }
--- a/protocol/txsubmission/client.go
+++ b/protocol/txsubmission/client.go
@@ -28,6 +28,8 @@ type Client struct {
 	config          *Config
 	callbackContext CallbackContext
 	onceInit        sync.Once
+	stateMutex      sync.Mutex
+	currentState    protocol.State
 }
 
 // NewClient returns a new TxSubmission client object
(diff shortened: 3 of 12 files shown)


jkawan:
> 1. Added changes to not return error while closing the connection for chainsync, block-fetch and tx-submission protocol- made changes in client,server and respective protocol files.
> 2. Added unit tests to test the change.

Closes #1112

protocol/blockfetch/client.go
```diff
@@ -82,6 +85,18 @@ func NewClient(protoOptions protocol.ProtocolOptions, cfg *Config) *Client {
...
+	c.stateMutex.Lock()
+	defer c.stateMutex.Unlock()
+	return c.currentState.Id == StateDone.Id
+}
+
+func (c *Client) setState(newState protocol.State) {
+	c.stateMutex.Lock()
+	defer c.stateMutex.Unlock()
+	c.currentState = newState
+}
```
agaffney, line 98: The `protocol.Protocol` base type already tracks the current state. It should be possible to add `IsDone()` there in a generic way by checking that the current state matches the initial state and/or the current state having `AgencyNone`.

agaffney (changes requested):

protocol/blockfetch/client.go
jkawan: updated, could you please review when you get a chance ?

connection.go
```diff
@@ -250,6 +251,49 @@ func (c *Connection) shutdown() {
 	close(c.errorChan)
 }
 
+// isConnectionReset checks if an error is a connection reset error
+func (c *Connection) isConnectionReset(err error) bool {
+	errStr := err.Error()
+	return strings.Contains(errStr, "connection reset") ||
+		strings.Contains(errStr, "broken pipe")
+}
```
agaffney, line 259: We should check for error types rather than substring matches of the `Error()` output. Do we actually get any "connection closed" error that isn't `io.EOF`?

connection.go
```diff
@@ -250,6 +251,49 @@ func (c *Connection) shutdown() {
...
+
+// checkProtocols checks if the protocols are explicitly stopped by the client- treat as normal connection closure
+func (c *Connection) checkProtocols() bool {
+	// Check chain-sync protocol
+	if c.chainSync != nil && (!c.chainSync.Client.IsDone() || !c.chainSync.Server.IsDone()) {
```
agaffney, line 264: `.Client` or `.Server` can be nil, so we need `nil` checks against both

connection.go
```diff
@@ -250,6 +251,49 @@ func (c *Connection) shutdown() {
...
+		return false
+	}
+
+	// Check block-fetch protocol
+	if c.blockFetch != nil && (!c.blockFetch.Client.IsDone() || !c.blockFetch.Server.IsDone()) {
```
agaffney, line 269: `.Client` or `.Server` can be nil, so we need `nil` checks against both

connection.go
```diff
@@ -250,6 +251,49 @@ func (c *Connection) shutdown() {
...
+		return false
+	}
+
+	// Check tx-submission protocol
+	if c.txSubmission != nil && (!c.txSubmission.Client.IsDone() || !c.txSubmission.Server.IsDone()) {
```
agaffney, line 274: `.Client` or `.Server` can be nil, so we need `nil` checks against both

connection.go
```diff
@@ -250,6 +251,49 @@ func (c *Connection) shutdown() {
...
+		return false
+	}
```
agaffney, line 276: Let's also add `LocalStateQuery`, `LocalTxMonitor`, and `LocalTxSubmission` to this. It should be safe to do the NtN and NtC protocols all together with the `nil` checks.

connection.go
```diff
@@ -250,6 +251,49 @@ func (c *Connection) shutdown() {
...
 
+// isConnectionReset checks if an error is a connection reset error
+func (c *Connection) isConnectionReset(err error) bool {
+	errStr := err.Error()
+	return strings.Contains(errStr, "connection reset") ||
+		strings.Contains(errStr, "broken pipe")
+}
+
+// checkProtocols checks if the protocols are explicitly stopped by the client- treat as normal connection closure
+func (c *Connection) checkProtocols() bool {
```
agaffney, line 262: Rename this to `checkProtocolsDone`

connection.go
```diff
@@ -250,6 +251,49 @@ func (c *Connection) shutdown() {
...
+		return nil
+	}
+
+	if c.checkProtocols() {
+		return nil
+	}
+
+	if errors.Is(err, io.EOF) || c.isConnectionReset(err) {
+		return err
+	}
```
agaffney, line 293: This block doesn't seem to be useful, since we unconditionally `return err` right below it

connection_test.go
```diff
@@ -82,3 +83,193 @@ func TestDoubleClose(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("unexpected error when creating Connection object: %s", err)
+	}
+
+	// Test through HandleConnectionError - should return error when protocols are active
+	testErr := io.EOF
+	err = oConn.HandleConnectionError(testErr)
+	if err != testErr {
+		t.Fatalf("expected original error when protocols are active, got: %v", err)
+	}
```
agaffney, line 114: This isn't really testing anything useful. Instead, you should start one of the protocols and then close the connection (both from the "remote" side) and see if you get the expected error(s).

connection_test.go
```diff
@@ -82,3 +83,193 @@ func TestDoubleClose(t *testing.T) {
...
+	)
+	if err != nil {
+		t.Fatalf("unexpected error when creating Connection object: %s", err)
+	}
+
+	// Test that the implementation correctly handles the requirement
+	t.Log("Testing that connection errors are ignored when main protocols (chain-sync, block-fetch, tx-submission) are explicitly stopped by client")
+
+	oConn.Close()
+}
```
agaffney, line 145: You have nothing to capture any async error that might be generated by the `Connection` to determine if the connection error is ignored

connection.go
```diff
@@ -250,6 +251,49 @@ func (c *Connection) shutdown() {
...
+	// Check tx-submission protocol
+	if c.txSubmission != nil && (!c.txSubmission.Client.IsDone() || !c.txSubmission.Server.IsDone()) {
+		return false
+	}
+
+	return true
+}
+
+// HandleConnectionError handles connection-level errors centrally
+func (c *Connection) HandleConnectionError(err error) error {
```
agaffney, line 282: This should really be an unexported function. It looks like it's only public to allow using it from the tests, which appears to be a misuse of this function anyway.

protocol/blockfetch/client.go
```diff
@@ -196,6 +200,8 @@ func (c *Client) messageHandler(msg protocol.Message) error {
 		err = c.handleBlock(msg)
 	case MessageTypeBatchDone:
 		err = c.handleBatchDone()
+	case MessageTypeClientDone:
+		return nil
```
agaffney, line 204: This doesn't belong here. The client will never receive the `ClientDone` message, only the server.

protocol/chainsync/client.go
```diff
@@ -570,6 +572,8 @@ func (c *Client) messageHandler(msg protocol.Message) error {
 		err = c.handleIntersectFound(msg)
 	case MessageTypeIntersectNotFound:
 		err = c.handleIntersectNotFound(msg)
+	case MessageTypeDone:
+		return nil
```
agaffney, line 576: This doesn't belong here. The client will never receive this message, only the server

protocol/protocol.go
```diff
@@ -126,6 +127,34 @@ func New(config ProtocolConfig) *Protocol {
...
+	}
+	return <-stateChan
+}
+
+// IsDone checks if the protocol is in a done/completed state
+func (p *Protocol) IsDone() bool {
+	currentState := p.CurrentState()
+	if entry, exists := p.config.StateMap[currentState]; exists {
+		return entry.Agency == AgencyNone
+	}
```
agaffney, line 144: This should also return `true` if the current state is the same as the configured initial protocol state.

protocol/txsubmission/client.go
```diff
@@ -86,12 +86,15 @@ func (c *Client) Init() {
...
 		Debug(fmt.Sprintf("%s: client message for %+v", ProtocolName, c.callbackContext.ConnectionId.RemoteAddr))
+
 	var err error
 	switch msg.Type() {
 	case MessageTypeRequestTxIds:
 		err = c.handleRequestTxIds(msg)
 	case MessageTypeRequestTxs:
 		err = c.handleRequestTxs(msg)
+	case MessageTypeDone:
+		return nil
```
agaffney, line 97: This isn't needed. The client will never receive this message, only the server

protocol/protocol.go
```diff
@@ -126,6 +127,34 @@ func New(config ProtocolConfig) *Protocol {
...
 }
 
+// CurrentState returns the current protocol state
+func (p *Protocol) CurrentState() State {
+	stateChan := make(chan State, 1)
+	p.stateTransitionChan <- protocolStateTransition{
+		stateRespChan: stateChan,
+	}
+	return <-stateChan
+}
```
agaffney, line 137: There's probably a cleaner way to do this. Instead, let's make `currentState` defined at the top of `stateLoop()` a field in the `Protocol` type, and then we can access it more directly.

agaffney (changes requested):

wolf31o2:
@coderabbitai review

agaffney (changes requested):

wolf31o2:
I'm not sure if it's helpful, but I have some (abandoned) code on a similar path here: https://github.com/blinklabs-io/gouroboros/pull/1244

protocol/protocol.go
```diff
@@ -132,6 +135,36 @@ func New(config ProtocolConfig) *Protocol {
 	return p
 }
 
+// CurrentState returns the current protocol state
+func (p *Protocol) CurrentState() State {
+	p.currentStateMu.RLock()
+	defer p.currentStateMu.RUnlock()
+	return p.currentState
+}
```
agaffney: There is already a `getCurrentState()` function that does this

connection.go
```diff
@@ -318,6 +324,7 @@ func (c *Connection) setupConnection() error {
 		Muxer:        c.muxer,
 		Logger:       c.logger,
 		ErrorChan:    c.protoErrorChan,
+		Context:      c.ctx,
```
agaffney: This doesn't seem to be used anywhere

wolf31o2:
@cubic-dev-ai review this PR

PR #1172, opened 2025-09-01
feat(ledger): Deposit should be collected when pool is retired and re-registered #640

arepala-uml:
1. Added PoolState interface type in ledger/common/state.go where it helps for querying the current pool state.
2. Modified all the rules.go files of all eras for replacing PoolRegistration() to PoolCurrentState.
3. Remove PoolRegistration() function from CertState.
4. Tested by pointing dingo to a local gouroboros, disabled check for TX validation threshold & ran dingo to sync "preview" network. I did not observe any errors no “value not conserved” failures with a 500A (500,000,000 lovelace) delta.

### DIFF
--- a/ledger/common/state.go
+++ b/ledger/common/state.go
@@ -28,13 +28,18 @@ type UtxoState interface {
 // CertState defines the interface for querying the certificate state
 type CertState interface {
 	StakeRegistration([]byte) ([]StakeRegistrationCertificate, error)
-	PoolRegistration([]byte) ([]PoolRegistrationCertificate, error)
+}
+
+// PoolState defines the interface for querying the current pool state
+type PoolState interface {
+	PoolCurrentState([]byte) (*PoolRegistrationCertificate, *uint64, error)
 }
 
 // LedgerState defines the interface for querying the ledger
 type LedgerState interface {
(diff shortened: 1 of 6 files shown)


ledger/common/state.go
```diff
@@ -28,13 +28,18 @@ type UtxoState interface {
 // CertState defines the interface for querying the certificate state
 type CertState interface {
 	StakeRegistration([]byte) ([]StakeRegistrationCertificate, error)
-	PoolRegistration([]byte) ([]PoolRegistrationCertificate, error)
+}
+
+// PoolState defines the interface for querying the current pool state
+type PoolState interface {
+	PoolCurrentState([]byte) (*PoolRegistrationCertificate, *uint64, error)
```
wolf31o2, line 35: This is pretty opaque. Comment what the return values represent.

wolf31o2 (changes requested):

ledger/common/state.go
arepala-uml: This is the PR in dingo https://github.com/blinklabs-io/dingo/pull/885 and where I have defined of this PoolCurrentState clearly

PR #1200, opened 2025-10-02
fix(ledger): use transactions for leios endorser blocks

wolf31o2:
Since Leios Endorser Blocks do not have the transaction contents in their CBOR representation, we do not have a container for them. Add `transactions` to allow us to store them in the block once resolved for later retrieval, such as via the `Transactions()` func.

### DIFF
--- a/ledger/leios/leios.go
+++ b/ledger/leios/leios.go
@@ -64,7 +64,8 @@ type LeiosEndorserBlock struct {
 	cbor.DecodeStoreCbor
 	cbor.StructAsArray
 	hash         *common.Blake2b256
-	TxReferences []common.TxReference
+	transactions []common.Transaction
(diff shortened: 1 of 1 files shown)


ledger/leios/leios.go
```diff
@@ -64,7 +64,8 @@ type LeiosEndorserBlock struct {
 	cbor.DecodeStoreCbor
 	cbor.StructAsArray
 	hash         *common.Blake2b256
-	TxReferences []common.TxReference
+	transactions []common.Transaction
+	TxReferences []common.TxReference `cbor:"0,keyasint"`
```
agaffney, line 68: This annotation doesn't make sense with `cbor.StructAsArray`, since it's only for a map

agaffney (approved):

PR #1248, opened 2025-11-16
fix(protocols): guard against empty leiosfetch messages

wolf31o2:
## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to block, block-transaction, and vote request handling to prevent sending nil/invalid responses. When an upstream callback supplies no response, the server now returns an error instead of proceeding, reducing the chance of invalid messages and improving stability and reliability during request processing.


agaffney (approved):

wolf31o2:
@coderabbitai review

PR #1255, opened 2025-11-18
test: create more benchmarks for serde

wolf31o2:
## Summary by CodeRabbit

* **Tests**
  * Added comprehensive performance benchmarks for CBOR encoding/decoding and block serialization/deserialization across all ledger eras (Byron, Shelley, Allegra, Mary, Alonzo, Babbage, Conway).
  * Added microbenchmarks for address and transaction (de)serialization paths.
  * Added targeted benchmarks for decode/encode hot paths.

* **New Features**
  * Added Mary-era transaction CBOR round-trip test and Mary transaction type validation.


wolf31o2:
These all came from the chain. If they don't decode, we have a problem. They're all known good.

PR #1317, opened 2025-12-07
fix(ledger): use raw bytes for metadata

wolf31o2:
---
## Summary by cubic
Preserve raw CBOR bytes for transaction metadata and for tx bodies/witness sets across eras to keep exact bytes and round-trip encoding. Metadata is decoded only when needed.

- **Bug Fixes**
  - Store metadata raw bytes and decode on demand where used (incl. tx array metadata and Leios).
  - Add TransactionMetadataSet.GetRawMetadata and use it in Leios Transactions().
  - Decode transactions as raw arrays across eras; preserve metadata bytes; parse TxIsValid where needed (Alonzo, Babbage, Conway).
  - Add MarshalCBOR and use cbor.RawMessage for metadata; Cbor() delegates to preserve original bytes.
  - On block unmarshal, set raw CBOR for each tx body and witness set using common.ExtractAndSetTransactionCbor (Shelley→Conway).
  - Expose Cbor() on TransactionMetadatum and export DecodeAuxiliaryDataToMetadata.

<sup>Written for commit d443a43e1e5574fafeb7fc32c2a3d43983ade4cb. Summary will update automatically on new commits.</sup>

## Summary by CodeRabbit

* **Refactor**
  * Transactions across eras now preserve original per-item CBOR and carry per-transaction metadata for reliable round‑tripping.

* **New Features**
  * Added public auxiliary-data decoder and metadatum CBOR accessor; transactions expose CBOR/encoding methods and pointer-based metadata accessors.
  * Block decoding now captures per-transaction body/witness CBOR for accurate reconstruction.

* **Bug Fixes**
  * Stronger validation of transaction CBOR structure and clearer error reporting for missing or malformed components.

### DIFF
--- a/ledger/conway/conway.go
+++ b/ledger/conway/conway.go
@@ -118,6 +118,17 @@ func (b *ConwayBlock) UnmarshalCBOR(cborData []byte) error {
 	b.TransactionMetadataSet = tmp.TransactionMetadataSet
 
 	b.SetCbor(cborData)
+
+	// Extract and store CBOR for each component
+	if err := common.ExtractAndSetTransactionCbor(
+		cborData,
+		func(i int, data []byte) { b.TransactionBodies[i].SetCbor(data) },
+		func(i int, data []byte) { b.TransactionWitnessSets[i].SetCbor(data) },
+		len(b.TransactionBodies),
+		len(b.TransactionWitnessSets),
+	); err != nil {
+		return err
+	}
 	return nil
 }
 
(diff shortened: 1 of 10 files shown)


ledger/shelley/shelley.go
```diff
@@ -555,16 +566,47 @@ type ShelleyTransaction struct {
...
+	}
+
+	// Decode witness set
+	if _, err := cbor.Decode([]byte(txArray[1]), &t.WitnessSet); err != nil {
+		return fmt.Errorf("failed to decode transaction witness set: %w", err)
+	}
+
+	// Handle metadata (component 3)
+	// DecodeAuxiliaryDataToMetadata already preserves raw bytes via DecodeMetadatumRaw
+	if len(txArray) > 2 && len(txArray[2]) > 0 {
```
agaffney: The TX length is not variable. The last element will always exist, and will either contain metadata or a, CBOR `nil`

ledger/conway/conway.go
```diff
@@ -601,16 +607,52 @@ type ConwayTransaction struct {
...
+	}
+
+	// Decode TxIsValid flag
+	if _, err := cbor.Decode([]byte(txArray[2]), &t.TxIsValid); err != nil {
+		return fmt.Errorf("failed to decode TxIsValid: %w", err)
+	}
+
+	// Handle metadata (component 4)
+	// DecodeAuxiliaryDataToMetadata already preserves raw bytes via DecodeMetadatumRaw
+	if len(txArray) > 3 && len(txArray[3]) > 0 {
```
agaffney: The TX length is not variable. The last element will always exist, and will either contain metadata or a, CBOR `nil`

agaffney (dismissed):
This all technically works, but it seems to be doing a lot of unnecessary things

agaffney (approved):

PR #1350, opened 2025-12-26
feat: update local state query result types with proper field mappings

cryptodj413:
- RewardProvenanceResult
- RewardInfoPoolsResult
- DebugEpochStateResult
- DebugNewEpochStateResult
- DebugChainDepStateResult

---
## Summary by cubic
Set concrete types for several Local State Query results to improve CBOR decoding and type safety. Adds field mappings for reward provenance and pool info; debug state results now use raw CBOR.

- **Refactors**
  - RewardProvenanceResult -> struct with mapped fields (e.g., EpochLength, PoolMints, MaxLovelaceSupply, TotalStake, TotalBlocks, RewardPot, ActiveStake, etc.).
  - RewardInfoPoolsResult -> struct mapping per-pool metrics: Stake, OwnerPledge, OwnerStake, Cost, Margin, PerformanceEstimate.
  - DebugEpochStateResult, DebugNewEpochStateResult, DebugChainDepStateResult -> cbor.RawMessage to defer decoding.

<sup>Written for commit cd48e77c2374e64924c5c61e8d45a47da4e6f793. Summary will update automatically on new commits.</sup>

## Summary by CodeRabbit

* **Refactor**
  * Updated query result types to use explicit, strongly-typed structures instead of generic types, improving type safety and field-based handling for state query responses.

### DIFF
--- a/protocol/localstatequery/queries.go
+++ b/protocol/localstatequery/queries.go
@@ -597,8 +597,7 @@ func (u *UtxoId) MarshalCBOR() ([]byte, error) {
 	return cbor.Encode(tmpData)
 }
 
-// TODO (#863)
-type DebugEpochStateResult any
+type DebugEpochStateResult cbor.RawMessage
 
 // TODO (#858)
 /*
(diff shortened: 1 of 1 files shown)


protocol/localstatequery/queries.go
```diff
@@ -597,8 +597,7 @@ func (u *UtxoId) MarshalCBOR() ([]byte, error) {
 	return cbor.Encode(tmpData)
 }
 
-// TODO (#863)
-type DebugEpochStateResult any
+type DebugEpochStateResult cbor.RawMessage
```
wolf31o2: This isn't the actual shape, so we should not remove the TODO (or mark that issue closed) as this is functionally no different than any.

protocol/localstatequery/queries.go
```diff
@@ -654,34 +653,29 @@ type GenesisConfigResultProtocolParameters struct {
 	MinPoolCost           int
 }
 
-// TODO (#864)
-type DebugNewEpochStateResult any
+type DebugNewEpochStateResult cbor.RawMessage
 
-// TODO (#865)
-type DebugChainDepStateResult any
+type DebugChainDepStateResult cbor.RawMessage
```
wolf31o2: Do not remove the TODO here, either.

protocol/localstatequery/queries.go
```diff
@@ -654,34 +653,29 @@ type GenesisConfigResultProtocolParameters struct {
...
+	TotalStake            int64
+	TotalBlocks           int
+	Decentralization      any
+	AvailableBlockEntries int
+	SuccessRate           any
+	RewardPot             int64
+	DeltaT1               int64
+	ActiveStake           int64
+	Pools                 any
+	Desirabilities        any
```
wolf31o2: We want to avoid `any` here. The purpose of these issues is to fully realize the actual shapes so they can be fully decoded.

protocol/localstatequery/queries.go
```diff
@@ -654,34 +653,29 @@ type GenesisConfigResultProtocolParameters struct {
...
+}
```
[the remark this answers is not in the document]
wolf31o2: These are all valid concerns.

cryptodj413:
I’m going to close this PR and follow up with split PRs after further investigation.

PR #1360, opened 2026-01-02
fix(ledger): return hashes for inline datums

wolf31o2:
---
## Summary by cubic
Compute and return the Blake2b256 hash for inline datums, and fix utxorpc datumHash encoding to correctly distinguish no datum option vs an inline datum.

- **Bug Fixes**
  - DatumHash(): if no stored hash but inline datum data exists, compute and return its Blake2b256.
  - Utxorpc(): 32 zero bytes when there’s no DatumOption; use stored hash if present; compute and use the hash for inline datum; empty bytes only when DatumOption is present but empty.

<sup>Written for commit 0980676e45fc7ec6ebfd0e0e620ffc09a181ee41. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Made datum/hash handling consistent across transaction outputs, ensuring correct representation when datum is present, absent, or only data is provided.

* **Tests**
  * Added extensive tests for inline datums, including serialization/deserialization and round-trip validation of datum, hash, address, and amount.

### DIFF
--- a/ledger/babbage/babbage.go
+++ b/ledger/babbage/babbage.go
@@ -699,7 +699,13 @@ func (o BabbageTransactionOutput) Assets() *common.MultiAsset[common.MultiAssetT
 
 func (o BabbageTransactionOutput) DatumHash() *common.Blake2b256 {
 	if o.DatumOption != nil {
-		return o.DatumOption.hash
+		if o.DatumOption.hash != nil {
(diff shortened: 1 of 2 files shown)


wolf31o2:
@cubic-dev-ai review this PR

agaffney (approved):

PR #1613, opened 2026-03-01
fix(protocol): correct max pending bytes, split n2n/n2c

wolf31o2:
---
## Summary by cubic
Aligned protocol limits with the spec: set NtN pending message byte limit to 462KB and removed size/time limits for N2C. The reader now applies backpressure and blocks instead of disconnecting when the limit is reached.

- **Bug Fixes**
  - Set MaxPendingMessageBytes to 462000 for NtN (spec Table 3.15); updated docs and tests.
  - Split ChainSync state maps: NtN keeps limits/timeouts; NtC has none (spec §3.7.5).
  - readLoop now waits for pending bytes to drain, rejects single messages over the limit, and blocks queue writes with shutdown checks.
  - Added test to ensure N2C states have no byte limits.

<sup>Written for commit ab14d9200d79ae000e29ccc1fec823d6034098db. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced message backpressure handling to gracefully delay acceptance during high load instead of immediate protocol violations.

* **Chores**
  * Increased per-connection message buffer capacity to 462 KB, aligning with specification requirements.
  * Simplified protocol behavior by removing per-state message byte limits for client connections.

### DIFF
--- a/protocol/PROTOCOL_LIMITS.md
+++ b/protocol/PROTOCOL_LIMITS.md
@@ -17,7 +17,7 @@ All limits are based on the [Ouroboros Network Specification](https://ouroboros-
 - `MaxRecvQueueSize = 100` - Maximum size of the receive message queue
 - `DefaultPipelineLimit = 50` - Conservative default for pipeline limit
 - `DefaultRecvQueueSize = 50` - Conservative default for receive queue size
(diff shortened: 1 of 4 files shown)


wolf31o2:
@cubic-dev-ai review

PR #1785, opened 2026-05-27
feat: initial dijkstra era support

### LINKED ISSUE #1690: dijkstra: era type definitions and CDDL codec
## Summary

The Dijkstra hard fork introduces a new ledger era. gouroboros must add the era type definitions, CBOR codec, and any new mini-protocol message variants so downstream consumers (notably Dingo) can decode and validate Dijkstra-era blocks and transactions.

## What Needs to Happen

1. Define Dijkstra-era types in `ledger/dijkstra/` (transaction body, witness set, block, header, etc.) once the CDDL spec is published.
2. Implement CBOR encode/decode for all new Dijkstra types using the existing constructor/decoder pattern.
3. Add Dijkstra to the era type registry so block decoding dispatches the right era handler.
4. Add new mini-protocol message variants if the spec introduces any.
5. Add round-trip CBOR tests against any official conformance vectors that ship with the spec.

## Blocked By

- Dijkstra CDDL spec from IOG/Intersect

## Downstream Consumers

- Dingo Dijkstra ledger rule updates and HFC work depends on these types

## Roadmap

Q4 2026 — Dijkstra hard fork readiness.

wolf31o2:
Closes #1690 

---
## Summary by cubic
Adds initial Dijkstra era support across ledger, transactions, validation, diagnostics, and networking. Leios features are delivered within Dijkstra with shims for compatibility. Closes #1690.

- **New Features**
  - Added `ledger/dijkstra` (blocks, headers, transactions, protocol parameters, rules) with tests.
  - CBOR diagnostics recognize 5- or 7-field blocks; added `leios_cert` and `peras_cert` labels.
  - Plutus V4: `ScriptRef` type 4, `PlutusV4Script`, `AuxiliaryData.PlutusV4Scripts`, and a witness-set helper; validation and reference-script checks include V4.
  - Added guarding redeemer tag and native `RequireGuard` with credential-based checks.
  - Networking: enabled Dijkstra in `protocol/versions`, BlockFetch includes Dijkstra, and `protocol/localstatequery` returns Dijkstra protocol params; header/KES helpers and era ordering updated.

- **Refactors**
  - Leios is no longer a standalone era; added aliases in `ledger/leios` to Dijkstra and removed legacy Leios files/tests.
  - Routed block/tx constructors, era maps, and transaction-type detection to Dijkstra.
  - Fee-size rule subtracts 1 byte only for Alonzo–Conway when the 4-element envelope is present; script-purpose building and redeemer info hardened to avoid panics.

<sup>Written for commit 7c7d1b11850fe0ae57947d2e6d43d596d873b302. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/1785?utm_source=github">Review in cubic</a>

## Summary by CodeRabbit

* **New Features**
  * Full Dijkstra-era support (block/tx detection, decoding, validation) and Plutus V4 script execution; native-script guard/guarding redeemer flow.

* **Bug Fixes / Validation**
  * Strengthened validation: fees, collateral, sizes, cost-models, redeemer/script witness consistency, and block body shapes (5- or 7-field variants).

* **Chores / Removals**
  * Consolidated Leios into Dijkstra-era compatibility (legacy Leios artifacts removed).

* **Tests**
  * Added extensive Dijkstra and Plutus V4 validation tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/blinklabs-io/gouroboros/pull/1785?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


wolf31o2:
@cubic-dev-ai review

wolf31o2:
@cubic-dev-ai review

wolf31o2:
@cubic-dev-ai review

wolf31o2:
@cubic-dev-ai review

ledger/dijkstra/rules_test.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	err := UtxoValidatePlutusScripts(
+		tx,
+		0,
+		mockledger.NewLedgerStateBuilder().Build(),
+		&DijkstraProtocolParameters{},
+	)
+	require.ErrorAs(t, err, &common.MissingScriptWitnessesError{})
+}
```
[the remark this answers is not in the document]
wolf31o2: Not available yet, bro. https://github.com/blinklabs-io/ouroboros-mock/issues/208

cbor/cardano_diagnostic_test.go
```diff
@@ -57,9 +57,11 @@ func TestCardanoWitnessLabels(t *testing.T) {
 }
 
 func TestCardanoBlockLabels(t *testing.T) {
-	require.Len(t, cbor.CardanoBlockLabels, 5)
+	require.Len(t, cbor.CardanoBlockLabels, 7)
 	assert.Equal(t, "header", cbor.CardanoBlockLabels[0])
 	assert.Equal(t, "invalid_transactions", cbor.CardanoBlockLabels[4])
+	assert.Equal(t, "leios_cert", cbor.CardanoBlockLabels[5])
+	assert.Equal(t, "peras_cert", cbor.CardanoBlockLabels[6])
 }
```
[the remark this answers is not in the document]
wolf31o2: It, in fact, did not fail to compile.

wolf31o2:
This is based on this: https://github.com/IntersectMBO/cardano-ledger/blob/c47305fcf47bd77437b837d0dfb9cb4181bfbc77/eras/dijkstra/impl/cddl/data/dijkstra.cddl CDDL which is the one being maintained by the Leios team which includes Leios on top of Dijkstra

ledger/dijkstra/dijkstra.go
```diff
@@ -0,0 +1,1640 @@
...
+type DijkstraBlockHeader struct {
+	babbage.BabbageBlockHeader
+}
+
+func (h *DijkstraBlockHeader) UnmarshalCBOR(cborData []byte) error {
+	var tmp babbage.BabbageBlockHeader
+	if _, err := cbor.Decode(cborData, &tmp); err != nil {
+		return err
+	}
+	h.BabbageBlockHeader = tmp
```
chrisguiney, line 615: **Missing outer `SetCbor` — `Cbor()` will return nil on this type.**

`DijkstraBlockHeader.UnmarshalCBOR` decodes into a local `tmp` and assigns to the embedded `BabbageBlockHeader`. The inner `babbage.BabbageBlockHeader.UnmarshalCBOR` does call `SetCbor` on the body, which is why KES verification still finds the right bytes — but the **outer** `DijkstraBlockHeader`'s `DecodeStoreCbor` is never populated, so any caller using `DijkstraBlockHeader.Cbor()` (e.g. for hashing) gets `nil`.

This is exactly the failure mode CLAUDE.md flags as recurring (`DecodeStoreCbor requires a custom UnmarshalCBOR that calls SetCbor(cborData)`). Suggest:

```go
func (h *DijkstraBlockHeader) UnmarshalCBOR(cborData []byte) error {
    // ...existing decode into tmp...
    h.SetCbor(cborData)
    return nil
}
```

Or drop the wrapper entirely and let the promoted method dispatch — but then no future drift can silently re-introduce this gap.

ledger/dijkstra/dijkstra.go
```diff
@@ -0,0 +1,1640 @@
...
+	md, err := common.DecodeAuxiliaryDataToMetadata(raw)
+	if err == nil && md != nil {
+		*metadata = md
+		return nil
+	}
+	return errors.New("decode Dijkstra auxiliary data")
+}
+
+func isCborNull(raw cbor.RawMessage) bool {
+	return len(raw) == 0 || (len(raw) == 1 && raw[0] == 0xf6)
```
chrisguiney: **`isCborNull` treats empty slice as null.**

`len(raw) == 0` (zero-length slice) isn't a valid CBOR encoding, but here it's silently accepted as null. If a caller wires up an empty `RawMessage` the optional `leios_cert` / `peras_cert` field disappears rather than erroring. Suggest tightening to:

```go
return len(raw) == 1 && raw[0] == 0xf6
```

ledger/dijkstra/rules.go
```diff
@@ -0,0 +1,1385 @@
...
+
+func dijkstraHasGuardScriptAt(
+	tx common.Transaction,
+	index uint32,
+) bool {
+	dijkstraTx, ok := tx.(*DijkstraTransaction)
+	if !ok || dijkstraTx.Body.TxGuards == nil {
+		return false
+	}
+	guards := dijkstraTx.Body.TxGuards
```
chrisguiney, line 1324: **Native-script guards likely fail with `MissingScriptWitnessesError`.**

`dijkstraHasGuardScriptAt` treats any `CredentialTypeScriptHash` guard as Plutus-bearing. But a native-script `RequireGuard` credential is also a script hash — so it falls into the same branch, gets required to have a Plutus redeemer in `UtxoValidateExtraneousRedeemers` / `validateGuardingPlutusScripts`, and `availableScripts` (Plutus-only, see line ~286) won't contain the native script, so it fails as `MissingScriptWitnessesError`.

Either (a) disambiguate by walking witness scripts / reference scripts to determine whether the guard's credential resolves to a Plutus or native script, or (b) document explicitly that `RedeemerTagGuarding` is only valid for Plutus-backed guards and the native path is something else.

ledger/conway/rules.go
```diff
@@ -2165,6 +2155,37 @@ func UtxoValidatePlutusScripts(
...
+			if !txInfoV3Built {
+				var err error
+				txInfoV3, err = script.NewTxInfoV3FromTransaction(ls, tx, resolvedInputs)
+				if err != nil {
+					return ScriptContextConstructionError{Err: err}
+				}
+				txInfoV3Built = true
+			}
+			redeemer := script.Redeemer{
+				Tag:     redeemerKey.Tag,
```
chrisguiney: **`PlutusV4Script` case added to Conway dispatch — unreachable but unsafe if reached.**

This V4 branch lives inside Conway's `UtxoValidatePlutusScripts`. Conway's `ScriptRef` CBOR doesn't expose type 4, so it's unreachable from real Conway data, but if it's ever hit (e.g. malformed input, future refactor), `conwayPparams.CostModels[3]` is `nil` and gets passed straight to `cek.NewEvalContext` — silent pass at best, panic at worst.

Suggest either gating with an explicit "Dijkstra+ only" check that returns a clean era error, or removing the branch entirely until Conway actually needs it.

ledger/leios/leios.go
```diff
@@ -14,617 +14,40 @@
...
-
-	TxTypeLeios = 7
+	EraIdLeios           = dijkstra.EraIdDijkstra
+	EraNameLeios         = dijkstra.EraNameDijkstra
+	BlockTypeLeios       = dijkstra.BlockTypeDijkstra
+	BlockHeaderTypeLeios = dijkstra.BlockHeaderTypeDijkstra
+	TxTypeLeios          = dijkstra.TxTypeDijkstra
+
+	BlockTypeLeiosRanking  = dijkstra.BlockTypeDijkstra
+	BlockTypeLeiosEndorser = dijkstra.BlockTypeDijkstra
```
chrisguiney: **Alias collision will break downstream `switch` statements at compile time.**

On `main`, `BlockTypeLeiosRanking = 8` and `BlockTypeLeiosEndorser = 9` (distinct). After this PR they both equal `dijkstra.BlockTypeDijkstra` (= 8). Any downstream code with `case leios.BlockTypeLeiosRanking:` and `case leios.BlockTypeLeiosEndorser:` in the same `switch` will fail to compile with "duplicate case". Any logic that branched on the distinction is also silently wrong now.

Similarly `TxTypeLeios` previously had its own value but now equals `TxTypeDijkstra = 7` — anything dispatching on the old number will misroute.

Options: (a) call this out in release notes as an explicit breaking change for the shim, (b) drop the dual aliases here and leave only `BlockTypeLeios`, forcing callers to update, or (c) keep them numerically distinct in the shim and document the mismatch with the on-wire encoding.

ledger/leios/leios.go
```diff
@@ -14,617 +14,40 @@
...
-
-	BlockTypeLeiosRanking  = 8
-	BlockTypeLeiosEndorser = 9
-
-	BlockHeaderTypeLeios = 7
-
-	TxTypeLeios = 7
+	EraIdLeios           = dijkstra.EraIdDijkstra
+	EraNameLeios         = dijkstra.EraNameDijkstra
+	BlockTypeLeios       = dijkstra.BlockTypeDijkstra
```
chrisguiney: Nit: `BlockTypeLeios` is newly introduced (didn't exist on `main`) but no caller dispatches on it. If it's purely aspirational, consider dropping. If it's the intended forward-compatible alias, it might be the only constant worth keeping while `BlockTypeLeiosRanking` / `BlockTypeLeiosEndorser` get removed (related to the comment below).

ledger/verify_block.go
```diff
@@ -105,6 +138,8 @@ func DetermineBlockType(headerCbor []byte) (uint, error) {
 			return BlockTypeBabbage, nil
 		case ProtoMajorConway:
 			return BlockTypeConway, nil
```
chrisguiney, line 141: **Conway `ProtoMajor` 10 and 11 fall through to `default`.**

`conway.MaxProtocolVersionConway = 11` (`ledger/conway/conway.go:43`), but `DetermineBlockType` only matches `case ProtoMajorConway` (= 9). A Conway block carrying proto major 10 or 11 hits the `default` branch and returns "unknown proto major".

This is pre-existing, but it's directly adjacent to the new `case ProtoMajorDijkstra, ProtoMajorDijkstra + 1:` line, so it'd be cheap to fix here:

```go
case major >= ProtoMajorConway && major < ProtoMajorDijkstra:
    return BlockTypeConway, nil
```

(Or list 9, 10, 11 explicitly.)

protocol/versions.go
```diff
@@ -185,6 +187,7 @@ var protocolVersions = map[uint16]ProtocolVersion{
 		EnableAlonzoEra:              true,
 		EnableBabbageEra:             true,
 		EnableConwayEra:              true,
+		EnableDijkstraEra:            true,
```
chrisguiney, line 190: **NtC v20 is now dual-purpose; the comment above (line 180, not in diff) still says "additional Conway governance queries."**

NtN gets a clean v15 bump for Dijkstra (correct), but NtC v20 has Dijkstra bolted onto an existing entry that was added for Conway governance queries. Upstream IOG typically bumps NtC in lockstep with new eras (Conway got NtC 16).

Options:
1. Add a new NtC 21 for Dijkstra and update the NtC capability matrix accordingly.
2. Update the v20 comment to acknowledge dual purpose if this is intentional alignment with whatever the node ships.

As-is, this is a footgun for anyone matching/searching by the comment.

ledger/dijkstra/pparams.go
```diff
@@ -0,0 +1,306 @@
...
+	"github.com/blinklabs-io/gouroboros/cbor"
+	"github.com/blinklabs-io/gouroboros/ledger/common"
+	"github.com/blinklabs-io/gouroboros/ledger/conway"
+	utxorpc "github.com/utxorpc/go-codegen/utxorpc/v1alpha/cardano"
+)
+
+type DijkstraProtocolParameters struct {
+	conway.ConwayProtocolParameters
+	MaxRefScriptSizePerBlock uint32
+	MaxRefScriptSizePerTx    uint32
```
chrisguiney, line 30: Nit: `MaxRefScriptSizePerBlock` / `MaxRefScriptSizePerTx` are parsed, JSON-tagged, and CBOR-round-tripped, but no rule reads them. Acceptable for "initial" support, but worth a tracking issue or a `TODO` comment so it isn't forgotten when stress-testing reference-script-heavy blocks.

ledger/common/rules.go
```diff
@@ -92,7 +93,14 @@ func TxSizeForFee(tx Transaction) (int, error) {
 	}
 	fullSize := len(cborData)
 	if tx.Type() >= txTypeAlonzo {
-		return fullSize - 1, nil
+		var txArray []cbor.RawMessage
```
chrisguiney: Nit: `TxSizeForFee` now decodes the entire tx CBOR into `[]cbor.RawMessage` on every call just to count the array length. For large transactions this is wasteful — the major-type / array-length header byte(s) at the start would suffice. Era gate logic itself is correct: Dijkstra (Type=7) takes the array-length branch and the 3-element envelope returns `fullSize` without the `-1` adjustment, matching the doc comment.

chrisguiney (commented):
## Review summary

Substantial PR (+5.1k / -1.6k across 41 files) adding Dijkstra end-to-end (block/header/tx types, pparams, rules, diagnostics), Plutus V4 + `ScriptRef` type 4, the guarding redeemer tag + `RequireGuard` native script, and a Leios → Dijkstra collapse via type aliases. CI green (go-test 1.25/1.26, lint, nilaway, CodeQL).

**Architecture:** sound — Dijkstra correctly embeds Conway and overrides only era-specific behaviour, matching the existing Conway/Babbage delegation pattern. Era ordering / type-map symmetry / `init()` registration / `DecodeStoreCbor` + `SetCbor` are correct everywhere except `DijkstraBlockHeader` (see inline). Leios aliases use true type identity (`type X = dijkstra.Y`), so callers don't break.

**No blockers.** A handful of non-blocking items worth addressing before/shortly after merge — left inline. The most likely to bite downstream is the missing outer `SetCbor` on `DijkstraBlockHeader`, since it silently makes `Cbor()` return nil and breaks hashing (the recurring footgun called out in CLAUDE.md).

Other notes that didn't have a clean line to attach to:
- Plutus V4 wiring is comprehensive — added to `script.go`, both Conway & Dijkstra rules (cost-models-present, script-data-hash, redeemer/script witnesses), language-view encoding (versions `0,1,2,3`), witness-set helper, and `AlonzoAuxiliaryData` key 5.
- `RedeemerTagGuarding` is plumbed through redeemer.go, `BuildScriptPurpose` (returns nil for non-Dijkstra), Conway extraneous-redeemer rejection, and the Dijkstra-specific paths.
- Conway's `UtxoValidateWithdrawals` body and `NOTE:` markers are intact — none silently removed.
- KES path is a pure type rename; no crypto regression.
- Block body 5-or-7 field discrimination is unambiguous; 6-field warns correctly.
- Test coverage for `validateGuardingPlutusScripts` only exercises the `MissingScriptWitnessesError` path — happy-path / `PlutusScriptFailedError` paths aren't covered yet.

chrisguiney (approved):

PR #1791, opened 2026-06-01
fix: update rewards calculation to handle case when total pool rewards are less than the fixed fee

brouwerQ:
The current code only accounts for the case where `totalPoolRewards` equals `poolCost`. If `totalPoolRewards` is less than `poolCost`, the calculation of `stakeholderRewardsTotal` would cause an underflow, giving massive rewards to delegators.

I changed it that it immediately checks for this and returns with all rewards given to the operator, skipping the other unnecessary calculations. `DelegatorRewards` will be an empty map because of this, but this was already the case when `totalPoolRewards` equals `poolCost`, so I figured this behavior is fine.

As the check also contains the equals case, the check if `totalPoolRewards` is greater than `poolCost` later on can be skipped.

I wonder why this hasn't caused any issues yet, but maybe the situation doesn't occur yet on the testnet where Dingo is used? Setting a really high fixed fee for a pool should immediately cause an issue though...

On mannet this situation is already very common these days.

---
## Summary by cubic
Fix rewards calculation when total pool rewards are less than or equal to the fixed fee. All rewards now go to the operator in this case, preventing underflow and bogus delegator payouts.

- **Bug Fixes**
  - Early return when `totalPoolRewards <= poolCost` with operator-only payout and empty `DelegatorRewards`.
  - Removed redundant `> poolCost` check; retained owner stake/margin math for remaining-rewards path.
  - Prevents underflow in stakeholder totals that could inflate delegator rewards.

<sup>Written for commit 888f561b756ea5b9ab94839b299869a13e0ce163. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/1791?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reward distribution calculations to correctly handle scenarios where pool rewards are at or below operational costs, and refined stake-based allocation logic between operators and delegators.

### DIFF
--- a/ledger/common/rewards.go
+++ b/ledger/common/rewards.go
@@ -437,28 +437,36 @@ func distributePoolRewards(
 		totalPoolStake += stake
 	}
 
+	// Check if poolCost already uses up all the available rewards
+	if totalPoolRewards <= poolCost {
+		return &PoolRewards{
+			OperatorRewards:  totalPoolRewards,
+			DelegatorRewards: make(map[AddrKeyHash]uint64),
+			TotalRewards:     totalPoolRewards,
+		}
+	}
+
 	// Calculate operator (leader) reward
 	operatorRewards := poolCost
-	if totalPoolRewards > poolCost {
-		// Find owner stake (owners are also delegators in the snapshot)
-		ownerStake := uint64(0)
-		for _, owner := range poolParams.PoolOwners {
-			if stake, exists := delegatorStake[owner]; exists {
-				ownerStake += stake
-			}
+	
+	// Find owner stake (owners are also delegators in the snapshot)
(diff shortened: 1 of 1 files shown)


wolf31o2 (approved):
There's a bunch of extra whitespace triggering the linter, but I'll fix that after merge.

PR #1917, opened 2026-07-28
feat(byron): validate block body proofs at decode

chrisguiney:
## Problem

`NewByronMainBlockFromCbor` and `NewByronEpochBoundaryBlockFromCbor` accept a `common.VerifyConfig` and never read it — they decode and return. Shelley and later validate their body hash at decode; Byron validated nothing.

A Byron header, and therefore the block hash and slot, can be genuine while the body has been replaced. Callers accepting blocks from a source they do not control had no way to detect that.

Reproduced against the bundled mainnet Byron block: replacing the transaction payload with an empty one changes the encoding from 894 to 636 bytes while `Hash()` and `SlotNumber()` stay identical, and decoding succeeds. `TestByronMainBlockHeaderUnchangedByTampering` pins exactly this.

Found while hardening Bark archive block retrieval in dingo (blinklabs-io/dingo#2038, PR blinklabs-io/dingo#2990), where the archive chooses the response body and local verification has to establish block identity. Credit to @wolf31o2 for identifying the gap in review.

## Change

Recompute the header's body proof from the body:

- **Main blocks** — body proof is `[tx_proof, ssc_proof, dlg_proof, upd_proof]`, with `tx_proof = [tx_count, merkle_root, witnesses_hash]`. Validates the transaction count, the merkle root over transaction bodies, the hash over the witness list, and the delegation and update payload hashes.
- **Epoch boundary blocks** — no transactions, so a single hash covers the body. The EBB body now preserves its original CBOR, matching how the main block body already keeps `dlgPayloadRaw`/`updPayloadRaw`.

On by default, skipped via `SkipBodyHashValidation`, matching the other eras.

### Byron merkle roots

`byron.MerkleRoot` implements cardano-ledger's `Cardano.Chain.Common.Merkle`: leaves and branches carry distinct tag bytes (`0x00`/`0x01`), the empty tree hashes the empty string, and each node splits at the largest power of two below its item count — which makes odd trees left-heavy. Computed over preserved CBOR, never a re-encoding.

Two details that are easy to get wrong and are covered by tests:

- The **witness list is an indefinite-length array**. A definite-length array over identical elements hashes differently and would reject every real block.
- The **split point** produces a specific tree shape; a wrong one still yields a plausible-looking hash, so `TestMerkleRootShape` asserts the structure for 0–5 items by recomputing leaves and branches directly from the spec rather than by calling `MerkleRoot`.

## Tests

- `TestByronMainBlockBodyProofValidates` — the bundled mainnet block, which carries **two** transactions and so exercises merkle branch combination, not just the single-leaf case.
- `TestByronEpochBoundaryBlockBodyProofValidates` — a real testnet EBB.
- `TestByronMainBlockRejectsSubstitutedBody`, `TestByronEpochBoundaryBlockRejectsSubstitutedBody` — tamper regressions, asserting `ErrBodyProofMismatch`.
- `TestByronMainBlockSkipBodyHashValidation` — the escape hatch still decodes.
- `TestByronMainBlockHeaderUnchangedByTampering` — documents why the proof is needed: hash and slot survive a body substitution untouched.
- `TestMerkleRootShape`, `TestMerkleRootOrderSensitive`, `TestMerkleRootLeafBranchDomainSeparation`.

Validating against blocks produced by the reference implementation is what pins the recomputation — the real fixtures are the oracle, not our own encoder.

## Verification

- `make test` — 3544 passing, 0 failures
- `make lint` — 0 issues across all 9 targets
- `make format` — no changes to the added files
- Conformance suite passes, including the existing Byron EBB test that now decodes with validation enabled

## Note on the conformance table

`CLAUDE.md` asks that the conformance snapshot table be updated when counts change. I left it alone: my count of `^func Test` is 1500 against the table's recorded 3461, so the figures are derived by different methods and I would rather not overwrite the baseline with an incompatible number. Happy to update it if you tell me how that figure is produced.

---
## Summary by cubic
Validate Byron block body proofs during decode to bind the body to the header and reject tampered bodies from untrusted sources. Enabled by default; set `common.VerifyConfig{SkipBodyHashValidation: true}` to bypass (same as Shelley+).

- New Features
  - `NewByronMainBlockFromCbor` and `NewByronEpochBoundaryBlockFromCbor` now recompute and validate header body proofs at decode.
  - Main blocks: verify tx count, Merkle root over tx bodies, witnesses hash, and `dlg`/`upd` payload hashes; EBBs: verify a single body hash with body CBOR derived from stored block bytes. Adds `byron.MerkleRoot` matching cardano-ledger (tagged leaves/branches; split at largest power of two) over preserved CBOR.

- Bug Fixes
  - Require preserved witness CBOR; missing witnesses now fail validation instead of hashing an empty list. Note: `ssc_proof` is not recomputed; changes limited to the SSC payload aren’t detected.

<sup>Written for commit 0d463350d991b362bcf4eaa24f00c59e93658f8e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/1917?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added verification of Byron block body proofs, including transaction, delegation, update, and epoch-boundary proofs.
  * Added Byron-specific Merkle root calculation.
  * Added an option to skip body-hash validation when decoding blocks.

* **Bug Fixes**
  * Byron blocks with substituted or tampered bodies are now rejected when validation is enabled.
  * Preserved body data for accurate epoch-boundary verification.

* **Documentation**
  * Clarified body validation behavior, proof formats, and security considerations when validation is disabled.

### DIFF
--- a/ledger/byron/byron.go
+++ b/ledger/byron/byron.go
@@ -1196,9 +1196,17 @@ type ByronEpochBoundaryBlock struct {
 	BlockHeader *ByronEpochBoundaryBlockHeader
 	Body        []common.Blake2b224
 	Extra       []any
+	bodyRaw     []byte // Original CBOR for body proof validation
 }
 
 func (b *ByronEpochBoundaryBlock) UnmarshalCBOR(cborData []byte) error {
+	// Decode as raw parts first so the body's original CBOR is preserved.
+	// The header's proof is a hash of exactly these bytes, and re-encoding
+	// is not guaranteed to reproduce them.
+	var rawParts []cbor.RawMessage
+	if _, err := cbor.Decode(cborData, &rawParts); err != nil {
+		return err
+	}
 	type tByronEpochBoundaryBlock ByronEpochBoundaryBlock
 	var tmp tByronEpochBoundaryBlock
 	if _, err := cbor.Decode(cborData, &tmp); err != nil {
@@ -1208,8 +1216,17 @@ func (b *ByronEpochBoundaryBlock) UnmarshalCBOR(cborData []byte) error {
 		return errors.New("byron EBB block missing header")
 	}
 	*b = ByronEpochBoundaryBlock(tmp)
+	if len(rawParts) >= 2 {
+		b.bodyRaw = []byte(rawParts[1])
+	}
 	b.SetCbor(cborData)
 	return nil
+}
+
+// BodyCbor returns the original CBOR bytes of the epoch boundary block body.
+// This is used for body proof validation.
+func (b *ByronEpochBoundaryBlock) BodyCbor() []byte {
+	return b.bodyRaw
 }
 
 func (ByronEpochBoundaryBlock) Type() int {
@@ -1265,9 +1282,22 @@ func (b *ByronEpochBoundaryBlock) BlockBodyHash() common.Blake2b256 {
 func NewByronEpochBoundaryBlockFromCbor(
 	data []byte, config ...common.VerifyConfig,
 ) (*ByronEpochBoundaryBlock, error) {
+	var cfg common.VerifyConfig
+	if len(config) > 0 {
+		cfg = config[0]
+	}
+	// Default: validation enabled (SkipBodyHashValidation = false)
+
 	var byronEbbBlock ByronEpochBoundaryBlock
 	if _, err := cbor.Decode(data, &byronEbbBlock); err != nil {
 		return nil, fmt.Errorf("decode Byron EBB block error: %w", err)
+	}
+	// Bind the body to the header. Without this the header, and so the
+	// block hash, can be genuine while the body has been substituted.
+	if !cfg.SkipBodyHashValidation {
+		if err := byronEbbBlock.ValidateBodyProof(); err != nil {
+			return nil, err
+		}
 	}
 	return &byronEbbBlock, nil
 }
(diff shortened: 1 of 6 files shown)


chrisguiney:
Both addressed in the follow-up commit.

**SSC proof** — I tried to close this rather than document it, and the encoding does not line up. The mainnet block's SSC payload is `8203d9010280`, i.e. `[3, tag258([])]`. Hashing the payload part directly gives `25777aca…`, but the header records `d36a2619…`, so the proof is taken over cardano-ledger's own encoding of `VssCertificatesMap`, not the bytes carried inline in the block.

I could have kept guessing at the encoding, but the only fixture available has an *empty* SSC payload, so anything I reverse-engineered would be validated solely against the degenerate case and would give false assurance for non-empty payloads. That seemed worse than a documented gap, so I took your alternative: a `NOTE:` recording the encoding mismatch with the observed hashes, why it is not reproduced, and the consequence — an alteration confined to the SSC payload is not detected, while transaction contents remain fully covered by the tx proof.

`ARCHITECTURE.md` states the same limitation so it is not only visible at the call site. Happy to implement SSC coverage in a follow-up if someone can point at a non-empty SSC fixture to validate against.

**Witness CBOR emptiness** — fixed; a transaction with no preserved witness CBOR is now rejected the same way a missing body is, rather than hashing an empty witness list into the proof.

`make test`, `make lint` (0 issues across all 9 targets), and the conformance suite all still pass.

ledger/byron/byron.go
```diff
@@ -1196,9 +1196,17 @@ type ByronEpochBoundaryBlock struct {
 	BlockHeader *ByronEpochBoundaryBlockHeader
 	Body        []common.Blake2b224
 	Extra       []any
+	bodyRaw     []byte // Original CBOR for body proof validation
```
wolf31o2: DecodeStoreCbor will already have this

wolf31o2 (changes requested):

ledger/byron/byron.go
chrisguiney: You're right — `DecodeStoreCbor` retains the whole block, so `bodyRaw` held a second copy of bytes that were a slice away. Removed in 0d46335: `BodyCbor()` now reads the body back out of the stored block CBOR, and `UnmarshalCBOR` is back to its original shape.

The hashed bytes are still the originals rather than a re-encoding, which is what the proof needs. `TestByronEpochBoundaryBlockBodyProofValidates` passing against the real testnet EBB confirms the slice is the right one.

wolf31o2 (approved):

PR #1923, opened 2026-07-30
fix(ledger): preserve unknown error constructors instead of defaulting eras

### LINKED ISSUE #1868: ledger: preserve unknown error constructors instead of defaulting eras
## Problem

Ledger error decoding and encoding has several fallback paths that hide unknown era/constructor mismatches:

- `ledger/error.go`: UTxO failure constants fall back to Babbage for "other" eras
- `ledger/error.go`: UTXOW unknown eras fall back to Babbage decoding
- `ledger/error.go`: unknown UTxO failure IDs decode as `GenericError`
- `ledger/error.go`: some UTxO failure marshal paths default to Conway constants when `Type` is unset

This makes forward-incompatible or malformed failure constructors look successfully decoded, and can lose the explicit unknown tag/era context.

## Expected behavior

Unknown eras and unknown failure constructor tags should be represented explicitly or returned as typed decode errors. They should not silently select Babbage/Conway structures.

## Suggested fix

Consider adding strict unknown error types, for example:

- `UnknownApplyTxFailureError`
- `UnknownUtxowFailureError`
- `UnknownUtxoFailureError`

These should preserve raw CBOR, era id, and constructor tag. Marshal paths should require the era/type to be set when the constructor number is era-dependent.

## Tests

Add tests covering:

- unknown Shelley-based era in `UtxowFailure`
- unknown UTxO failure constructor id
- `ScriptsNotPaidUtxo` / `CollateralContainsNonADA` marshal with unset `Type`
- no silent fallback to Babbage or Conway constants

cryptodj413:
## Summary

Unrecognized UTXO/UTXOW/LEDGER failure constructors were previously decoded
by silently guessing another era's constructor numbering (or falling back
to an opaque `GenericError`), which could misdecode the failure payload
since eras such as Conway completely renumbered these tags.

- Added three new error types, each carrying `Era uint8`, `FailureType int`,
  and the raw `Cbor []byte`, following the file's existing plain-field
  `GenericError` convention:
  - `UnknownApplyTxFailureError`
  - `UnknownUtxowFailureError`
  - `UnknownUtxoFailureError`
- `getEraSpecificUtxoFailureConstants` now returns an explicit error for
  unrecognized eras instead of silently falling back to Babbage's
  constructor numbering.
- Updated decode paths to build the new typed `Unknown*` errors on an
  unrecognized era or constructor tag, instead of guessing another era's
  numbering or falling back to `GenericError`:
  - `UtxoFailure.UnmarshalCBOR`
  - `UtxowFailure.UnmarshalCBOR` and its era-specific
    Shelley/Alonzo/Babbage/Conway `UnmarshalCBOR` methods
  - `ApplyTxError.UnmarshalCBOR`
- Hardened `ScriptsNotPaidUtxo.MarshalCBOR` and
  `CollateralContainsNonADA.MarshalCBOR` to require `Type` be explicitly
  set to a recognized era constructor, returning an error instead of
  silently defaulting to Conway's numbering when unset.
- As a side benefit, fixed a pre-existing bug in `ApplyTxError.UnmarshalCBOR`'s
  fallback path where the wrong CBOR slice was captured (with panic risk on
  short arrays); it now correctly captures the current failure item.

Fixes #1868

## Test plan

- `go build ./...` and `go vet ./...` clean
- `go test -race -count=1 ./ledger/...` all pass
- `golangci-lint run ./ledger/...` — 0 issues (two justified
  `//nolint:nilerr` added where an unknown-era/tag result is intentionally
  converted into a typed value rather than propagated as a decode error)
- `gofmt -l` clean
- Added test coverage for unknown-era and unknown-constructor-tag paths
  across `UtxoFailure`, `UtxowFailure` (all eras), and `ApplyTxError`, plus
  the new `MarshalCBOR` validation for `ScriptsNotPaidUtxo` and
  `CollateralContainsNonADA`
- Conformance suite remains at 315/315 (no regressions)

---
## Summary by cubic
Preserves unknown UTXO/UTXOW/LEDGER failure constructors by keeping their era, tag, and raw CBOR, fixes decode edge cases, and completes Dijkstra support with a dedicated UTXO map and UTXOW constructors 19–20. Also tightens era scoping and validation to prevent misdecodes (fixes #1868).

- **Bug Fixes**
  - Added UnknownApplyTxFailureError, UnknownUtxowFailureError, and UnknownUtxoFailureError (carry Era, FailureType, Cbor); decoding now emits these instead of guessing eras or using GenericError.
  - getEraSpecificUtxoFailureConstants now errors on unknown eras (no Babbage fallback); pre‑Alonzo maps split correctly: Shelley (tags 0–10) and Allegra/Mary (0–10 plus 12=OutputTooBigUTxO).
  - Dijkstra UTXO map: shares Conway tags 0–7; tag 8 is absent; tags 9+ are shifted; tags 19–24 added; tag 23 intentionally unmapped.
  - Dijkstra UTXOW: added tags 19=MissingRequiredGuards and 20=MalformedGuardDatums with dedicated decoding; shared Conway tags 0–18 remain unchanged.
  - UtxoFailure.UnmarshalCBOR now propagates constructor‑id extraction failures and decode errors for known tags; only truly unknown tags return UnknownUtxoFailureError.
  - Fixed wrong CBOR slice capture in ApplyTxError.UnmarshalCBOR.

- **Migration**
  - Set Type explicitly to a valid era constructor before MarshalCBOR for ScriptsNotPaidUtxo and CollateralContainsNonADA (no Conway defaulting; Dijkstra tags supported).
  - If you handled GenericError or relied on Babbage fallback, handle the new Unknown* errors instead.

<sup>Written for commit e7ae05a72ea545419230aba862193984fe11f1ef. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/1923?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of unknown ledger, UTXOW, and UTXO failure types.
  * Preserved era information, constructor identifiers, and raw data when decoding unrecognized failures.
  * Prevented unknown eras from being misinterpreted as another era’s failure format.
  * CBOR marshaling now requires explicit constructor type information for affected failure payloads.

* **Tests**
  * Added coverage for unknown eras, constructor identifiers, and required type fields.
  * Expanded round-trip validation for supported eras.


cryptodj413:
Addressed the CodeRabbit review comment about `UtxoFailure.UnmarshalCBOR` masking real decode errors for recognized constructor tags — commit a131f74 checks `errorMap[failureType]` first and now propagates the actual decode error instead of collapsing it into `UnknownUtxoFailureError`. Added `TestUtxoFailure_KnownConstructorMalformedPayload` to cover it. The other nitpick suggestions from the review (additional era test coverage, extracting a shared helper) are intentionally out of scope for this PR.

ledger/error.go
```diff
@@ -225,14 +229,23 @@ func getEraSpecificUtxoFailureConstants(
...
 		}
-		return conwayMap, ConwayUtxoOutputTooBigUTxO, ConwayUtxoScriptsNotPaidUTxO, ConwayUtxoExUnitsTooBigUTxO, ConwayUtxoCollateralContainsNonADA
+		return conwayMap, ConwayUtxoOutputTooBigUTxO, ConwayUtxoScriptsNotPaidUTxO, ConwayUtxoExUnitsTooBigUTxO, ConwayUtxoCollateralContainsNonADA, nil
+	case EraIdShelley, EraIdAllegra, EraIdMary:
+		// Shelley, Allegra, and Mary predate Plutus scripts and
+		// collateral, so the Alonzo/Babbage/Conway-only failure kinds
+		// (OutputTooBigUtxo, ScriptsNotPaidUtxo, ExUnitsTooBigUtxo,
+		// CollateralContainsNonADA) cannot occur in these eras. baseMap
+		// alone (without any era-specific additions) is correct here.
+		return baseMap, 0, 0, 0, 0, nil
```
wolf31o2: Blocking: `baseMap` is not valid for all three early eras. It already contains Alonzo-only constructors such as tag 12 (`InsufficientCollateral`) and tags 17-20. I reproduced `[EraIdShelley, [12, 1, 2]]`; it decodes as `*InsufficientCollateral` instead of `*UnknownUtxoFailureError`. Please use truly era-specific maps so tags that did not exist in Shelley/Allegra/Mary are not treated as recognized later-era errors.

ledger/error.go
```diff
@@ -491,8 +568,15 @@ func (e *UtxowFailure) UnmarshalCBOR(data []byte) error {
 		// Conway uses flat enumeration (no wrapping)
 		return e.unmarshalConway(data, tmpFailure, failureType)
 	default:
-		// For unknown eras (Byron or future eras), try Babbage as fallback
-		return e.unmarshalBabbage(data, tmpFailure, failureType)
+		// Unknown era (Byron or a future era this decoder doesn't yet
+		// know about): we don't know this era's UTXOW constructor
+		// numbering, so don't guess by decoding as Babbage.
+		e.Err = &UnknownUtxowFailureError{
```
wolf31o2, line 744: Blocking compatibility issue: `EraIdDijkstra` (7) is registered and supported by this repository, but it reaches this default and every Dijkstra UTXOW failure becomes `UnknownUtxowFailureError`. Dijkstra validation largely inherits Conway. Please add explicit Dijkstra decoding behavior (using the appropriate constructor schema), or document and test an intentional unsupported-wire-schema decision.

ledger/error.go
```diff
@@ -707,13 +795,48 @@ func (e *UtxoFailure) UnmarshalCBOR(data []byte) error {
 		return err
 	}
 	e.Era = tmpData.Era
-	errorMap, _, _, _, _ := getEraSpecificUtxoFailureConstants(tmpData.Era)
+
+	// Best-effort extraction of the raw constructor tag, for error
+	// context, independent of whether we recognize the era/tag below.
+	failureType, idErr := cbor.DecodeIdFromList(tmpData.Err)
+	if idErr != nil {
+		failureType = -1
```
wolf31o2: Blocking: swallowing `idErr` turns malformed structure into a successful decode. I reproduced a Conway `UtxoFailure` whose inner value was a CBOR string rather than a constructor list; `UnmarshalCBOR` returned nil and stored `UnknownUtxoFailureError{FailureType: -1}`. Please propagate structural/tag-extraction errors and reserve `UnknownUtxoFailureError` for a valid numeric constructor tag that is absent from the era map.

wolf31o2 (changes requested):
Requesting changes for three blocking decoder-semantics issues. I reproduced the early-era tag collision and malformed-constructor behavior with focused tests against this exact head. The standard suite, race tests, lint, build, and vet pass, but those probes fail. DCO is also currently blocked because commits 8c1a9d1 and a131f74 lack Signed-off-by lines.

cryptodj413:
The 3 blocking bugs from the review are fixed:

- Shelley/Allegra/Mary now use an explicit constructor map that excludes the Alonzo+-only tags (`InsufficientCollateral`, `WrongNetworkInTxBody`, `OutsideForecast`, `TooManyCollateralInputs`, `NoCollateralInputs`) instead of reusing `baseMap` unmodified, which previously let those tags misdecode a genuinely unknown/malformed constructor pre-Alonzo.
- Dijkstra UTXOW/UTXO decoding is now wired to share Conway's constructor numbering instead of falling back to `UnknownUtxowFailureError` for every Dijkstra failure.
- `UtxoFailure.UnmarshalCBOR` now propagates the constructor-tag-extraction error (`idErr`) instead of swallowing it into a bogus `FailureType: -1` placeholder.

Each is covered by a regression test reproducing the exact repro case from the review. `make test`, `make lint`, and the ledger conformance suite (315/315) are all green.

Also rebuilt commit history so every commit now carries both a GPG signature and a DCO `Signed-off-by` trailer (previously missing on the first two commits), and force-pushed with `--force-with-lease`.

ledger/error.go
```diff
@@ -191,15 +195,22 @@ func getEraSpecificUtxoFailureConstants(
...
 	case EraIdBabbage:
 		baseMap[UtxoFailureOutputTooBigUtxoBabbage] = &OutputTooBigUtxo{}
 		baseMap[UtxoFailureScriptsNotPaidUtxoBabbage] = &ScriptsNotPaidUtxo{}
 		baseMap[UtxoFailureExUnitsTooBigUtxoBabbage] = &ExUnitsTooBigUtxo{}
 		baseMap[UtxoFailureCollateralContainsNonAdaBabbage] = &CollateralContainsNonADA{}
-		return baseMap, UtxoFailureOutputTooBigUtxoBabbage, UtxoFailureScriptsNotPaidUtxoBabbage, UtxoFailureExUnitsTooBigUtxoBabbage, UtxoFailureCollateralContainsNonAdaBabbage
-	case EraIdConway:
-		// Conway completely renumbered UTXO failure tags - use Conway-specific map
+		return baseMap, UtxoFailureOutputTooBigUtxoBabbage, UtxoFailureScriptsNotPaidUtxoBabbage, UtxoFailureExUnitsTooBigUtxoBabbage, UtxoFailureCollateralContainsNonAdaBabbage, nil
+	case EraIdConway, EraIdDijkstra:
```
wolf31o2: [P1] Dijkstra cannot reuse Conway’s UTXO constructor map. The current cardano-ledger Dijkstra encoding assigns tags 9/10/11/18 to OutputBootAddrAttrsTooBig, OutputTooBigUTxO, InsufficientCollateral, and NoCollateralInputs, respectively; this map decodes those same tags as OutputTooSmallUTxO, OutputBootAddrAttrsTooBig, OutputTooBigUTxO, and TooManyCollateralInputs. That turns valid Dijkstra failures into the wrong concrete errors (or payload decode failures). Please give Dijkstra its own UTXO map and leave genuinely unsupported Dijkstra-only constructors as UnknownUtxoFailureError. Source: https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/dijkstra/impl/src/Cardano/Ledger/Dijkstra/Rules/Utxo.hs

ledger/error.go
```diff
@@ -225,14 +236,48 @@ func getEraSpecificUtxoFailureConstants(
 			ConwayUtxoBabbageOutputTooSmallUTxO:   &BabbageOutputTooSmallUTxO{},
 			ConwayUtxoBabbageNonDisjointRefInputs: &BabbageNonDisjointRefInputs{},
 		}
-		return conwayMap, ConwayUtxoOutputTooBigUTxO, ConwayUtxoScriptsNotPaidUTxO, ConwayUtxoExUnitsTooBigUTxO, ConwayUtxoCollateralContainsNonADA
+		return conwayMap, ConwayUtxoOutputTooBigUTxO, ConwayUtxoScriptsNotPaidUTxO, ConwayUtxoExUnitsTooBigUTxO, ConwayUtxoCollateralContainsNonADA, nil
+	case EraIdShelley, EraIdAllegra, EraIdMary:
```
wolf31o2: [P1] Correction to my earlier review: I overgeneralized tag 12. It is invalid in Shelley, but Allegra and Mary use tag 12 for OutputTooBigUTxO. These three eras therefore cannot share this map. The authoritative encodings give Shelley tags 0–10, and Allegra/Mary tags 0–10 plus 12; tag 11 is not TriesToForgeADA in any of them. As written, Shelley [11] decodes as TriesToForgeADA while valid Allegra/Mary tag 12 is reported unknown. Please split the maps by actual era and add regressions for both cases. Sources: https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs and https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/allegra/impl/src/Cardano/Ledger/Allegra/Rules/Utxo.hs

wolf31o2 (changes requested):
Requesting changes: the malformed-payload and tag-extraction fixes address the earlier findings, but the replacement era maps still misclassify valid wire constructors. Focused local probes reproduce both blockers below. The full test, lint, formatting, build, and 315/315 conformance checks pass, so these are semantic decoding issues rather than mechanical failures.

ledger/error.go
```diff
@@ -225,14 +254,118 @@ func getEraSpecificUtxoFailureConstants(
...
+			ConwayUtxoInputSetEmptyUTxO:           &InputSetEmptyUtxo{},
+			ConwayUtxoFeeTooSmallUTxO:             &FeeTooSmallUtxo{},
+			ConwayUtxoValueNotConservedUTxO:       &ValueNotConservedUtxo{},
+			ConwayUtxoWrongNetwork:                &WrongNetwork{},
+			ConwayUtxoWrongNetworkWithdrawal:      &WrongNetworkWithdrawal{},
+			DijkstraUtxoOutputBootAddrAttrsTooBig: &OutputBootAddrAttrsTooBig{},
+			DijkstraUtxoOutputTooBigUTxO:          &OutputTooBigUtxo{},
+			DijkstraUtxoInsufficientCollateral:    &InsufficientCollateral{},
+			DijkstraUtxoScriptsNotPaidUTxO:        &ScriptsNotPaidUtxo{},
+			DijkstraUtxoExUnitsTooBigUTxO:         &ExUnitsTooBigUtxo{},
```
wolf31o2: Blocking: this map stops at tag 18, but Dijkstra defines valid UTxO failure constructors at tags 19–24 (including IncorrectTotalCollateralField, BabbageOutputTooSmallUTxO, BabbageNonDisjointRefInputs, PtrPresentInCollateralReturn, WrongNetworkInDirectDeposit, and WithdrawalsExceedAccountBalance). Because UtxoFailure falls back to UnknownUtxoFailureError when a tag is absent, valid Dijkstra failures lose their typed decoding. At minimum, add mappings/tests for the already-supported tags 19–21; add the remaining Dijkstra error types or explicitly preserve them with dedicated typed raw errors. See the upstream wire map: https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/dijkstra/impl/src/Cardano/Ledger/Dijkstra/Rules/Utxo.hs#L2440-L2488

wolf31o2 (changes requested):
Requesting changes: the Dijkstra UTxO constructor map is incomplete. Local validation passes (make test, go test ./..., formatting, and diff checks), but valid Dijkstra wire constructors are still decoded as UnknownUtxoFailureError.

ledger/error.go
```diff
@@ -487,12 +715,27 @@ func (e *UtxowFailure) UnmarshalCBOR(data []byte) error {
 	case EraIdBabbage:
 		// Babbage wraps Alonzo failures in tag 1, adds Babbage-specific tags
 		return e.unmarshalBabbage(data, tmpFailure, failureType)
-	case EraIdConway:
-		// Conway uses flat enumeration (no wrapping)
+	case EraIdConway, EraIdDijkstra:
```
wolf31o2: Blocking: this groups EraIdDijkstra with Conway and calls unmarshalConway, whose switch only covers tags 0-18. Upstream Dijkstra adds tag 19 (MissingRequiredGuards) and tag 20 (MalformedGuardDatums), encoded here: https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/dijkstra/impl/src/Cardano/Ledger/Dijkstra/Rules/Utxow.hs#L364-L416. Please add Dijkstra-specific decoding and tests; otherwise valid Dijkstra wire failures lose their concrete type and are incorrectly surfaced as UnknownUtxowFailureError.

wolf31o2 (changes requested):
Request changes: Dijkstra UTXOW decoding is still incomplete. The PR's tests pass, but valid Dijkstra constructors 19 and 20 are currently reported as UnknownUtxowFailureError. Please add the Dijkstra-specific cases/types and regression tests before merge. Upstream reference: https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/dijkstra/impl/src/Cardano/Ledger/Dijkstra/Rules/Utxow.hs#L364-L416

wolf31o2 (approved):

PR #1924, opened 2026-07-30
fix(ledger): reject unmatched redeemer purposes instead of dropping them

### LINKED ISSUE #1869: ledger: reject unmatched redeemer purposes instead of dropping them
## Problem

Script purpose builders return `nil` for unmatched redeemer tags/indices, and context construction silently skips those redeemers:

- `ledger/common/script/purpose.go`: `scriptPurposeBuilder`
- `ledger/common/script/purpose.go`: `BuildScriptPurpose`
- `ledger/common/script/context.go`: `redeemersInfo`

Conway and Dijkstra have explicit `UtxoValidateExtraneousRedeemers` rules, but Alonzo and Babbage do not include an equivalent rule in their validation lists. Direct context callers can also receive contexts with redeemers omitted instead of an error.

## Expected behavior

Unmatched redeemer purpose construction should be observable as an error, not a silent omission.

## Suggested fix

Options:

- Change purpose builders to return `(ScriptPurpose, error)`.
- Add shared redeemer tag/index validation and wire it into Alonzo/Babbage as well as Conway/Dijkstra.
- Make `redeemersInfo` fail closed when a redeemer cannot be mapped to a purpose.

## Tests

Add tests for:

- unknown redeemer tag
- out-of-range redeemer index for each purpose type
- Alonzo/Babbage validation rejects extraneous redeemers
- context construction does not silently drop invalid redeemers

cryptodj413:
## Summary

- `toScriptPurposeFunc` / `scriptPurposeBuilder` / `BuildScriptPurpose` (in `ledger/common/script/purpose.go`) now return `(ScriptPurpose, error)` instead of a nilable `ScriptPurpose`. Every former silent `return nil` (out-of-range index, unresolved input, the `RedeemerTagGuarding` tag, or an unrecognized tag) now returns a new `script.UnmatchedRedeemerError{RedeemerKey}`. All call sites were updated for the new signature: `redeemersInfo` (`ledger/common/script/context.go`) now returns `(KeyValuePairs[ScriptPurpose, Redeemer], error)` and fails closed on the first unmatched redeemer instead of silently dropping it, and `NewTxInfoV1/V2/V3FromTransaction` propagate that error, as does the `BuildScriptPurpose` call site in `conway.UtxoValidatePlutusScripts`.
- Extracted Conway's existing extraneous-redeemer check into a new shared `common.ValidateExtraneousRedeemers(tx Transaction) error` (`ledger/common/rules.go`), backed by a new `common.ExtraneousRedeemerError` type (`ledger/common/errors.go`). It checks every redeemer's tag/index against the actual input/mint/cert/withdrawal/voter/proposal counts; `RedeemerTagGuarding` and unrecognized tags are always treated as extraneous. `conway.UtxoValidateExtraneousRedeemers` now delegates to this shared function, translating back to the pre-existing `ExtraRedeemerError` type via `errors.As` so Conway's external behavior is unchanged.
- **Behavior change for Alonzo and Babbage:** both eras now wire a new `UtxoValidateExtraneousRedeemers` rule (thin wrappers over the shared function) into their `UtxoValidationRules` lists, immediately before `UtxoValidatePlutusScripts`, matching Conway's/Dijkstra's existing rule ordering. Previously neither era had an equivalent rule, so a transaction containing a redeemer that didn't match any real script purpose could pass phase-1 validation undetected. It will now be rejected. This is new phase-1 rejection behavior, not just an error-message-quality fix, and is the highest-risk part of this change.
- Dijkstra was intentionally left untouched — it already has its own independent guarding-aware `UtxoValidateExtraneousRedeemers` and doesn't call any of the functions changed here.

Fixes #1869

## Test plan

- `go test -race ./...` (whole repo, including example modules via `make test`) — all pass.
- `golangci-lint run ./...` (main module + example modules via `make lint`) — 0 issues.
- `gofmt -l` — clean on all changed files.
- Ledger rule conformance suite (`go test -v ./internal/test/conformance/...`) re-run specifically because of the Alonzo/Babbage behavioral risk called out above: 315/315 vectors passing, no regressions versus baseline.
- New/updated unit tests: `ledger/common/script/purpose_test.go`, `ledger/common/script/context_test.go`, `ledger/common/rules_test.go` (new), `ledger/alonzo/rules_test.go`, `ledger/babbage/rules_test.go` — covering unknown/unmatched redeemer tags, out-of-range indices per purpose type, Alonzo/Babbage rejecting extraneous redeemers, and context construction failing closed instead of silently dropping.

---
## Summary by cubic
Reject unmatched redeemer purposes instead of silently dropping them, add a shared extraneous-redeemer check, and surface unknown tags; Alonzo and Babbage now reject non-matching redeemers during phase-1.

- **Bug Fixes**
  - Context builders and `conway.UtxoValidatePlutusScripts` now return `script.UnmatchedRedeemerError` for unmatched redeemers instead of dropping them.
  - Added `common.ValidateExtraneousRedeemers` and `common.ExtraneousRedeemerError`; `conway.UtxoValidateExtraneousRedeemers` delegates to it while preserving its existing error type.
  - Alonzo/Babbage run `UtxoValidateExtraneousRedeemers` before Plutus, matching Conway/Dijkstra; unknown/guarding tags or out-of-range indices now fail phase-1.
  - `sortedRedeemerKeys` now iterates all tags and sorts by (tag, index) so unknown tags are checked and rejected; Dijkstra’s guarding path wraps tx with `transactionWithoutGuardingRedeemers` when building TxInfo to exclude genuine guarding redeemers during guard-script evaluation.

- **Migration**
  - In Alonzo/Babbage, ensure each redeemer maps to a valid purpose/index: Spend→input, Mint→distinct policy, Cert→certificate, Reward→withdrawal. Voting/Proposing are unsupported; Guarding is invalid. Transactions with stray, unknown, or misindexed redeemers will now be rejected.

<sup>Written for commit 6687261289d3240549018bf7e79215cfd4ff1048. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/1924?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Transactions with extraneous, unknown, unsupported, or out-of-range redeemers are now rejected consistently.
  * Invalid redeemer purposes now produce clear, typed validation errors instead of being silently ignored.
  * Improved handling of guarding redeemers during script validation and transaction information construction.
  * Validation now covers spending, minting, certificates, withdrawals, voting, and proposing redeemers across supported ledger eras.

* **Tests**
  * Added comprehensive coverage for valid, invalid, missing, and guarding redeemer scenarios.


ledger/common/script/context.go
```diff
@@ -689,21 +698,21 @@ func buildWitnessDatums(
...
+		return KeyValuePairs[ScriptPurpose, Redeemer]{}, nil
 	}
 	redeemerKeys := sortedRedeemerKeys(redeemers)
 	ret := make(KeyValuePairs[ScriptPurpose, Redeemer], 0, len(redeemerKeys))
 	for _, key := range redeemerKeys {
 		redeemerValue := redeemers.Value(uint(key.Index), key.Tag)
-		purpose := toScriptPurpose(key)
-		if purpose == nil {
-			continue
+		purpose, err := toScriptPurpose(key)
```
wolf31o2, line 701: Blocking: `redeemerKeys` comes from `sortedRedeemerKeys`, which only requests the known Spend-through-Proposing tags. An unknown tag is absent from this loop, so `toScriptPurpose` never gets the chance to return `UnmatchedRedeemerError`; `NewTxInfoV1/V2/V3FromTransaction` still succeeds with that redeemer silently omitted. I reproduced this with `RedeemerTag(99)`: V3 returned nil. Please iterate and sort the actual keys (filtering Dijkstra guarding explicitly in its adapter), or otherwise reject unsupported keys before building the context.

ledger/common/script/context_test.go
```diff
@@ -614,3 +615,54 @@ func TestScriptContextV3(t *testing.T) {
...
+// dropping a redeemer that doesn't map to any valid script purpose, for
+// each of the TxInfo builders (V1/V2/V3).
+func TestNewTxInfoFromTransactionUnmatchedRedeemer(t *testing.T) {
+	// A spending redeemer with index 0, but the transaction has no inputs,
+	// so the redeemer can't be matched to any input.
+	newTx := func() *conway.ConwayTransaction {
+		tx := &conway.ConwayTransaction{}
+		tx.WitnessSet.WsRedeemers = conway.ConwayRedeemers{
+			Redeemers: map[common.RedeemerKey]common.RedeemerValue{
+				{Tag: common.RedeemerTagSpend, Index: 0}: {},
```
wolf31o2, line 630: This exercises an out-of-range recognized tag, but it does not cover the unknown-tag case explicitly required by #1869. Please add an unknown `RedeemerTag(99)` case for each TxInfo builder; that case currently fails because `sortedRedeemerKeys` omits the key before `redeemersInfo` sees it.

wolf31o2 (changes requested):
Requesting changes: direct TxInfo construction still silently drops unknown redeemer tags, so #1869's fail-closed requirement remains incomplete. The standard local suite passed, but a focused unknown-tag probe returned nil instead of UnmatchedRedeemerError. The failing DCO sign-off must also be resolved before merge.

cryptodj413:
Thanks for the review — both issues are addressed:

**1. Unknown-tag redeemers silently dropped in `TxInfo` construction**

`sortedRedeemerKeys` (`ledger/common/script/context.go`) now walks the
actual entries in `redeemers.Iter()` instead of enumerating a fixed list
of 6 known `RedeemerTag` values via `Indexes()`, sorting the resulting
keys deterministically by `(Tag, Index)`. This means every redeemer that's
actually present is now considered, so an unrecognized tag correctly
fails closed with `UnmatchedRedeemerError` instead of being dropped.
Added `TestNewTxInfoFromTransactionUnknownRedeemerTag` reproducing your
exact `RedeemerTag(99)` repro for V1/V2/V3.

Making all redeemers visible this way also surfaced a follow-on issue:
genuine Dijkstra guarding redeemers are now visible to the shared purpose
builder too, which intentionally rejects `RedeemerTagGuarding` as
unmatched (it's not a real script purpose in the shared `TxInfo` model).
`validateGuardingPlutusScripts` (`ledger/dijkstra/rules.go`) now wraps
`tx` with the existing `transactionWithoutGuardingRedeemers` at all four
`NewTxInfoV1/V2/V3FromTransaction` call sites before building context for
guarding-script evaluation, mirroring how `UtxoValidatePlutusScripts`
already excludes guarding redeemers from the main Plutus-validation path.
Added `TestNewTxInfoFromTransactionGuardingRedeemer`, which proves the
wrapper is both necessary (unwrapped tx with a real guarding redeemer now
fails closed) and sufficient (wrapped succeeds).

Verified: `go build ./...`, `go vet ./...`, `go test -race ./...` (full
repo, including the 4 pre-existing Dijkstra guarding regression tests,
still passing), `golangci-lint run ./...` (0 issues), `gofmt -l` clean,
and conformance re-run at 315/315 (no regressions).

**2. DCO sign-off**

Both commits on this branch are now GPG-signed *and* carry a
`Signed-off-by:` trailer:

- `d704a4a` — `fix(ledger): reject unmatched redeemer purposes instead of dropping them` (rewritten to add sign-off; content unchanged)
- `6687261` — `fix(ledger): include all redeemer tags when building script context` (new commit for the fixes above)

Force-pushed with `--force-with-lease` to rewrite the history and add the
missing sign-off.

wolf31o2 (approved):

PR #1925, opened 2026-07-30
fix(ledger): classify block types by declared protocol ranges

### LINKED ISSUE #1919: fix(ledger): DetermineBlockType rejects valid Alonzo 6 and Babbage 8 headers
## Problem

`ledger.DetermineBlockType` does not recognize the complete protocol-major
ranges declared by the era packages:

- `alonzo.MinProtocolVersionAlonzo = 5`
- `alonzo.MaxProtocolVersionAlonzo = 6`
- `babbage.MinProtocolVersionBabbage = 7`
- `babbage.MaxProtocolVersionBabbage = 8`

The helper currently matches Alonzo only when the major is exactly 5 and
Babbage only when it is exactly 7. Consequently, valid protocol-major 6
Alonzo headers fail with `unknown proto major 6 for Shelley-like`, and valid
protocol-major 8 Babbage headers fail with
`unknown proto major 8 for 10-field header`.

## Impact

Callers that use `DetermineBlockType` to independently classify a decoded
header reject valid historical blocks even though the era-specific block
constructor can decode and body-validate them.

This was discovered while reviewing
blinklabs-io/dingo#2990. That PR correctly avoids trusting an archive's
era claim and delegates classification to this upstream helper; the range
omission is in gouroboros rather than in the PR's approach.

## Reproduction

Against gouroboros v0.190.0:

1. Construct a `BabbageBlock` with `ProtoVersion.Major = 8` and a valid body
   hash.
2. Encode it and call
   `ledger.NewBlockFromCbor(ledger.BlockTypeBabbage, raw)`.
3. Decoding and body validation succeed.
4. Call `ledger.DetermineBlockType(decoded.Header().Cbor())`.
5. Classification fails with:

   ```
   unknown proto major 8 for 10-field header
   ```

The equivalent range omission is visible for Alonzo major 6.

## Suggested fix

- Match Alonzo and Babbage using their exported min/max protocol-version
  constants, as is already done for Conway and Dijkstra.
- Add table tests covering every supported major, especially Alonzo 6 and
  Babbage 8.
- Keep a genuinely unsupported major in the negative test coverage.

chrisguiney:
Fixes #1919.

## Problem

`DetermineBlockType` matched Alonzo on major 5 and Babbage on major 7 alone, while the era packages declare ranges:

| Era | Declared range | Previously matched |
|---|---|---|
| Alonzo | 5–6 | 5 only |
| Babbage | 7–8 | 7 only |

Valid major 6 and major 8 headers were rejected even though the era constructors decode and body-validate them, so callers that classify a header independently refused real blocks. Conway and Dijkstra already matched by range; the inconsistency is what let the two drift.

**This is not hypothetical.** A `TestVerifyBlockBody` case in this repo held a real mainnet Babbage block with major 8 and asserted it should *fail* — with `expectedErr: "unknown proto major 8 for 10-field header"`. The block is valid; the test had encoded the bug as expected behaviour.

## Change

Match every era against the min and max its own package declares. The era packages are the authority on which majors belong to them, and duplicating a single value in `verify_block.go` is precisely what allowed the drift.

**One case the report did not mention:** the ten-field header path had the same omission for Alonzo, so major 6 was rejected in *both* header shapes, not just Shelley-like. Covered.

The exported `ProtoMajor*` constants are kept for callers that reference them — removing exported API in a bug fix seemed wrong — and are now documented as naming an era's first major rather than its extent.

## Tests

`TestDetermineBlockTypeCoversDeclaredProtocolRanges` walks every major each era declares, in both header shapes, driving the cases from the era constants rather than literals so a future range change is covered without touching the test. Before the fix it failed on exactly three cases:

```
Alonzo major 6:            unknown proto major 6 for Shelley-like
Alonzo ten-field major 6:  unknown proto major 6 for 10-field header
Babbage major 8:           unknown proto major 8 for 10-field header
```

`TestDetermineBlockTypeRejectsUndeclaredProtocolMajor` keeps the negative side honest — majors 0, 1, and 99 are still rejected in both shapes, so widening the ranges did not turn the helper into one that accepts anything.

The `TestVerifyBlockBody` case described above becomes a success case, which is stronger evidence than the synthetic headers: a real Babbage major-8 block now verifies end to end. Its negative role is replaced by the classification tests above, so nothing loses coverage.

## Verification

- `make test` — 3601 passing, 0 failures
- `make lint` — 0 issues across all 9 targets
- `make format` — no changes to the edited files

## Impact

Discovered while reviewing blinklabs-io/dingo#2990, which classifies archive-supplied headers independently and fails closed when classification fails. With `v0.190.0` that combination rejects valid Alonzo-6 and Babbage-8 blocks served from a Bark archive. A release carrying this fix clears that.

---
## Summary by cubic
Classifies block headers by each era’s declared protocol-major ranges instead of single majors. Fixes rejection of valid Alonzo-6 and Babbage-8 blocks and aligns `DetermineBlockType` with the era packages.

- **Bug Fixes**
  - Match by min/max from `shelley`, `allegra`, `mary`, `alonzo`, `babbage`, `conway`, `dijkstra` in both 15-field and 10-field header paths.
  - Accept Alonzo major 6 and Babbage major 8; also correct the ten-field path for Alonzo.
  - Keep `ProtoMajor*` constants (now documented as first-major only) for compatibility.
  - Add tests to cover all declared ranges and reject undeclared majors; update the Babbage-8 verification to succeed.

- **Refactors**
  - Use `github.com/stretchr/testify/require` in the new block type tests for consistency; no functional change.

<sup>Written for commit 6c3b492d43c97156fdf804cdb8fdd7a8bb375a6b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/1925?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


chrisguiney:
> **Use Testify for fatal test assertions.**

Done in 6c3b492 — the new tests now use `require.NoError`, `require.Equalf`, and `require.Errorf`.

Worth recording why they weren't written that way initially: `ledger/verify_block_test.go` uses `t.Fatalf` for all 32 of its existing assertions and doesn't import testify, so matching it kept the file internally consistent. Checking more widely, though, testify is the more common choice across the repository — 125 test files to 103, including four elsewhere in `ledger/` — so your read of the convention holds even though this particular file is the exception.

I converted only the added tests. Rewriting the file's existing assertions would bury a small classification fix under unrelated diff; better raised separately if the file should be normalised.

Verified the conversion didn't weaken detection: reverting Babbage to an exact match on major 7 still fails `Babbage major 8` with the intended message. `make test` 3601 passing, `make lint` 0 issues across all 9 targets.

wolf31o2 (approved):

PR #1944, opened 2026-08-03
fix(ledger): prevent invalid Bech32 addresses from falling back to Base58

### LINKED ISSUE #1930: ledger/common: NewAddress falls back to base58 on a bech32 checksum failure, silently accepting corrupted mainnet addresses
### Summary

`common.NewAddress` treats **any** bech32 decode failure — including an invalid
checksum — as a signal to try base58 instead. Shelley-era addresses are composed
entirely of base58-legal characters, so a corrupted `addr1…` string does not fail;
it silently decodes to unrelated bytes.

`ledger/common/address.go` (v0.190.0):

```go
func NewAddress(addr string) (Address, error) {
	var decoded []byte
	_, data, err := bech32.DecodeNoLimit(addr)
	if err == nil {
		decoded, err = bech32.ConvertBits(data, 5, 8, false)
		...
	} else {
		// bech32 failed — try base58 (Byron addresses)
		decoded = base58.Decode(addr)
		if len(decoded) == 0 {
			return Address{}, err
		}
	}
	...
}
```

The bech32 error — which may be `ErrInvalidChecksum` or `ErrMixedCase` — is
discarded unless base58 also yields nothing.

### Reproduction

Every single-character substitution within the data part of a valid mainnet
address, HRP left intact:

```go
good, _ := common.NewAddressFromParts(common.AddressTypeKeyKey, 1, pay, stake)
txt := good.String()
// substitute each position with each other bech32 charset character
// and count how many common.NewAddress accepts
```

Output against v0.190.0:

```
valid mainnet address: addr1qx4242424242...ascmu829
single-char substitutions tried: 2883, ACCEPTED: 2697

first accepted corruption:
  addr1qx424q42424242424242424242424242424242424242424mhwamhwa...ascmu829
  -> accepted, type=0 network=6 payment=00e6bdb85086a3fa
```

**2697 of 2883** corruptions are accepted. Note the decoded network id is `6`,
which does not exist on Cardano — see the companion issue about
`populateFromBytes` not validating header nibbles.

Testnet addresses are largely immune only incidentally, because `_` is not in the
base58 alphabet. The failure mode is therefore concentrated on mainnet, where real
funds are.

### Impact

A caller that validates a user-supplied address by calling `NewAddress` and
checking the error gets no signal at all for a mistyped mainnet address. Any
downstream code — a wallet, a transaction builder, an explorer — will treat the
corrupted address as valid and act on unrelated key-hash bytes.

In practice most of these decode to non-canonical payload lengths, so
cardano-ledger rejects the resulting transaction rather than misdirecting funds.
That is a thin margin to rely on, and it is not guaranteed: a corruption landing
on a canonical layout would be indistinguishable from a legitimate address.

### Suggested fix

Only attempt base58 when the input is not bech32-shaped at all. A checksum
failure on a string carrying a known Shelley HRP (`addr`, `addr_test`, `stake`,
`stake_test`, …) should be returned as an error, not retried as Byron. It would
also help to verify that the decoded HRP agrees with the header byte's network and
address type; at present the HRP is parsed and then ignored.

Found while auditing [Salvionied/apollo](https://github.com/Salvionied/apollo)
ahead of its 2.0 release. Version: `v0.190.0`.

arepala-uml:
1. Prevent malformed Shelley addresses from being reinterpreted as Byron addresses and validate Bech32 HRPs against decoded address headers.
2. Recognized Shelley payment and stake address prefixes.
3. Returned Bech32 checksum and mixed-case errors directly.
4. Verified that the Bech32 prefix matches the decoded address type and network.
5. Preserved Base58 decoding for valid Byron addresses.
6. Added tests for single-character address corruption.
7. Added tests for incorrect network and address-type prefixes.
8. Optimized prefix detection to avoid unnecessary allocations.

Closes #1930 

---
## Summary by cubic
Prevent invalid Shelley `bech32` addresses from falling back to Byron `base58`, and reject `bech32`-encoded Byron addresses. Validate the HRP against the decoded address header and return clearer errors.

- **Bug Fixes**
  - If a Shelley HRP (`addr1`, `addr_test1`, `stake1`, `stake_test1`) is present, do not attempt `base58` fallback on `bech32` errors; return the checksum/mixed-case error.
  - Ensure the `bech32` HRP matches the decoded address type and network; provide a mismatch error with both HRPs.
  - Enforce Byron-only `base58`: reject `bech32`-encoded Byron addresses while preserving valid Byron `base58` decoding.
  - Add tests for single-character corruption and wrong HRPs; optimize case-insensitive prefix checks.

<sup>Written for commit fd2559d7f0c7c425df541d1238cd67d68f52e5a2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/1944?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for Shelley addresses.
  * Malformed addresses are now rejected instead of being misinterpreted.
  * Address network and type information is verified against the encoded format.
  * Address prefix matching is now case-insensitive.

### DIFF
--- a/ledger/common/address.go
+++ b/ledger/common/address.go
@@ -21,6 +21,7 @@ import (
 	"hash/crc32"
 	"io"
 	"math/big"
+	"strings"
 
 	"github.com/blinklabs-io/gouroboros/cbor"
 	"github.com/blinklabs-io/plutigo/data"
@@ -86,18 +87,22 @@ type Address struct {
 	byronAddressAttr ByronAddressAttributes
 }
 
-// NewAddress returns an Address based on the provided bech32/base58 address string
-// It detects if the string has mixed case assumes it is a base58 encoded address
-// otherwise, it assumes it is bech32 encoded
+// NewAddress returns an Address based on the provided bech32/base58 address
+// string.
 func NewAddress(addr string) (Address, error) {
 	var decoded []byte
-	_, data, err := bech32.DecodeNoLimit(addr)
+	hrp, data, err := bech32.DecodeNoLimit(addr)
 	if err == nil {
 		decoded, err = bech32.ConvertBits(data, 5, 8, false)
 		if err != nil {
 			return Address{}, err
 		}
 	} else {
+		// A string with a known Shelley HRP was intended to be bech32. Do not
+		// reinterpret a checksum or mixed-case failure as a Byron address.
+		if hasShelleyAddressHRP(addr) {
+			return Address{}, fmt.Errorf("invalid bech32 address: %w", err)
+		}
 		// bech32 failed — try base58 (Byron addresses)
 		decoded = base58.Decode(addr)
 		if len(decoded) == 0 {
@@ -109,7 +114,29 @@ func NewAddress(addr string) (Address, error) {
 	if err != nil {
 		return Address{}, err
 	}
+	if hrp != "" && !strings.EqualFold(hrp, a.generateHRP()) {
+		return Address{}, fmt.Errorf(
+			"address HRP %q does not match expected HRP %q",
+			hrp,
+			a.generateHRP(),
+		)
+	}
 	return a, nil
+}
+
+func hasShelleyAddressHRP(addr string) bool {
+	lowerAddr := strings.ToLower(addr)
+	for _, prefix := range []string{
+		"addr1",
+		"addr_test1",
+		"stake1",
+		"stake_test1",
+	} {
+		if strings.HasPrefix(lowerAddr, prefix) {
+			return true
+		}
+	}
+	return false
 }
 
 // NewAddressFromBytes returns an Address based on the raw bytes provided
--- a/ledger/common/address.go
+++ b/ledger/common/address.go
@@ -102,7 +102,7 @@ func NewAddress(addr string) (Address, error) {
 		// A string with a known Shelley HRP was intended to be bech32. Do not
 		// reinterpret a checksum or mixed-case failure as a Byron address.
 		if hasShelleyAddressHRP(addr) {
-			return Address{}, fmt.Errorf("invalid bech32 address: %w", err)
+			return Address{}, err
 		}
 		// bech32 failed — try base58 (Byron addresses)
 		decoded = base58.Decode(addr)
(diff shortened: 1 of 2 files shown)


arepala-uml:
@cubic-dev-ai

ledger/common/address.go
```diff
@@ -109,9 +115,31 @@ func NewAddress(addr string) (Address, error) {
 	if err != nil {
 		return Address{}, err
 	}
+	if isBech32 {
+		expectedHRP := a.generateHRP()
```
wolf31o2, line 124: This check derives the expected HRP with generateHRP(), but generateHRP() treats AddressTypeByron as a generic addr/addr_test address. As a result, a valid Byron payload can be converted to Bech32 with the matching inferred HRP and NewAddress accepts it, even though Byron addresses are Base58-only. I reproduced this by converting the valid Byron address Ae2tdPwUPEYwFx4dmJheyNPPYXtvHbJLeCaA96o6Y2iiUL18cAt7AizN2zG to raw bytes, encoding them as addr_test1stvpskppsdvpcpyxtepdyde6mklt6hf2e7quwcxgfztszs5gnalwwccfrwsqqxhsrytd2vqclr3, and observing NewAddress return nil. Reject Bech32 when the decoded type is Byron (or otherwise handle Byron explicitly), and add a regression test.

wolf31o2 (changes requested):
Request changes: the new HRP validation does not reject Bech32-encoded Byron payloads. Please address the inline comment and add a regression test before merge.

ledger/common/address.go
arepala-uml: Made changes where `NewAddress` now rejects Byron payloads decoded from `Bech32` while continuing to accept valid Base58-encoded Byron addresses. I also added a regression test using the reported address.

wolf31o2 (approved):

PR #1953, opened 2026-08-06
Fix/byron stateful ssc body proof validation

### LINKED ISSUE #1940: ledger/byron: add stateful SSC body-proof validation
Byron body-proof validation checks transaction, delegation, and update proofs, but SSC proof hashes are only checked structurally. The actual SSC hashes depend on epoch-accumulated commitments/openings/shares/certificates and cannot be validated from a single block alone.

ledger/byron.ByronMainBlock.ValidateBodyProof also deliberately skips ssc_proof, so an alteration confined to SSC payload data is not detected by the exposed body-proof validator.

Impact: the Byron body validator is not a complete body-proof/consensus validator when used without additional epoch state.

References:
- https://github.com/blinklabs-io/gouroboros/blob/main/ledger/byron/bodyproof.go
- https://github.com/blinklabs-io/gouroboros/blob/main/consensus/byron/validate.go

Acceptance criteria:
- Define a stateful Byron body-proof validation API that accepts the required epoch SSC state and verifies the accumulated proof hashes.
- Ensure block/header validation uses that API when state is available.
- Clearly distinguish the stateful full validator from the existing block-local structural validator.
- Add positive and tampered-SSC fixtures.

cryptodj413:
Closes #1940 

---
## Summary by cubic
Validates Byron `ssc_proof` using only the block’s SSC payload, with opt-in hash comparison and always-on structural checks. Removes the epoch-state dependency, tightens CBOR shape checks, and fixes certificate hashing so real mainnet blocks validate correctly when enabled.

- **New Features**
  - `ledger/common`: added `VerifyConfig.EnableByronSscProofHashValidation` to toggle full `ssc_proof` hash validation.
  - `ledger/byron` and `consensus/byron`: validators accept `cfg` and forward the flag; stateful SSC validators removed.
  - Added focused SSC hash-validation tests and golden/real vectors; updated Byron bench limits.

- **Bug Fixes**
  - Restored structural fallback in `consensus/byron.ValidateBodyHash` when the flag is off.
  - Enforced container shapes before hashing (tag-258 set for commitments/identity sets, CBOR map for openings/shares); stricter CertificatesProof length; proof-type vs payload cross-check.
  - Fixed certificate pubkey field index; require a non-empty CBOR byte string; preserved-wire-byte hashing for commitments/openings/shares; certificates hash a rebuilt CBOR map.

<sup>Written for commit 3dd84d415a8e5ef42ae2a2e056e69c24bdfaebb4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/1953?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added epoch-level validation for Byron SSC proofs.
  * Added accumulation and verification of commitments, openings, shares, and certificates across blocks.
  * Added validation errors for missing state, malformed proofs, and tampered historical data.

* **Bug Fixes**
  * SSC proofs are now checked against accumulated epoch state, improving detection of invalid block data.

* **Tests**
  * Added coverage for valid proofs, malformed data, missing state, tampering, and supported payload formats.


wolf31o2 (changes requested):
Request changes: the new stateful Byron SSC validator is not ready to merge as a production-facing API. The implementation is tested against an empty SSC map, but non-empty canonical encoding and accumulation semantics are explicitly unconfirmed. Until those are proven against cardano-ledger/non-empty vectors, callers can reject valid Byron blocks. See the two blocking inline comments for concrete fix options. The duplicate proof work is also worth simplifying, but it is secondary to the correctness gate.

ledger/byron/sscstate.go
```diff
@@ -0,0 +1,494 @@
...
+// preserved CBOR bytes, with keys ordered per RFC 8949's core deterministic
+// encoding (ascending, since all keys here are fixed-length byte strings).
+//
+// This mirrors cardano-sl's own commitment/opening/share/certificate
+// hashes, which are computed over the CBOR encoding of a genuine map
+// (HashMap StakeholderId ...), not a bespoke concatenation of raw bytes:
+// for the bundled mainnet fixture's empty VssCertificatesMap, this produces
+// blake2b256(0xa0) -- the CBOR encoding of an empty map -- which matches
+// the real header's ssc_proof hash for that block exactly.
+func canonicalMapHash(m map[common.Blake2b224][]byte) common.Blake2b256 {
```
wolf31o2, line 798: Blocking correctness issue: canonicalMapHash is consensus-critical, but this PR only proves it for the empty map. The implementation explicitly says non-empty parity with cardano-ledger is unconfirmed. Please add non-empty golden vectors generated from cardano-ledger (commitments, openings, shares, and certificates, with multiple stakeholders and preserved entry encodings) and assert exact hashes before relying on this validator for those blocks. This is a correctness requirement, not an API-stability concern.

consensus/byron/validate.go
```diff
@@ -987,6 +994,62 @@ func ValidateBodyHash(block *byron.ByronMainBlock) error {
...
+// Wiring point for future work: this package tracks per-epoch consensus
+// parameters (ByronConfig.SlotsPerEpoch, SlotToEpoch,
+// IsEpochBoundarySlot) but does not itself carry an accumulated
+// byron.ByronEpochSscState across the sequence of blocks in an epoch --
+// there is no block-validation loop in this package to thread it
+// through. Callers that do drive such a loop (e.g. a chain follower)
+// should construct one byron.ByronEpochSscState per epoch, call
+// AccumulateBlock for every main block of that epoch in order, and pass
+// it to this function instead of calling ValidateBodyHash alone.
+func ValidateBodyHashWithSscState(
```
wolf31o2: The concern here is correctness, not 0.x API stability. Keeping this exported function is fine if that is the project choice; however, its current documentation recommends it for chain-following validation while the non-empty SSC encoding is unverified. Please either prove the implementation with non-empty cardano-ledger vectors, or clearly document that callers must not use this path for production validation until that parity work is complete. Example test gate: a non-empty real/generated SSC fixture must pass ValidateBodyHashWithSscState and tampering each accumulated entry must fail.

wolf31o2:
Correction to my earlier review: the 0.x status means API stability/experimental naming is not a merge blocker, and I withdraw that suggestion. The remaining request is correctness-only: please establish non-empty SSC canonical encoding and accumulation parity with cardano-ledger using authoritative vectors, or document the production-validation limitation until that work is complete. The current CHANGES_REQUESTED review remains for that reason.

ledger/byron/sscstate.go
```diff
@@ -0,0 +1,656 @@
...
+			return fmt.Errorf(
+				"%w: ssc payload type %d requires 2 elements after the "+
+					"type, got %d",
+				ErrBodyProofMismatch, sscType, len(rest),
+			)
+		}
+		if err := checkHash(
+			"ssc primary hash",
+			proofSlice[1],
+			common.Blake2b256Hash(rest[0]),
```
wolf31o2, line 470: Request changes: the primary SSC field is hashed here without validating its wire shape. The Haskell Byron decoder is explicit: [dropSscPayload](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/byron/ledger/impl/src/Cardano/Chain/Ssc.hs#L77-L95) dispatches to [dropCommitmentsMap/dropOpeningsMap/dropSharesMap](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/byron/ledger/impl/src/Cardano/Chain/Ssc.hs#L191-L235), requiring a tag-258 set for commitments and a CBOR map for openings/shares. rest[0] is only a preserved RawMessage here, so an untagged array (or another wrong-shaped value) can be paired with a freshly recomputed header hash and pass this check. That accepts a payload the Byron node would reject; a body-proof hash authenticates bytes, but it does not make invalid SSC CBOR valid.

Please decode/validate the primary field with the corresponding shape before hashing (the existing decodePrimaryEntries path is a likely place to reuse), and add independent coverage using the Byron golden vectors: [CommitmentsMap](https://github.com/IntersectMBO/cardano-ledger/tree/master/eras/byron/ledger/impl/golden/cbor/ssc/CommitmentsMap), [OpeningsMap](https://github.com/IntersectMBO/cardano-ledger/tree/master/eras/byron/ledger/impl/golden/cbor/ssc/OpeningsMap), [SharesMap](https://github.com/IntersectMBO/cardano-ledger/tree/master/eras/byron/ledger/impl/golden/cbor/ssc/SharesMap), and [VssCertificatesMap](https://github.com/IntersectMBO/cardano-ledger/tree/master/eras/byron/ledger/impl/golden/cbor/ssc/VssCertificatesMap).

ledger/byron/sscstate_test.go
```diff
@@ -0,0 +1,721 @@
...
+	assert.ErrorIs(t, err, byron.ErrBodyProofMismatch)
+}
+
+// TestByronEpochSscStateRejectsUntaggedCommitmentsSet confirms the same
+// tag-258 enforcement applies to CommitmentsPayload's commitments field
+// (ssccomms = #6.258([* ssccomm]) per the CDDL), not just VSS certificates:
+// decodeIdentitySet is the shared decode point for both fields, so a fix
+// scoped there covers commitments and certificates alike rather than
+// needing a duplicated per-caller check.
+func TestByronEpochSscStateRejectsUntaggedCommitmentsSet(t *testing.T) {
```
wolf31o2, line 751: This test does not prove that malformed commitments are rejected when the proof matches the malformed bytes: it builds the block with the helper's placeholder/stale proof, so ValidateBodyProof can fail merely because the hash does not match. Please forge the proof using blake2b256(untaggedComms) plus a valid certificate hash, then assert rejection. Without that case, the production code at sscstate.go:351 can accept an untagged array because it hashes rest[0] directly and never invokes the commitments-set decoder. The Haskell Byron decoder requires commitments to be a tag-258 set ([source](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/byron/ledger/impl/src/Cardano/Chain/Ssc.hs#L77-L95)); see the [CommitmentsMap golden vector](https://github.com/IntersectMBO/cardano-ledger/tree/master/eras/byron/ledger/impl/golden/cbor/ssc/CommitmentsMap).

wolf31o2 (dismissed):
The SSC proof validation now has real mainnet coverage, but the primary payload shape is still not enforced before hashing. See the inline comments.

ledger/byron/bodyproof.go
```diff
@@ -46,35 +47,43 @@ const (
...
-	// in internal/testdata the payload part hashes to 25777aca... while the
-	// header records d36a2619..., so the two are not the same bytes. Modelling
-	// SscPayload well enough to reproduce it is a separate piece of work, and
-	// guessing the encoding from the one available fixture -- whose SSC
-	// payload is empty -- would give false assurance for the non-empty case.
-	//
-	// The consequence is that an alteration confined to the SSC payload is not
-	// detected here. SSC carries shared-seed material, not transactions, so
-	// transaction contents remain fully covered by the tx proof above.
+	if err := b.ValidateSscProof(); err != nil {
```
chrisguiney: **Blocking (blast radius, not correctness):** adding the SSC hash check here makes it decode-gating for every Byron main block, and that's a bigger step than the surrounding proofs represent.

`ValidateBodyProof` is called by default from `NewByronMainBlockFromCbor` (`ledger/byron/byron.go:1330`, on unless `SkipBodyHashValidation`). So an `ssc_proof` mismatch on any real block doesn't fail a validation pass — the block fails to *decode*.

In dingo specifically, the network-ingest paths don't opt out:

| Path | `SkipBodyHashValidation` |
|---|---|
| `ouroboros/chainsync.go:1442` | not set — validation **on** |
| `ouroboros/blockfetch.go:134` | not set — validation **on** |
| `database/models/block.go:42` | not set — validation **on** |
| `api/utxorpc/query.go:785`, `watch.go:52` | not set — validation **on** |
| `ledger/verify_header.go:168`, `internal/node/backfill.go:933`, `internal/node/load.go:726`, `mithril/sync_gap.go:476` | set — skipped |

So the paths that pull Byron blocks off the wire during a from-genesis sync, and the paths that re-read stored blocks, all run this.

What makes that harder to accept than the transaction/delegation/update proofs already validated above is that **there's no upstream oracle for this one**. Modern cardano-ledger doesn't compute it at all:

```haskell
data SscProof = SscProof              -- unit type
instance DecCBOR SscProof where
  decCBOR = do { dropSscProof; pure SscProof }
```

`dropSscProof` `dropBytes` the hashes, and `EncCBOR` emits a hardcoded `CertificatesProof` constant (`d36a2619…ad6c`) regardless of payload. So "mainnet has run on cardano-node for years" — the argument that makes the other body proofs safe to enforce — says nothing about whether a recomputed `ssc_proof` matches every historical block. cardano-node would accept a block this code rejects.

Against that, the current evidence is 3 real blocks (epoch 1, slots 21601-21603), all `CommitmentsPayload`, plus one certificates case — out of ~5M Byron blocks, four payload types, 208 epochs. The commit history here also shows two hashing bugs found *during* this PR, which is fair evidence the construction is subtle rather than settled.

I'm not suggesting dropping any of this. A proportionate shape would be either:

- keep `ValidateBodyProof`'s SSC handling structural-only and expose the hash check as opt-in (the `SkipBodyHashValidation` pattern already in `VerifyConfig` is the natural precedent), flipping the default once evidence is broad; or
- gate flipping the default on a replay across a large contiguous span of real Byron blocks covering all four payload types — the strongest possible result, and one that would also retire the "no oracle" problem for good.

Either way the validator, the golden vectors, and the bug fixes all stay.

ledger/byron/sscstate.go
```diff
@@ -0,0 +1,710 @@
...
+// 21601, 21602, 21603), each block's own real header ssc_proof commitments
+// hash matches blake2b256 of *that block's own* commitments set bytes
+// alone -- not a hash of any multi-block union, and not a hash of a
+// stakeholder-keyed map rebuilt from that set. The same holds for a real
+// OpeningsPayload block (slot 30240) against its own openings map. A real
+// CommitmentsPayload block carrying 7 VSS certificates from 7 distinct
+// stakeholders (slot 129601, epoch 6) likewise matches its own certificate
+// set, but that block happens to be its epoch's *first* main block, so this
+// vector alone cannot distinguish "this block's own certificates" from "the
+// epoch's accumulated certificates so far" for the certificates field
```
chrisguiney, line 259: **Blocking (evidence gap):** this comment names the exact hole, and it happens to sit under the one design decision the whole PR turns on.

The 7-certificate vector (slot 129601) is its epoch's *first* main block, so as written here it "cannot distinguish 'this block's own certificates' from 'the epoch's accumulated certificates so far'". Since this PR's central change is exactly that pivot — from the epoch-accumulated model #1940 assumed, to block-local — the certificates field is the one place where the two models are still observationally identical in every fixture present.

The type-level argument from cardano-sl below is a good reason to *believe* block-locality, and I don't think it's wrong. But it's the only thing carrying the certificates case right now, and a type signature doesn't pin a hash.

A vector from a **non-first main block in an epoch that already carries certificates** would settle it in one test: under block-local it hashes only that block's own certificate set; under epoch-accumulated it wouldn't. If such a block turns out to be hard to find on mainnet, that's itself worth recording here — it would mean the distinction is unobservable in practice, which is a much more comfortable place to be than "untested".

Credit for flagging this yourself rather than letting it pass silently — that's what made it findable.

internal/bench/regression_test.go
```diff
@@ -480,7 +480,15 @@ func TestRegressionCBOR(t *testing.T) {
...
-		"Byron":   550,  // Baseline: 500
+		// Byron baseline moved from 500 to 557 on 2026-08-06: decoding a
+		// Byron main block now fully validates ssc_proof's real hashes
+		// (see ledger/byron's checkSscProofLocal), not just its structural
+		// shape -- a genuine, deliberate increase in per-block decode work
+		// in exchange for actually detecting SSC-payload substitution,
+		// which was previously undetected. See
+		// TestByronEpochSscStateRealMainnet in ledger/byron for why this
+		// is now known to be correct and necessary.
+		"Byron":   615,  // Baseline: 557
```
chrisguiney: Non-blocking, and I think this is the right call — just flagging it stays visible.

500 → 557 baseline is ~11% more allocation on every Byron main block decode, and unlike most of this suite that cost lands ~5M times during a from-genesis sync rather than once. The comment justifies it well and the trade (detecting SSC-payload substitution that previously went undetected) is a reasonable thing to spend it on.

Worth being aware that if the check ends up opt-in per the other comment, this baseline should move back down for the default path, so the regression test keeps measuring what operators actually run.

chrisguiney (changes requested):
Thanks for the depth here — the cardano-sl citations, the golden vectors pulled from cardano-ledger, and the two real hashing bugs you found and fixed along the way are all genuinely good work, and the doc comments do an unusually honest job of recording what is and isn't established. @wolf31o2's two earlier findings (non-empty `canonicalMapHash` parity, and primary-field wire-shape validation) do look addressed by the golden vectors and `cda054c`.

My concern is a different one, and it's about blast radius rather than about whether the hashing is right.

This change moves `ssc_proof` from unvalidated to **decode-gating**: `ValidateBodyProof` runs by default inside `NewByronMainBlockFromCbor`, so a mismatch means the block fails to decode rather than failing a separable validation step. And unlike the transaction/delegation/update proofs it now sits alongside, there is no upstream implementation computing this value to check against — modern cardano-ledger decodes `SscProof` as a unit type, drops the hashes, and re-encodes a hardcoded constant. So the usual safety net ("mainnet has been running on cardano-node for years, so the chain must satisfy this") doesn't apply to this particular field.

That combination — first implementation of a check, no reference oracle, hard failure at decode time, on ~5M historical blocks — is what I'd like to see de-risked before it ships on by default. Details inline, along with a concrete suggestion that keeps all of this work.

There's also an issue-alignment point worth settling explicitly. #1940 asks for a *stateful* API taking epoch SSC state, premised on SSC hashes being unverifiable from a single block. This PR concludes the opposite and removes that dependency — `ByronEpochSscState` survives only for stakeholder tracking and is never consulted by `ValidateBodyProof`. If the block-local conclusion is right (and the evidence you've assembled is decent), that's a *better* outcome than the issue asked for — but `Closes #1940` currently reads as though the stated acceptance criteria were met, and none of the four are. Worth rewriting the close to say the premise was wrong and why, so the next person doesn't read the criteria as satisfied.

**Verified clean:** build, `go test -race ./ledger/byron/... ./consensus/byron/...`, gofmt, `golangci-lint` (0 issues), and all 10 CI checks. No concurrency surface at all — the new code is pure functions over block bytes, with no goroutines, channels, or shared state.

chrisguiney (approved):

PR #1962, opened 2026-08-07
fix(consensus)!: apply Genesis density-first selection for deep forks

### LINKED ISSUE #1936: consensus: apply Genesis density-first selection for deep forks
PraosChainSelector.CompareWithDensity compares the ordinary chain ordering first and only compares density when the ordinary comparison ties. Its documentation says it is for deep forks, but the Genesis rule is different: forks no deeper than k use longest-chain selection; deeper forks must use the density rule.

The repository has a separate consensus/genesis selector with density logic, but the public PraosChainSelector path remains unsafe for deep forks. IsDeepFork exists but is not used by CompareWithDensity or PreferredWithDensity.

Impact: a longer but less dense deep fork can be selected over a shorter, denser candidate.

References:
- https://github.com/blinklabs-io/gouroboros/blob/main/consensus/selection.go
- https://ouroboros-consensus.cardano.intersectmbo.org/docs/references/glossary/

Acceptance criteria:
- Make the selection path explicitly distinguish forks at most k deep from forks deeper than k.
- Use longest-chain comparison within k and density-first comparison beyond k.
- Add a regression test where the longer deep fork has lower density and must lose.

adarshpuri20:
## Semantics first — one point to ratify (@wolf31o2)

#1936 asks the selection path to "use longest-chain comparison within k and density-first comparison beyond k". While implementing it we checked the shipped canonical implementation, and its shape is slightly different in a way that matters for consensus safety:

- **Forks requiring rollback of more than k blocks are refused, never density-adopted.** `Ouroboros.Consensus.Protocol.Abstract`: *"we never switch to chains that fork off more than @k@ blocks ago"*; the ChainSync client terminates such peers with `ForkTooDeep`. And k bounds **blocks**, not slots — `Ouroboros.Consensus.Config.SecurityParam`: *"NOTE: This talks about the number of /blocks/ we can roll back, not the number of /slots/"*.
- **The density rule chooses among syncing candidates** (GDD governor, `Ouroboros/Consensus/Genesis/Governor.hs`): each candidate suffix is clipped to the genesis window after the common intersection — `firstSlotAfterGenesisWindow = succWithOrigin loeIntersectionSlot + SlotNo sgen`, `AF.splitAtSlot firstSlotAfterGenesisWindow candidateSuffix` — and compared by **integer block count** (`lowerBound = fromIntegral $ AF.length clippedFragment`), with equal density a loss (`guard $ lb1 >= (if idling0 then lb0 else ub0)`).
- **The window sgen is the stability window 3k/f, with ceiling division.** cardano-ledger `eras/shelley/impl/src/Cardano/Ledger/Shelley/StabilityWindow.hs`: `computeStabilityWindow k asc = ceiling $ (3 * fromIntegral k) /. f`, consumed by ouroboros-consensus `shelleyEraParams` as `eraGenesisWin = GenesisWindow stabilityWindow` (mainnet: 3·2160/0.05 = 129,600 slots).

So this PR implements the canonical decomposition rather than the issue's literal routing: `CompareWithDensity` is the Genesis ordering for syncing candidates (density-first within the window, equal density falls through to the ordinary Praos comparison — the same adaptation Dingo's fork resolution uses, since a pure ordering cannot "disconnect both" like the GDD does on ties), and `IsDeepFork` is the block-based depth predicate that routes between the two. **The issue's own regression criterion — a longer deep fork with lower density must lose — passes either way** (`TestCompareWithDensityIssue1936Regression`). If you'd prefer the literal ">k ⇒ density" routing instead, say so and we'll adjust — but we believe this shape is the consensus-safe one.

## The three defects fixed

1. **Density was unreachable** — `CompareWithDensity` ran the longest-chain comparison first and consulted density only on a full length+VRF tie, so the longer-but-sparser adversarial chain always won (the exact shape #1936 describes). Now density-first within the genesis window.
2. **Wrong metric** — density was a `float64` blocks/slots ratio over the whole fork suffix. Now the canonical integer block count within the window (slots `forkSlot+1..forkSlot+windowSlots` inclusive, matching `densityDisconnect`'s clipping), supplied through a new optional `WindowBlockCounter` interface implemented by `WindowedChainTip`. The selector takes an explicit `GenesisWindowSlots` (derive with `genesis.ComputeGenesisWindow`).
3. **Wrong unit** — `IsDeepFork` compared a **slot** distance against k (≈20× too eager at mainnet f=0.05). `IsDeepFork` now takes a `ForkPoint` (the intersection's slot and block number) plus the current tip's block number and compares block distances, and it is wired into both `CompareWithDensity` and `PreferredWithDensity`.

## Separable sub-change: `ComputeGenesisWindow` floor → ceiling

While pinning the window semantics we found `genesis.ComputeGenesisWindow` used floor division (`big.Int.Div`), but canonical is **ceiling** (`computeStabilityWindow` above). Mainnet is provably unaffected (3k/f is exactly integral there); the result differed by one slot on any non-integral 3k/f. Fixed with a citation in-code, plus a guard rejecting negative active-slot coefficients. **Happy to split this into its own PR if you prefer single-concern changes.**

## Verification

- **Fail-before / pass-after**: probes on `main` demonstrate the inverted order (`CompareWithDensity` returned −1 for denser-vs-longer on a deep fork) and the slot/block misclassification; the new suite passes after. All tests run with `-race` in a clean container (Go 1.25.8).
- New tests: issue regression (both argument orders), exact window boundary (`forkSlot+window` in, `+window+1` out, `forkSlot` itself out), equal-density fallthrough end-to-end through the VRF tiebreak, zero-window fallback, `IsDeepFork` at k vs k+1 in blocks, ceiling division (direct, via `NewGenesisSelector`, degenerate coefficients), mainnet 129,600 cross-check.
- **`Compare` is byte-identical** to `main` — the ordinary/shallow selection path is provably unchanged, locked by the retained tests plus new reverse-VRF and empty-VRF cases.
- `make test` (full race suite) and `make lint` (golangci-lint v2) green.
- Independently re-verified by a second, isolated review pass that wrote its own tests from the issue text (own constants, both directions, boundary cases) and confirmed the pre-patch code fails its regression assertion while the patch passes it.

## API changes (v0)

The exported `ChainTip` interface is **unchanged** — `consensus/consensus.go` is not in this diff — so existing implementations keep compiling and behave as they do today. `NewSimpleChainTipWithDensity` is retained. No exported symbol is removed.

Three method signatures change, to carry the depth inputs the routing needs:

```
IsDeepFork(forkSlot, currentSlot uint64)
  -> IsDeepFork(fork ForkPoint, tipBlockNumber uint64)
CompareWithDensity(a, b ChainTip, forkSlot, currentSlot uint64)
  -> CompareWithDensity(a, b ChainTip, fork ForkPoint, tipBlockNumber uint64)
PreferredWithDensity(candidates []ChainTip, forkSlot, currentSlot uint64)
  -> PreferredWithDensity(candidates []ChainTip, fork ForkPoint, tipBlockNumber uint64)
```

The update is mechanical: pass `ForkPoint{Slot: forkSlot, BlockNumber: forkBlockNumber}` and the current selection's tip block number in place of `currentSlot`. New: `ForkPoint`, `WindowBlockCounter`, `WindowedChainTip`, `NewWindowedChainTip`, `NewPraosChainSelectorWithWindow`.

## Disclosure

This contribution was developed with AI assistance (Claude), under human direction, with the verification described above run in a clean container environment. 

Closes #1936

---
## Summary by cubic
Applies Genesis density‑first ordering only to deep forks and routes depth by blocks, not slots. Fixes #1936 so a longer but sparser deep fork loses while shallow forks keep ordinary Praos behavior.

- `consensus`: `IsDeepFork(ForkPoint, tipBlockNumber)` gates selection; shallow (≤k blocks) uses `Compare`, deep uses density‑first in `CompareWithDensity`/`PreferredWithDensity` with tie‑back to Praos.
- Genesis metric: prefer higher integer block count within the configured window (`BlocksInWindow(forkSlot, windowSlots)`); otherwise fall back to `ChainTip.Density` with a one‑time warning.
- Types: `SimpleChainTip` no longer claims window counting; new `WindowedChainTip` stores per‑block slots and derives a legacy ratio for mixed comparisons.
- `genesis.ComputeGenesisWindow`: ceiling for `3k/f`; rejects non‑positive `f`.
- Tests: add invariant sweep that shallow forks never consult density, an end‑to‑end Genesis selection scenario, and “zero blocks in window” loses; retain boundary and fallback cases.

**Migration**
- Update call sites:
  - `IsDeepFork(forkSlot, currentSlot)` → `IsDeepFork(ForkPoint{Slot, BlockNumber}, tipBlockNumber)`
  - `CompareWithDensity(a,b,forkSlot,currentSlot)` → `CompareWithDensity(a,b,ForkPoint,tipBlockNumber)`
  - `PreferredWithDensity(cands,forkSlot,currentSlot)` → `PreferredWithDensity(cands,ForkPoint,tipBlockNumber)`
- To use the canonical metric, compute the window with `genesis.ComputeGenesisWindow(k,f)`, build with `NewPraosChainSelectorWithWindow`, and implement `WindowBlockCounter`. Tests relying on window counting should use `NewWindowedChainTip`. Without the counter, comparisons use the legacy ratio.

<sup>Written for commit 2c2bae76e223a05ca1a7af6d3d3578b3236d4655. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/1962?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Improved chain selection during synchronization using Genesis density comparisons within a configured window.
  * Added block-window counting for more accurate chain comparisons.
  * Added rollback-limit checks based on block depth.

* **Bug Fixes**
  * Genesis windows now round up correctly and handle invalid coefficients safely.
  * Improved tie handling and fallback behavior for equal or unavailable density data.

* **Documentation**
  * Updated architecture and consensus documentation to reflect chain selection, density comparison, and rollback behavior.

### DIFF
--- a/consensus/genesis/genesis_test.go
+++ b/consensus/genesis/genesis_test.go
@@ -17,6 +17,8 @@ package genesis
 import (
 	"math/big"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
 )
 
 // testGenesisConfig returns a GenesisConfig for testing with mainnet-like parameters.
@@ -47,33 +49,34 @@ func TestComputeGenesisWindow(t *testing.T) {
 func TestComputeGenesisWindowCeiling(t *testing.T) {
 	// k=2160, f=7/100: 3*2160*100/7 = 648000/7 = 92571.43... -> 92572.
 	window := ComputeGenesisWindow(2160, big.NewRat(7, 100))
-	if window != 92572 {
-		t.Errorf(
-			"expected ceiling(648000/7) = 92572, got %d",
-			window,
-		)
-	}
+	assert.Equal(
+		t, uint64(92572), window, "expected ceiling(648000/7) = 92572",
+	)
 
 	// Exact division is unaffected: k=10, f=1/2 -> 3*10*2 = 60.
 	window = ComputeGenesisWindow(10, big.NewRat(1, 2))
-	if window != 60 {
-		t.Errorf("expected 60, got %d", window)
-	}
+	assert.Equal(t, uint64(60), window, "expected exact division 60")
 }
 
 // TestComputeGenesisWindowDegenerateCoefficients pins the guard: nil, zero,
 // and negative active-slot coefficients all yield a zero window rather than
 // a garbage value.
 func TestComputeGenesisWindowDegenerateCoefficients(t *testing.T) {
-	if got := ComputeGenesisWindow(2160, nil); got != 0 {
-		t.Errorf("nil coefficient: expected 0, got %d", got)
-	}
-	if got := ComputeGenesisWindow(2160, big.NewRat(0, 1)); got != 0 {
-		t.Errorf("zero coefficient: expected 0, got %d", got)
-	}
-	if got := ComputeGenesisWindow(2160, big.NewRat(-1, 20)); got != 0 {
-		t.Errorf("negative coefficient: expected 0, got %d", got)
-	}
+	assert.Zero(
+		t,
+		ComputeGenesisWindow(2160, nil),
+		"nil coefficient: expected 0",
+	)
+	assert.Zero(
+		t,
+		ComputeGenesisWindow(2160, big.NewRat(0, 1)),
+		"zero coefficient: expected 0",
+	)
+	assert.Zero(
+		t,
+		ComputeGenesisWindow(2160, big.NewRat(-1, 20)),
+		"negative coefficient: expected 0",
+	)
 }
 
 // TestNewGenesisSelectorDerivedWindowCeiling pins the derived-window path
@@ -85,12 +88,10 @@ func TestNewGenesisSelectorDerivedWindowCeiling(t *testing.T) {
 		ActiveSlotCoeff: big.NewRat(7, 100),
 	})
 	// 3*2160*100/7 = 648000/7 = 92571.43... -> 92572.
-	if selector.config.GenesisWindow != 92572 {
-		t.Errorf(
-			"expected derived window 92572, got %d",
-			selector.config.GenesisWindow,
-		)
-	}
+	assert.Equal(
+		t, uint64(92572), selector.config.GenesisWindow,
+		"expected derived window 92572",
+	)
 }
 
 func TestGenesisConfig(t *testing.T) {
--- a/consensus/genesis/genesis_test.go
+++ b/consensus/genesis/genesis_test.go
@@ -40,6 +40,59 @@ func TestComputeGenesisWindow(t *testing.T) {
 	}
 }
 
+// TestComputeGenesisWindowCeiling pins the rounding to CEILING division,
+// matching cardano-ledger's computeStabilityWindow
+// ("ceiling $ (3 * fromIntegral k) /. f"): whenever 3k/f is not integral,
+// the window rounds up, never down.
+func TestComputeGenesisWindowCeiling(t *testing.T) {
+	// k=2160, f=7/100: 3*2160*100/7 = 648000/7 = 92571.43... -> 92572.
+	window := ComputeGenesisWindow(2160, big.NewRat(7, 100))
+	if window != 92572 {
+		t.Errorf(
+			"expected ceiling(648000/7) = 92572, got %d",
+			window,
+		)
+	}
+
+	// Exact division is unaffected: k=10, f=1/2 -> 3*10*2 = 60.
+	window = ComputeGenesisWindow(10, big.NewRat(1, 2))
+	if window != 60 {
+		t.Errorf("expected 60, got %d", window)
+	}
+}
+
+// TestComputeGenesisWindowDegenerateCoefficients pins the guard: nil, zero,
+// and negative active-slot coefficients all yield a zero window rather than
+// a garbage value.
+func TestComputeGenesisWindowDegenerateCoefficients(t *testing.T) {
+	if got := ComputeGenesisWindow(2160, nil); got != 0 {
+		t.Errorf("nil coefficient: expected 0, got %d", got)
+	}
+	if got := ComputeGenesisWindow(2160, big.NewRat(0, 1)); got != 0 {
+		t.Errorf("zero coefficient: expected 0, got %d", got)
+	}
+	if got := ComputeGenesisWindow(2160, big.NewRat(-1, 20)); got != 0 {
+		t.Errorf("negative coefficient: expected 0, got %d", got)
+	}
+}
+
+// TestNewGenesisSelectorDerivedWindowCeiling pins the derived-window path
+// through NewGenesisSelector with a non-integral 3k/f, so the constructor
+// inherits the ceiling semantics too.
+func TestNewGenesisSelectorDerivedWindowCeiling(t *testing.T) {
+	selector := NewGenesisSelector(GenesisConfig{
+		SecurityParam:   2160,
+		ActiveSlotCoeff: big.NewRat(7, 100),
+	})
+	// 3*2160*100/7 = 648000/7 = 92571.43... -> 92572.
+	if selector.config.GenesisWindow != 92572 {
+		t.Errorf(
+			"expected derived window 92572, got %d",
+			selector.config.GenesisWindow,
+		)
+	}
+}
+
 func TestGenesisConfig(t *testing.T) {
 	config := testGenesisConfig()
 
--- a/consensus/selection.go
+++ b/consensus/selection.go
@@ -370,6 +370,10 @@ func (p *PraosChainSelector) PreferredWithDensity(
 }
 
 // SimpleChainTip is a simple implementation of ChainTip for testing.
+//
+// It deliberately does NOT implement WindowBlockCounter: it holds no
+// per-block slots, so it cannot answer a window count. See
+// WindowedChainTip for a tip that can.
 type SimpleChainTip struct {
 	slot        uint64
 	blockNumber uint64
@@ -377,9 +381,6 @@ type SimpleChainTip struct {
 	// For the legacy density ratio
 	blocksAfterFork uint64
 	slotsAfterFork  uint64
-	// blockSlots holds the slot of each block on this chain, used for the
-	// canonical window metric (see WindowBlockCounter).
-	blockSlots []uint64
 }
 
 // NewSimpleChainTip creates a new SimpleChainTip.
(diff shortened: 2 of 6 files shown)


adarshpuri20:
This is the semantics point the PR description leads with (the section asking @wolf31o2 to ratify), so tagging it here too. The canonical decomposition deliberately does not route between comparison modes by rollback depth:

- A candidate whose adoption needs a rollback of more than k **blocks** is refused, never compared — ouroboros-consensus terminates such peers with `ForkTooDeep` ("we never switch to chains that fork off more than @k@ blocks ago"). In this PR that guard is `ExceedsMaxRollback`, and its doc states a true result means "refuse the candidate," not "switch comparison rule."
- The density rule is not a deep-fork comparator: it chooses among syncing candidates (the GDD governor), whatever their depth. So mode selection is intentionally at the caller — one of the two options you suggest — with the refusal guard as a separate check, matching upstream. The k / k+1 boundary tests exist in block units in `TestExceedsMaxRollback`.

Routing `CompareWithDensity` by rollback depth would implement #1936's literal ">k ⇒ density" reading, which we believe is not consensus-safe (a >k fork that wins a density comparison would still violate the immutability of the k-deep block). If the maintainers prefer the literal routing regardless, we'll implement it — that's exactly the ratification question in the PR description.

consensus/genesis/genesis_test.go
```diff
@@ -40,6 +40,59 @@ func TestComputeGenesisWindow(t *testing.T) {
...
+		ActiveSlotCoeff: big.NewRat(7, 100),
+	})
+	// 3*2160*100/7 = 648000/7 = 92571.43... -> 92572.
+	if selector.config.GenesisWindow != 92572 {
+		t.Errorf(
+			"expected derived window 92572, got %d",
+			selector.config.GenesisWindow,
+		)
+	}
+}
```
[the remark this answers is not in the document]
adarshpuri20: Thanks — addressed in 4ca464a. The bare `t.Error`/`t.Errorf`/`t.Fatal` calls in this PR's test hunks (`consensus/genesis/genesis_test.go` and `consensus/selection_test.go`) are now testify `assert`/`require`, per the policy in AGENTS.md. `require` is reserved for the two load-bearing preconditions — the mainnet stability-window cross-check and the non-nil selection dereferenced on the following line — while `assert` covers the independent checks. It's a test-only change: no production code touched, and the `-race` suite stays green.

consensus/selection.go
```diff
@@ -113,24 +177,28 @@ func (p *PraosChainSelector) CompareWithDensity(
...
+	if aBlocks != bBlocks {
+		if aBlocks > bBlocks {
+			return 1
+		}
 		return -1
 	}
 
-	return 0
+	// Equal density: fall through to the ordinary Praos comparison.
+	return p.Compare(a, b)
```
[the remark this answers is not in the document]
adarshpuri20: This is the semantics point the PR description leads with (the section asking @wolf31o2 to ratify), so tagging it here too. The canonical decomposition deliberately does not route between comparison modes by rollback depth:

- A candidate whose adoption needs a rollback of more than k **blocks** is refused, never compared — ouroboros-consensus terminates such peers with `ForkTooDeep` ("we never switch to chains that fork off more than `k` blocks ago"). In this PR that guard is `ExceedsMaxRollback`, and its doc states a true result means "refuse the candidate," not "switch comparison rule."
- The density rule is not a deep-fork comparator: it chooses among syncing candidates (the GDD governor), whatever their depth. So mode selection is intentionally at the caller — one of the two options you suggest — with the refusal guard as a separate check, matching upstream. The k / k+1 boundary tests exist in block units in `TestExceedsMaxRollback`.

Routing `CompareWithDensity` by rollback depth would implement #1936's literal ">k ⇒ density" reading, which we believe is not consensus-safe (a >k fork that wins a density comparison would still violate the immutability of the k-deep block). If the maintainers prefer the literal routing regardless, we'll implement it — that's exactly the ratification question in the PR description.

consensus/selection.go
```diff
@@ -113,24 +177,28 @@ func (p *PraosChainSelector) CompareWithDensity(
...
 
-	// For deep forks, compare density
-	aDensity := a.Density(forkSlot)
-	bDensity := b.Density(forkSlot)
-
-	if aDensity > bDensity {
-		return 1
-	}
-	if bDensity > aDensity {
+	// Genesis rule: density within the window decides first.
```
wolf31o2: This removes the depth gate required by issue #1936. With a configured window, every call is density-first, including shallow forks; a shorter/denser shallow candidate can therefore beat a longer chain that should win by ordinary Praos comparison. `PreferredWithDensity` also no longer accepts current depth, so callers cannot enforce the distinction. Please restore explicit shallow-versus-deep/syncing routing with a consistent k unit, and add a regression for a shorter/denser shallow fork.

consensus/selection.go
```diff
@@ -170,26 +241,42 @@ func (p *PraosChainSelector) PreferredWithDensity(
...
+// Callers should treat a true result as "refuse the candidate", not as a
+// signal to route to a different comparison rule; the Genesis density rule
+// (CompareWithDensity) is about choosing among syncing candidates, not
+// about adopting deep forks.
+//
+// This method replaces IsDeepFork, which compared a SLOT distance against k
+// and therefore misclassified (~20x too eager at mainnet f=0.05) while also
+// suggesting the wrong role ("deep ⇒ use density" instead of "exceeds max
+// rollback ⇒ refuse"). The rename makes the semantic change compile-visible.
+func (p *PraosChainSelector) ExceedsMaxRollback(
```
wolf31o2: This predicate is not wired into `CompareWithDensity` or `PreferredWithDensity`; the submitted branch has no production caller for it. Consequently the claimed rollback refusal is not enforced, and this helper cannot make the selection path safe by itself. Please integrate the check into the actual selection entry point, including the data needed to compute depth, or redesign the API so invalid/deep candidates cannot be silently selected. Add an end-to-end public-selection test.

consensus/consensus.go
```diff
@@ -145,8 +145,19 @@ type ChainTip interface {
 	BlockNumber() uint64
 	// VRFOutput returns the VRF output of the tip block
 	VRFOutput() []byte
-	// Density returns block density for the chain (blocks / slots) - for deep fork comparison
-	Density(fromSlot uint64) float64
+	// BlocksInWindow returns the number of blocks on this chain whose slots
```
wolf31o2: This breaks the exported `ChainTip` interface: downstream implementations now need `BlocksInWindow`, while `Density` and related helpers are removed. A repository search cannot establish that there are no external consumers. Please preserve compatibility where practical, or add an explicit migration/deprecation path and coordinate the breaking release before merging.

wolf31o2 (changes requested):
## Changes requested

I reviewed commit `4ca464aefbcee2cf837078ac2765873ab1242666` locally. The branch passes Go tests, consensus race tests, vet, and lint, but I cannot approve it: the selection contract is not implemented coherently and the PR conflicts with current `main` because PR 1970 changed the same selection code.

The main correctness issue is that this patch removes shallow/deep routing. `CompareWithDensity` is density-first whenever a window is configured, while `PreferredWithDensity` has no current-depth input. Callers can therefore select density for shallow forks, where ordinary longest-chain selection is required. `ExceedsMaxRollback` is only a helper and is not used by the selection path.

### Agent-ready prompt

> Rebase onto current `main` and resolve the conflict with PR 1970 without reverting its behavior. Make depth handling explicit and impossible to misuse: shallow forks must use ordinary Praos comparison; only eligible deep/syncing candidates may use Genesis density comparison. If k is block-based, expose the required current/fork block heights and wire the check into the actual selection entry point; if the issue contract is slot-based, preserve the current-slot input. Do not leave `ExceedsMaxRollback` unused while claiming rollback enforcement.
>
> Apply the same contract to `PreferredWithDensity`. Add tests for shorter/denser shallow forks, longer/sparser deep forks, exact k and k+1 boundaries, equal-density ties, zero/invalid windows, and uint64 boundary arithmetic. Re-check the exported `ChainTip` API break and provide compatibility or an explicit migration plan. Run `make test`, `make lint`, and `go build ./...`; keep unrelated changes out.

Please update the branch and request re-review after resolving these issues.

wolf31o2 (approved):
Current head includes explicit block-depth routing, windowed density handling, boundary coverage, and passing required checks.

PR #1963, opened 2026-08-07
fix(consensus): range-reduce ln/exp arguments in leader threshold calculation

### LINKED ISSUE #1941: consensus: differential-test leader threshold cutoff behavior
consensus.CertifiedNatThresholdWithMode computes the TPraos/CPraos threshold with big.Float arithmetic and fixed 100-term Taylor approximations for log and exp. The upstream TPraos implementation uses a bounded/conclusive comparison around the exponential cutoff.

Existing tests cover algebraic accuracy and broad leader-election behavior, but no upstream boundary vectors prove identical eligibility for outputs immediately below, at, and above the cutoff.

Risk: a rounding or approximation difference at the eligibility boundary could cause nodes to disagree about slot leadership.

References:
- https://github.com/blinklabs-io/gouroboros/blob/main/consensus/threshold.go
- https://cardano-ledger.cardano.intersectmbo.org/cardano-protocol-tpraos/Cardano-Protocol-TPraos-BHeader.html

Acceptance criteria:
- Add differential vectors covering both TPraos and CPraos, multiple active-slot coefficients, stake ratios, and values adjacent to the cutoff.
- Compare against the upstream implementation or an independently proven exact/bounded reference.
- If any discrepancy is found, replace the approximation or make the comparison provably conservative.

cryptodj413:
## Summary

Fixes #1941.

`CertifiedNatThresholdWithMode`'s `lnOneMinusFloat`/`expFloat` used a fixed 100-term Taylor series applied directly to the raw argument. This loses precision badly for active-slot coefficients near 0.9-0.99: several percent divergence in the resulting eligibility threshold, and up to 33 orders of magnitude error in `expFloat` at `x=-50`. Since the number of Taylor terms needed to converge grows with the magnitude of the input, a fixed term count that's adequate near zero is nowhere near enough once the raw ln/exp arguments get large in this range.

## Root cause

Both helpers fed the raw (unreduced) argument straight into a fixed-length Taylor expansion:
- `lnOneMinusFloat` needs `ln(1-x)` for `x` close to 1 when the active-slot coefficient is high, which drives the Taylor argument to large magnitude where the series converges too slowly to matter within 100 terms.
- `expFloat` sees correspondingly large negative arguments (e.g. `x=-50`) from the log-domain threshold math, where a direct Taylor sum without argument reduction underflows/loses precision catastrophically (33 orders of magnitude off in the case observed).

Neither function decoupled the required term count from the size of the input; both implicitly assumed inputs small enough that 100 terms was "enough," which breaks down well within the actual range of active-slot coefficients cardano-node uses.

## Fix

Range-reduce the arguments before applying the series, so the number of terms needed is bounded independently of input magnitude and independent of the 1024-bit `thresholdPrecision`:
- **ln**: `MantExp`-based mantissa normalization, then the fast-converging `ln(m) = 2*atanh((m-1)/(m+1))` series on the normalized mantissa.
- **exp**: halve-then-square (compute `exp(x/2)` via the series, then square the result), repeated as needed to bring the effective argument into the fast-converging range.

## Testing

Added `consensus/threshold_boundary_test.go`: differential/boundary tests against an independently-implemented reference (continued-fraction `ln` + ceil-scaled Taylor `exp`, mirroring upstream cardano-ledger's `Cardano.Ledger.NonIntegral`), covering both TPraos and CPraos modes across multiple active-slot coefficients and stake ratios, plus values immediately below/at/above the true eligibility cutoff.

Confirmed via `git stash` that these tests fail against the pre-fix code and pass with the fix applied.

- `go build ./...` — clean
- `go vet ./...` — clean
- `gofmt -l .` — clean
- `golangci-lint run ./...` — 0 issues
- `go test -race ./...` — all pass, including conformance suite and allocation-regression benchmarks
- `make test` — clean
- `make lint` — clean

## Test plan

- [x] `make test`
- [x] `make lint`
- [x] New differential/boundary tests in `consensus/threshold_boundary_test.go` confirmed to fail pre-fix (via `git stash`) and pass post-fix

---
## Summary by cubic
Improves leader threshold accuracy and boundary handling by range‑reducing ln/exp, computing 1−f exactly, and using exact‑rational plus interval‑bounded evaluation to avoid off‑by‑one and precision loss. Precision escalation now errors explicitly if the 16,384‑bit cap is reached; the legacy `CertifiedNatThreshold` still returns 0 on invalid input.

- **Bug Fixes**
  - Compute 1−f as exact `big.Rat`; domain guard: f ≤ 0 → 0, f == 1 → upper bound, f > 1 → error.
  - Exact‑rational fast path: resolve (1−f)^σ exactly via integer Nth‑root with no denominator cap.
  - Range‑reduced series: ln via MantExp + 2·atanh((m−1)/(m+1)); exp via halve‑then‑square; term counts sized from target bits.
  - Interval bounds with escalation: return threshold only when floor(lo·UB) == floor(hi·UB); otherwise double precision up to a 16,384‑bit cap, and now return an explicit error if unresolved (via extracted `escalateThreshold`, counter removed).
  - Legacy `CertifiedNatThreshold`: never returns nil; returns 0 on invalid f.

- **Tests**
  - Independent reference (continued‑fraction ln + ceil‑scaled Taylor exp); tightened boundary checks to exact equality for table cases; added cap‑reached error‑path test.
  - Coverage includes exact‑rational and large‑denominator cases (e.g., f=1−2^-4097, σ=1/4097) and near‑one f without hangs.

<sup>Written for commit 91ad5d9051e04bbd1668c2c3dc5a3ea0605b76c5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/1963?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the accuracy and reliability of threshold calculations, particularly for large active-slot coefficients.
  - Strengthened leader eligibility boundary handling.

- **Tests**
  - Added high-precision reference comparisons across supported consensus modes and stake ratios.
  - Added boundary coverage for eligibility just below, at, and above calculated thresholds.
  - Improved test readability without changing existing behavior.


cryptodj413:
Pushed a follow-up commit (b9559bd) addressing review feedback:

**Fixed (confirmed via execution):**
- **Off-by-one at exact rational cutoffs**: a full-stake pool at `f=1/2` was previously rejecting the leader value immediately below `upperBound/2` due to floating-point approximation error at that exact boundary. Added an exact `big.Int` fast path for `sigma=1` to avoid the float approximation entirely at that cutoff.
- **Missing domain guards for `activeSlotCoeff` outside `(0,1]`**: values outside this range previously fell through to the ln/exp series outside its convergence radius, silently producing wrong thresholds instead of failing loudly. Added explicit domain guards.

Also included two trivial test nits from review.

**Determined to be false positives / out of scope (not changed):**
- *"Boundary test derives values from cutoff instead of threshold, passes vacuously"* — checked algebraically: any nonzero drift between the two produces a loud assertion failure, not a vacuous pass. The test does exercise the intended comparison.
- *"Replace all `t.Error`/`t.Errorf` with testify `assert.*` in threshold_test.go"* — this PR only reformatted pre-existing stdlib-style calls and didn't introduce that pattern. The repo has a genuine mixed convention already (94 files using stdlib `testing` assertions vs 132 using testify), so this is pre-existing style debt unrelated to this diff.

`make test` and `make lint` are green on the branch.

cryptodj413:
Confirmed, this was a valid finding — `CertifiedNatThreshold` could return a nil `*big.Int` for an out-of-domain `activeSlotCoeff` (f > 1), since it discarded the error from `CertifiedNatThresholdWithMode` without checking it. Any caller invoking `.Cmp`/`.Sign`/`.Bytes`/etc. on the result without a nil check would panic.

Fixed in e88a2c1: `CertifiedNatThreshold` now checks the error and falls back to `big.NewInt(0)` (the same conservative "never lead" default already used for other degenerate input, e.g. `activeSlotCoeff.Sign() <= 0`), with a doc comment pointing callers who need to detect invalid input at `CertifiedNatThresholdWithMode` instead. Added a regression test (`TestCertifiedNatThresholdLegacyAPIAboveOneNeverNil`) confirming the wrapper returns a non-nil `big.NewInt(0)` for f > 1 rather than nil.

`make test` and `make lint` are green on the updated branch.

consensus/threshold.go
```diff
@@ -142,7 +200,8 @@ func CertifiedNatThresholdWithMode(
 	// (1-f)^σ ≈ exp(σ * ln(1-f))
 	//
 	// We use big.Float internally to avoid the O(n²) GCD normalization
-	// cost of big.Rat arithmetic over 100 Taylor series terms.
+	// cost of big.Rat arithmetic over the many Taylor series terms
+	// involved (see lnOneMinusFloat/expFloat below).
 	f := new(big.Float).SetPrec(prec).SetRat(activeSlotCoeff)
```
wolf31o2: Blocking: this conversion feeds an approximate big.Float into the final floor, so it can be one integer below an exact cutoff. For f=3/4 and sigma=1/2, (1-f)^sigma=1/2, so the exact threshold is 2^(N-1); this implementation returns 2^(N-1)-1 in both CPraos (N=256) and TPraos (N=512). Because eligibility uses a strict `< threshold`, this rejects the valid leader value 2^(N-1)-1. Please add this regression case and use an exact or directed-rounding strategy for partial-stake rational cutoffs.

consensus/threshold.go
```diff
@@ -166,62 +225,236 @@ func CertifiedNatThresholdWithMode(
...
 
+// lnOneMinusFloat computes ln(1-x) for 0 < x < 1 by delegating to
+// lnPositiveFloat on y = 1-x. See the NOTE above this section for why a
+// direct fixed-term power series in x is unsafe as x -> 1.
+func lnOneMinusFloat(x *big.Float) *big.Float {
 	prec := x.Prec()
 	one := new(big.Float).SetPrec(prec).SetInt64(1)
-	result := new(big.Float).SetPrec(prec).Set(one) // Start with 1
-	term := new(big.Float).SetPrec(prec).Set(one)   // x^n / n!
+	y := new(big.Float).SetPrec(prec).Sub(one, x)
```
wolf31o2: Blocking: SetRat at the fixed 1024-bit precision can round a valid rational f<1 to exactly 1 before this subtraction, making y zero. For example, f=(2^2000-1)/2^2000 with sigma=1/2 has the exact threshold 2^N-1, but this path produces a materially smaller value. The domain check uses the original Rat, so it does not catch this. Avoid the lossy conversion before computing 1-f, or derive precision from the Rat's distance to 1.

wolf31o2 (dismissed):
Request changes. The numerical threshold path still has correctness failures at exact cutoffs and for arbitrary-precision rational inputs; the two inline comments below are blocking.

consensus/threshold.go
cryptodj413: Fixed in 6bd161d. This case (f=3/4, sigma=1/2) is now handled by a new exact-rational fast path: (1-f)^sigma is resolved via integer Nth-root when it has an exact rational/integer result, so 1/2 = (1/2)^1 resolves exactly and the threshold returned is the exact 2^(N-1) rather than the previous 2^(N-1)-1. Added a regression test covering this exact case (and the general family of coincidental-exact-cutoff rationals) in consensus/threshold_test.go.

consensus/threshold.go
cryptodj413: Fixed in 6bd161d. 1-f is now computed as an exact big.Rat before any float conversion (new(big.Rat).Sub(one, activeSlotCoeff)), so f=(2^2000-1)/2^2000 no longer rounds to 1 during SetRat -- the exact 1-f=1/2^2000 is preserved and only converted to big.Float afterward, at precision derived from that exact value. Combined with the interval-arithmetic escalation loop (floor(lo)==floor(hi) required before returning), this case now yields the exact threshold 2^N-1. Added a regression test for this f alongside the sigma=1/2 case in consensus/threshold_test.go.

consensus/threshold.go
```diff
@@ -129,99 +161,542 @@ func CertifiedNatThresholdWithMode(
...
+// before exactOneMinusFPowerSigma gives up and defers to the escalating-
+// precision interval computation instead. For a genuine stake ratio (whose
+// reduced denominator is typically on the order of totalStake, often up to
+// ~2^64), an exact m-th root essentially never exists, and root-checking it
+// anyway would be needlessly expensive (an Exp of a value with on the order
+// of totalStake's own bit length). This cap keeps the fast-path attempt
+// itself cheap to try-and-fail for realistic inputs, while remaining large
+// enough to cover any plausible small-denominator boundary case (such as
+// the sigma=1/2 cases this generalizes from the sigma=1 fast path).
+const maxExactRootDegree = 4096
```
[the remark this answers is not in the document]
cryptodj413: Fixed in eb62987. `maxExactRootDegree` has been removed entirely — `exactIntegerNthRoot` is now a complete decision procedure for any root degree `k`: `r=1` is checked in O(1) regardless of `k`, and for `r>=2` it proves non-existence whenever `k >= bitlen(n)` instead of giving up. This closes the bug class structurally instead of raising the constant, since sigma's reduced denominator is fully attacker/operator-controlled. Added a regression test, `TestCertifiedNatThresholdSigmaDenominatorAboveExactRootCap` (f=1-2^-4097, sigma=1/4097), which reproduced the 2+ minute hang and wrong-by-one result on the old code and now passes in 0.00s.

consensus/threshold_test.go
```diff
@@ -1016,7 +1114,288 @@ func TestLnOneMinusConvergence(t *testing.T) {
...
+func TestCertifiedNatThresholdPartialStakeExactCutoff(t *testing.T) {
+	f := big.NewRat(3, 4)
+
+	cases := []struct {
+		name       string
+		mode       ConsensusMode
+		bits       int
+		upperBound *big.Int
+	}{
+		{"CPraos", ConsensusModeCPraos, vrfOutputBitsCPraos, twoTo256},
```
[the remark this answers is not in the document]
cryptodj413: Fixed in eb62987. Removed the unused `upperBound` field (and the `twoTo256`/`twoTo512` arguments) from the test's case struct in `TestCertifiedNatThresholdPartialStakeExactCutoff`. Confirmed similarly-named fields in other test structs in this file are actually used and left those alone.

consensus/threshold.go
```diff
@@ -129,99 +167,576 @@ func CertifiedNatThresholdWithMode(
...
+// maxThresholdEscalationBits caps how far thresholdFromBoundedProbability's
+// caller grows the working precision (targetBits) before giving up and
+// returning a best-effort result. This is far beyond what any plausible
+// protocol parameter (activeSlotCoeff, pool/total stake) would require to
+// resolve for a genuinely irrational (1-f)^sigma; an *exact rational*
+// cutoff is always caught first by exactOneMinusFPowerSigmaThreshold (a
+// complete decision procedure with no cap on sigma's reduced denominator),
+// so reaching this cap without resolving is expected to be unreachable in
+// practice.
+const maxThresholdEscalationBits = 1 << 20 // ~1,048,576 bits
```
chrisguiney: **Blocking (perf/liveness):** this cap allows roughly an hour of uninterruptible CPU in a single call.

I measured one `thresholdFromBoundedProbability` evaluation on this branch:

```
targetBits=576      83µs
targetBits=1152     247µs
targetBits=2304     835µs
targetBits=4608     4.1ms
targetBits=9216     20.5ms
targetBits=18432    109ms
targetBits=36864    607ms
targetBits=73728    3.36s
```

That's ~5.5x per doubling. Extrapolating to the cap, the escalation loop would spend ~9 min at 589,824 bits and ~50 min at 1,179,648 bits before returning — with no context and no cancellation path, in an exported library API.

The doc comment argues this is unreachable, and I think that's probably right: I confirmed `exactIntegerNthRoot` is a complete decision procedure, and `(p/q)^(n/m)` is rational iff `p` and `q` are both perfect m-th powers, so exact-rational cutoffs really are always caught first. But the residual risk rests on the error bound declared in `oneMinusFPowerSigmaBounds`, which is asserted in a comment rather than proven — and if that bound is ever off, this is the branch that pays for it.

The good news is that lowering the cap costs nothing. I instrumented the loop and ran 5,600 varied inputs (14 `f` values x poolStake 1..40 x 5 totalStake values x both modes): **the loop never escalated even once** — every single call resolved at the first 576-bit level.

Suggest `1 << 14` (16,384 bits, still 32x the largest upper bound), which bounds the worst case at roughly 0.1s. If reaching the cap should ever be visible, a counter or log line there would be worth having too.

consensus/threshold_boundary_test.go
```diff
@@ -0,0 +1,506 @@
...
+				// test above for the tight bound); here we only need it
+				// exact enough that the boundary values below don't
+				// straddle a difference between the two.
+				diff := new(big.Int).Sub(threshold, cutoff)
+				diff.Abs(diff)
+				require.True(t, diff.BitLen() <= 4,
+					"production threshold too far from reference cutoff "+
+						"to test boundary meaningfully")
+
+				below := new(big.Int).Sub(cutoff, big.NewInt(1))
```
chrisguiney, line 504: **Blocking (test correctness):** `below`, `at` and `above` are derived from `cutoff` (the reference value) but `checkBelow` compares them against `threshold` (the production value).

CodeRabbit raised this on `16cd9ec` and I think it's still open — worth resolving because the tolerance and the assertions disagree about what the test requires:

- `require.True(t, checkBelow(below))` needs `cutoff - 1 < threshold`, i.e. `threshold >= cutoff`
- `require.False(t, checkBelow(at))` needs `cutoff >= threshold`

Together those force `threshold == cutoff` exactly, so the `diff.BitLen() <= 4` guard at line 458 (and the comment above it saying the tolerance keeps the boundary values from straddling) doesn't hold — *any* nonzero difference straddles. As written the test advertises a 15-integer tolerance while silently requiring exactness, and a one-integer drift on either side would fail at line 497 with a message pointing at the wrong thing.

The happy part: exactness is what actually holds. I checked `refThreshold` against my own exact integer-only oracle and it agreed in every case I tried, as did the production value — so the two are genuinely equal here, not equal by luck.

Either of these fixes it:
- tighten `diff` to `require.Zero(t, diff.Sign())` and keep deriving from `cutoff`, or
- derive `below`/`at`/`above` from `threshold` so the assertions test the strict `<` comparison on its own terms, leaving `TestCertifiedNatThresholdDifferentialAgainstIndependentReference` to cover accuracy.

I'd lean toward the first — it documents the stronger property you've actually achieved.

consensus/threshold_boundary_test.go
```diff
@@ -0,0 +1,506 @@
...
+						)
+						require.NoError(t, err)
+						return ok
+					}
+					return VRFOutputToInt(toBytes(v)).Cmp(threshold) < 0
+				}
+
+				require.True(t, checkBelow(below),
+					"value immediately below cutoff must be eligible")
+				require.False(t, checkBelow(at),
```
chrisguiney: **Non-blocking, but worth recording:** this assertion encodes gouroboros' own cutoff semantics rather than checking them against upstream, and #1941 specifically asks to prove identical eligibility *at* the cutoff.

`checkLeaderNatValue` in cardano-ledger (`libs/cardano-protocol/src/Cardano/Protocol/TPraos/BlockHeader.hs:380`) is `BELOW <=> recip_q < exp(x)`, which unfolds to `certNat < certNatMax * (1 - (1-f)^sigma)` over the reals. gouroboros compares `v < floor(X)`. When `X` isn't an integer, upstream admits `v = floor(X)` and we reject it — a genuine off-by-one, though at probability ~2^-256 per header and entirely pre-existing, so definitely not something to fix in this PR.

Since `threshold == cutoff` by construction here, the `at` and `above` assertions are tautologies (`threshold < threshold` is trivially false), so they can't catch this either way.

I'd suggest just a sentence in the PR body or a comment here noting the residual +/-1 at the exact cutoff as known and accepted — and that upstream's own `Fixed E34` (`fpPrecision = 10^34`, ~113 bits) makes bit-exact boundary agreement with cardano-node unattainable by construction anyway. That closes out the issue's "at the cutoff" criterion honestly rather than leaving it looking unaddressed.

consensus/threshold.go
```diff
@@ -129,99 +167,576 @@ func CertifiedNatThresholdWithMode(
...
+	// 1-f > 0.
+	if fCmpOne == 0 {
+		return new(big.Int).Set(upperBound), nil
+	}
+	// sigma == 1 (full stake) means (1-f)^sigma == 1-f exactly, so the
+	// threshold can be computed exactly via rational arithmetic instead
+	// of the ln/exp pipeline, avoiding any residual floating-point error
+	// that could otherwise push an exact-rational cutoff across an
+	// integer boundary (see the full-stake, f=1/2 differential test).
+	if poolStake == totalStake {
```
chrisguiney: **Non-blocking (duplication):** this sigma==1 branch is fully subsumed by `exactOneMinusFPowerSigmaThreshold` below.

With `poolStake == totalStake` the reduced sigma is `1/1`, so `exactIntegerNthRoot(n, 1)` hits its `k.Cmp(bigIntOne) == 0` fast return and gives back `n` unchanged — `(1-f)^1`, and then the same `upperBound * num / denom` flooring. I checked this directly and the two paths produce identical values for f in {1/20, 1/3, 1/2, 7/11, 999/1000, 1/7} against both `twoTo256` and `twoTo512`.

Dropping it removes ~10 lines of duplicated integer logic and one more place that has to stay in sync if the flooring convention ever changes. The `oneMinusF` comment just below would need a small tweak since it currently reads as following on from this block.

Entirely your call — if you'd rather keep it as an explicit cheap short-circuit, a one-line note that it's a redundant fast path (not the only handler for sigma==1) would save the next reader the same trip.

consensus/threshold.go
```diff
@@ -129,99 +167,576 @@ func CertifiedNatThresholdWithMode(
...
+
+	// Initial over-estimate: x0 ~= 2^(ceil(bitlen(n)/k)+1).
+	guessBits := n.BitLen()/int(kInt) + 1
+	x := new(big.Int).Lsh(big.NewInt(1), uint(guessBits))
+
+	// Newton's method for the integer k-th root converges monotonically
+	// downward from an over-estimate to floor(n^(1/k)).
+	for {
+		xPow := new(big.Int).Exp(x, kMinus1, nil)
+		if xPow.Sign() == 0 {
```
chrisguiney: **Non-blocking (dead code):** this branch is unreachable, and would spin forever if it ever were reached.

`x` stays `>= 1` throughout the Newton loop: it starts at `2^guessBits >= 2`, only ever decreases (the `next.Cmp(x) >= 0` guard breaks otherwise), and `next = ((k-1)*x + n/x^(k-1)) / k >= 1` whenever `x >= 1`, since `n >= 2` at this point. So `xPow = x^(k-1) >= 1` always and `xPow.Sign() == 0` can't happen.

If it somehow did, `x` would be 0 and `x.Lsh(x, 1)` leaves it at 0 — an infinite loop rather than a recovery.

Suggest deleting it, or turning it into a defensive `break`/error return so a future refactor that breaks the `x >= 1` invariant fails loudly instead of hanging. (For what it's worth I fuzzed `exactIntegerNthRoot` over n in {2..2^20} x k in {1..24} and found no missed roots and no false positives — the function itself looks solid.)

consensus/threshold.go
```diff
@@ -129,99 +167,576 @@ func CertifiedNatThresholdWithMode(
...
+
+// computeLn2AtTarget computes ln(2) = -ln(0.5) using lnNormalized (0.5 is
+// already in the required [0.5, 1) range), to targetBits bits of series
+// accuracy, for lnPositiveFloatAtTarget's callers. It is deliberately not
+// cached: it is only invoked a handful of times per
+// CertifiedNatThresholdWithMode call (at most once per precision-
+// escalation retry), so the cost of recomputing it at whatever targetBits
+// the caller currently needs is negligible next to caching it at a single
+// fixed precision that later callers might need to exceed anyway.
+func computeLn2AtTarget(prec, targetBits uint) *big.Float {
```
chrisguiney, line 700: **Non-blocking (perf):** worth revisiting the "deliberately not cached" reasoning now that the escalation loop turns out to be dead in practice.

The doc argues caching at a fixed precision isn't worth it because later callers might exceed it. But `lnPositiveFloatAtTarget` short-circuits on `exp == 0`, so this is never reached for f <= 1/2 (mainnet, and the 0.4 devnets) — and for f > 1/2 it runs ~218 artanh terms at 704 bits on *every* call, which is a meaningful slice of the total. Meanwhile I instrumented the escalation loop over 5,600 varied inputs and it never escalated once, so in practice `targetBits` is always exactly `seriesTargetBits`.

A `sync.Once`-cached value at the baseline with a recompute fallback for anything above it would get both properties, and the fallback path would essentially never execute. Purely optional, but it'd claw back a chunk of the overhead noted below.

consensus/threshold.go
```diff
@@ -129,99 +167,576 @@ func CertifiedNatThresholdWithMode(
...
+
+	if x.Sign() == 0 {
+		return new(big.Float).SetPrec(prec).SetInt64(1)
+	}
+
+	mant := new(big.Float).SetPrec(prec)
+	e := x.MantExp(mant) // x = mant * 2^e, 0.5 <= |mant| < 1
+
+	k := e + expReductionBits
+	if k < 0 {
```
chrisguiney: Nit: `modernize` flags this one — `k = max(k, 0)` reads a little cleaner.

gouroboros CI doesn't run `modernize`, so purely take-it-or-leave-it. (The only other hit in the diff is a range-over-int in `threshold_boundary_test.go:112`.)

consensus/threshold_test.go
```diff
@@ -1016,7 +1114,376 @@ func TestLnOneMinusConvergence(t *testing.T) {
...
+			// For TPraos, IsVRFOutputBelowThresholdWithMode compares the
+			// raw bytes directly, so the exact boundary values can be
+			// driven through the public API. For CPraos, the public API
+			// hashes the input first (BLAKE2b-256 with "L" prefix), so
+			// instead replicate the same comparison the implementation
+			// performs internally after hashing
+			// (VRFOutputToInt(leaderValue).Cmp(threshold) < 0), using the
+			// exact boundary values as if they were already the
+			// post-hash leader value.
+			checkBelow := func(v *big.Int) bool {
```
chrisguiney: Nit (duplication): this `checkBelow` closure is now repeated verbatim in three places — here, `threshold_test.go:1465`, and `threshold_boundary_test.go:484` — comments included. cubic flagged this and I think it's still open.

A single file-level helper taking `(t, bits, mode, threshold, value)` would keep them in sync. It matters a bit more than usual here because the strict `<` vs `<=` question is exactly the thing these tests exist to pin down — if that convention ever changes, three copies is three chances for one to drift.

While you're in the neighbourhood: a fair chunk of this file's diff (~250 lines) is pure `golines` reformatting of tests this PR doesn't otherwise touch. Not wrong at all given the repo has a `golines` target, but it does bury the substantive additions. No need to unwind it now — just a note for next time that splitting the reformat into its own commit makes the review much easier to follow.

chrisguiney (changes requested):
Thanks for sticking with this one — the extra rounds since the first review clearly paid off. I re-derived the math independently and the core of this is correct; the two earlier blocking findings are genuinely resolved. I have two items I'd like to see before approving, plus some optional cleanups. Details inline.

## What I verified independently

I built an exact, integer-only oracle for `floor(UB * (1 - (1-f)^(p/q)))` with no floating point anywhere — it reduces to "smallest `j` with `j^q * B >= UB^q * A`" and binary-searches `j`, so it's exact by construction.

- **This PR matches the oracle bit-for-bit** across 392 combinations (14 `f` values x 14 sigma values x both modes), including f in {0.9, 0.95, 0.99, 0.999, 0.9999} and sigma denominators up to 31. Nice work — that's a stronger result than the `diff.BitLen() <= 4` tolerance in the differential test claims.
- **The old code was worse than the PR description says.** Relative error vs. exact:

| f | old relative error |
|---|---|
| 1/20 (mainnet) | below the 2^-256 floor — exact |
| 1/2 | 2^-107 |
| 9/10 | **2^-20** |
| 19/20 | **2^-13** |
| 99/100 | **2^-6** |
| 999/1000 | **2^-2** (25% off) |

For calibration: cardano-ledger's own check runs in `Fixed E34` (`fpPrecision = 10^34`, ~113 bits) via `taylorExpCmp`, so upstream itself is only ~2^-113 precise. The old code sat at or below that floor for f <= 0.5 but was far outside it for f >= 0.9 — so this is a real eligibility-divergence bug on high-f devnets, not a theoretical one. Good catch, and worth putting those numbers in the PR body.

Mainnet CPraos thresholds come out byte-identical to before; TPraos and higher-f values shift by ~2^-497 / ~2^-233 relative, so no practical consensus impact from the change itself.

## Other checks

`go build`, `go vet`, `gofmt`, `golangci-lint` (0 issues) and `nilaway` are all clean, and `go test -race ./consensus/...` passes. No concurrency concerns — the package-level `big.Int`/`big.Rat` values stay read-only, the `activeSlotCoeff.Num()`/`.Denom()` aliases are never mutated, and every function here is pure. All three in-repo callers plus both dingo callers (`ledger/verify_header.go:932`, `ledger/leader/schedule.go:296`) already propagate the new error, so the domain-guard change is safe downstream.

One unrelated note: `protocol/chainsync.TestClientStartStopStart` fails under `-race` on this branch, but it also fails at this PR's merge-base (`81080f9`) and passes on current `main` — it was fixed by the ouroboros-mock v0.15.0 -> v0.15.1 bump. Not caused by this PR; a rebase onto `main` clears it locally. PR CI is green either way.

## Requested changes

1. `consensus/threshold.go:269` — bound `maxThresholdEscalationBits` (measured ~1 hour worst case).
2. `consensus/threshold_boundary_test.go:462` — derive the boundary values from `threshold`, not `cutoff`.

The rest of the inline comments are non-blocking suggestions. Happy to re-review as soon as those two land.

cryptodj413:
Thanks for the detailed re-review and the independent oracle verification — both blocking items are fixed in 323ebe7 (pushed to this branch):

1. **Escalation cap (`consensus/threshold.go`)**: lowered `maxThresholdEscalationBits` from `1<<20` to `1<<14` (16,384 bits, 32x the largest baseline `seriesTargetBits`). Per your measured ~5.5x cost-per-doubling, this bounds the worst case (if the cap is ever actually reached) to roughly 0.1s instead of the multi-minute/near-hour hang the old cap permitted in this exported, non-cancellable API. Also added a cheap atomic counter, `thresholdEscalationCapReachedCount`, incremented only if the cap is ever reached without resolving, so a nonzero value in practice would flag either a pathological input or a bug in the declared error bound.

2. **Boundary test (`consensus/threshold_boundary_test.go`)**: fixed the below/at/above values to actually correspond to what's being compared — they're derived from `cutoff` but `checkBelowThreshold` compares against `threshold`, so the tolerance assertion between the two now requires exact equality (`diff.Sign() == 0`) rather than the old `diff.BitLen() <= 4` approximation. Verified all 16 cases (8 f/sigma combinations x 2 modes) pass at exact equality — confirming `threshold == cutoff` genuinely holds for this table, not just approximately.

Also picked up the non-blocking items from this review round while I was in there:
- Documented the known pre-existing ±1 boundary-precision residual (strict `<` vs upstream's real-number semantics) as an explicit out-of-scope comment on the boundary test.
- Removed the now-redundant `sigma == 1` fast-path branch in `threshold.go` (subsumed by the general exact-rational path — confirmed identical results).
- Deleted the dead `xPow.Sign() == 0` branch in `exactIntegerNthRoot`'s Newton loop.
- Applied the `max()` modernize nit.
- Extracted the 3x-duplicated `checkBelow` closure into one shared helper, `checkBelowThreshold`, now used by `threshold_test.go` (x2) and `threshold_boundary_test.go`.

`make test` and `make lint` are clean on top of this commit (0 lint issues, all consensus tests passing under `-race`). Let me know if you'd like the remaining CodeRabbit/cubic nitpicks (epsilon construction, comment wording, `lnPositiveFloat` domain guard) addressed in a follow-up as well.

cryptodj413:
Thanks for the detailed follow-up review. Addressed in [`91ad5d9`](https://github.com/blinklabs-io/gouroboros/commit/91ad5d9051e04bbd1668c2c3dc5a3ea0605b76c5):

**Finding 1 (P1, consensus/threshold.go)** — Fixed. Extracted the escalation loop into a new `escalateThreshold` function. It now returns an explicit `(nil, error)` when `maxThresholdEscalationBits` is reached without resolving, instead of silently falling back to the unproven lower bound. `CertifiedNatThreshold`'s existing error→zero fallback covers this automatically, and both real callers (`consensus/leader.go`, `consensus/validate.go`) already propagate the error from `CertifiedNatThresholdWithMode` correctly — verified via grep, no caller changes were needed. Added `TestEscalateThresholdCapReachedReturnsError`, which forces a tiny cap to exercise the error path directly.

**Finding 3 (P3, consensus/threshold.go)** — Fixed. Removed `thresholdEscalationCapReachedCount` (and the now-unused `sync/atomic` import) — it's redundant now that Finding 1 replaces the silent-counter approach with a typed error, which is a strictly better signal for both tests and production monitoring.

**Finding 2 (P2, consensus/threshold_boundary_test.go)** — Investigated, left as-is. Measured the actual residual margin for the current test cases concretely: ~2^-1477 absolute error against the largest upper bound, i.e. ~1450 bits of headroom before `floor()` could flip. The precision-drift regime you're flagging (refPrec rounding pushing an exact-rational cutoff across an integer boundary) is real per the file's own NOTE, but it only manifests when pushing `refPrec` much higher for `f` values very close to 1 — a regime these test cases deliberately don't exercise. Documented this margin measurement and reasoning in-file, with an explicit trigger to re-check if `refPrec` or the test cases ever change.

`make test` (incl. `go test ./consensus/... -race -v`) and `make lint` are green on the pushed commit.

chrisguiney (approved):

PR #1968, opened 2026-08-10
fix(conway): enforce complete GOV proposal and voting validation

### LINKED ISSUE #1938: conway: enforce complete GOV proposal and voting validation
The Conway UtxoValidationRules registry covers proposal shape/network checks, bootstrap restrictions, certificates, and the PV11 constitutional-committee voting restriction. It does not contain the full Conway GOV validation set.

Missing or unaccounted-for checks include bad hard-fork actions, action well-formedness, proposal refund address, non-registered accounts, proposal deposit, return-address network, action-specific policy and consistency checks, ancestry, unknown voters, unknown governance action IDs, bootstrap votes, expired actions, and voter authorization.

UtxoValidateCCVotingRestrictions explicitly skips unknown action IDs with the comment that another validation rule handles them, but no corresponding rule is registered in the current Conway path.

Impact: malformed or unauthorized governance proposals/votes may pass transaction validation, and the implementation does not match the Conway GOV transition rules.

References:
- https://github.com/blinklabs-io/gouroboros/blob/main/ledger/conway/rules.go
- https://cardano-scaling.github.io/cardano-blueprint/ledger/block-validation.html
- https://github.com/IntersectMBO/cardano-ledger

Acceptance criteria:
- Map every Conway GOV transition check to an implementation rule or an explicitly invoked state-transition function.
- Reject unknown voters and governance action IDs.
- Validate proposal deposits, refund/return addresses, ancestry, expiry, action-specific constraints, and voter authorization.
- Add negative conformance vectors for each missing rule across PV9, PV10, and PV11 where behavior differs.

cryptodj413:
## Summary

Refs #1938

Implements the missing Conway GOV validation rules identified in #1938:

- Gov-action well-formedness (`MalformedProposal` + `ConflictingCommitteeUpdate`)
- Hard-fork-can-follow (`ProposalCantFollow` base case)
- Proposal ancestry (`InvalidPrevGovActionId`)
- Proposal deposit correctness (`ProposalDepositIncorrect`)
- Proposal / treasury-withdrawal return-account existence
  (`ProposalReturnAccountDoesNotExist` / `TreasuryWithdrawalReturnAccountsDoNotExist`),
  correctly skipped during bootstrap per the real cardano-ledger source
  `conwayGovTransition` in `Cardano.Ledger.Conway.Rules.Gov`
- Unknown voters (`VotersDoNotExist`)
- Unknown governance action IDs (`GovActionsDoNotExist`) — the specific gap
  that `UtxoValidateCCVotingRestrictions`'s existing code comment referenced
  but which had no implementation
- Voting-on-expired-action (`VotingOnExpiredGovAction`)
- Bootstrap voting restrictions (`DisallowedVotesDuringBootstrap`)
- Stake-pool voting restrictions (`DisallowedVoters`, SPO side)

This PR deliberately does not close #1938 (`Refs #1938`, not `Closes`) —
several items from that issue remain out of scope, documented via `NOTE`
comments in the code rather than silently dropped:

- Guardrails-script-hash matching (constitution policy has no script-hash
  field modeled)
- Current-epoch accessor (no epoch accessor on `LedgerState`), which affects
  committee-member `ExpiryEpoch` checks and the expiration-epoch-too-small
  check
- Full ancestor protocol-version chain comparison
- Per-purpose current-root tracking for ancestry
- SPO security-group parameter classification for `ParameterChange` votes
- Unelected-committee-voter flag
- Dijkstra-era parity (this PR is Conway-only per the issue's scope)

## Review history

This PR went through two `code-reviewer` passes:

1. First pass flagged one blocking concern: the bootstrap exemption in
   return-account validation looked uncited/suspicious. This was
   investigated against the actual upstream `cardano-ledger` Haskell source
   and confirmed correct per spec; the exemption is now cited in a `NOTE`
   comment rather than removed.
2. Second pass approved with 0 blocking items.

## Known follow-up (not blocking, pre-existing, not introduced by this PR)

`UtxoValidateCCVotingRestrictions` lacks a nil-voter guard that all the new
sibling rules added in this PR have, which could theoretically panic on a
malformed nil `*Voter` map key. Worth a separate follow-up PR/issue.

## Test coverage

New `ledger/conway/rules_gov_transition_test.go` with ~100+ subtests across
PV9/PV10 covering pass and fail paths for each new rule.

## Test plan

- `make test` — green
- `make lint` — green
- `go test -race ./...` — green
- `go test ./internal/test/conformance/...` — 315/315 ledger rule vectors pass

---
## Summary by cubic
Enforces the full set of Conway governance proposal and voting validations to block invalid proposals and votes, and avoids false rejections when expiry is unmodeled. Adds stricter return-account and voter type checks and stabilizes error messages. Refs #1938.

- **New Features**
  - Gov-action well-formedness: policy hash length, constitution hash length, conflicting committee updates.
  - Hard-fork-can-follow check (base case).
  - Proposal ancestry: referenced action exists and matches purpose.
  - Deposits and return accounts: deposit equals GovActionDeposit; reward/treasury return accounts must exist (skipped in PV9).
  - Voting checks: unknown voters, unknown governance action IDs, and votes on expired actions are rejected.
  - Voting restrictions: PV9 bootstrap limits; SPOs cannot vote on NewConstitution or TreasuryWithdrawal.
  - Tests: 100+ subtests across PV9/PV10.

- **Bug Fixes**
  - Accept only reward_account address types for return accounts; reject base/pointer/enterprise.
  - Reject out-of-range voter types in unknown-voter checks.
  - Sort error details for deterministic messages (unknown action IDs, conflicting committee creds, bad treasury accounts).
  - Align BootstrapVotingRestrictionError.VoterId to `common.Blake2b224`.
  - Treat unset `GovActionState.ExpirySlot` as “expiry not modeled” to avoid rejecting valid votes.

<sup>Written for commit ccc0a7997527295e346cd800bc69b820df0b0286. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/1968?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive governance validation for proposals, voting, deposits, return accounts, hard forks, and governance-action relationships.
  * Added checks for unknown voters and actions, expired votes, malformed actions, and restricted voting scenarios.
* **Bug Fixes**
  * Improved validation feedback with detailed errors for invalid governance transactions.
* **Tests**
  * Added extensive coverage for valid and invalid governance transitions across protocol versions.


cryptodj413:
Pushed a follow-up commit (5f87631) addressing review feedback from both the CodeRabbit and cubic-dev-ai automated reviews.

**cubic-dev-ai** — both P1 findings fixed:
- `UtxoValidateUnknownVoters`: added a default/reject case for out-of-range voter types instead of silently passing them through.
- `UtxoValidateProposalReturnAccounts`: now restricted to actual reward-account address types (`AddressTypeNoneKey`/`AddressTypeNoneScript`) rather than accepting any address with a staking payload.

Left as-is: the P3 performance finding (linear scan for CC hot-key lookup) — committee size is small in practice, not worth the added complexity right now.

**CodeRabbit** — fixed:
- Aligned `BootstrapVotingRestrictionError.VoterId` to `common.Blake2b224` for consistency with sibling error types.
- Sorted three previously-nondeterministic error-collection slices before constructing errors, for reproducible error messages.
- Added a pin-test for the documented SPO/ParameterChange NOTE.

Assessed and not applied:
- Suggestion to add a PV11 duplicate test case for `TestUtxoValidateBootstrapVotingRestrictions`: `isInConwayBootstrapPhase` only distinguishes PV9 from everything else (PV10/PV11 behave identically), confirmed via code read, so a PV11 case would add no real additional coverage.
- The "update conformance snapshot table" comment on rules.go:39-44: that location is the `UtxoValidationRules` slice declaration, unrelated to any conformance snapshot table — looks like a false positive/misattached comment.

`make test` and `make lint` both green on the updated branch.

ledger/conway/rules.go
```diff
@@ -2633,6 +3032,315 @@ func PoolValidateVrfKeyUniqueness(
...
+			if actionId == nil {
+				return UnknownGovActionIdError{
+					ActionIds: []common.GovActionId{{}},
+				}
+			}
+			actionState, err := ls.GovActionById(*actionId)
+			if err != nil || actionState == nil {
+				continue
+			}
+			if slot > actionState.ExpirySlot {
```
chrisguiney, line 3214: **Blocking:** this rejects every governance vote when run against dingo's production ledger.

The chain:

1. Before this PR, `common.GovActionState.ExpirySlot` was declared but read by nothing in gouroboros — `git grep ExpirySlot` at the merge base returns only the field declaration in `ledger/common/state.go`. This rule is its first consumer.
2. dingo's production `LedgerView.GovActionById` (`ledger/view.go:730-748`) builds the state with only `ActionId` and `ActionType`:
   ```go
   return &lcommon.GovActionState{
       ActionId:   id,
       ActionType: lcommon.GovActionType(proposal.ActionType),
   }, nil
   ```
   `ExpirySlot` is never assigned, so it stays zero.
3. dingo builds its Conway rule list straight from `conway.UtxoValidationRules` (`ledger/eras/conway.go:283`), so this runs in production block and tx validation.
4. `slot > 0` is true at every slot except 0.

I ran it against a `LedgerState` shaped exactly like dingo's — `ActionId` + `ActionType`, `ExpirySlot` unset — with an ordinary DRep vote on a live `Info` action:

```
slot 1:         REJECTED: vote cast on expired governance action ... (expired at slot 0, current slot 1)
slot 1000:      REJECTED: ... (expired at slot 0, current slot 1000)
slot 150000000: REJECTED: ... (expired at slot 0, current slot 150000000)
```

The test suite can't catch this today because the only fixture sets `ExpirySlot: 1000`. dingo's *conformance* provider does populate it (`ExpiresAfter * 432000`, itself commented "Approximate"), which is why the conformance run stays green while production would not be.

Suggested fix, using a convention this PR already established — treat an unpopulated `ExpirySlot` as "not modeled" and skip, the same way this rule already skips on `err != nil || actionState == nil`, and the same way `UtxoValidateUnknownVoters` treats `len(members) == 0` as unmodeled committee state:

```go
// ExpirySlot is optional in the LedgerState contract: a state
// provider that does not model gov-action expiry leaves it zero.
// Treat that as "expiry not modeled" rather than "expired at slot
// 0", which would reject every vote at any slot > 0.
if actionState.ExpirySlot == 0 {
    continue
}
if slot > actionState.ExpirySlot {
```

A fixture with `ExpirySlot` unset that asserts acceptance would pin it.

Separately (and not this PR's job): populating `ExpirySlot` in dingo is worth doing — the data is already there as `GovernanceProposal.ExpiresEpoch` — but I'd still want the library rule to not hard-reject on a field its callers were never required to fill.

ledger/conway/rules.go
```diff
@@ -2633,6 +3032,315 @@ func PoolValidateVrfKeyUniqueness(
...
+			if actionId == nil {
+				// A nil action id cannot be looked up in the ledger
+				// state and is therefore treated the same as a
+				// reference to a nonexistent action, matching
+				// UtxoValidateCCVotingRestrictions's convention of
+				// erroring rather than silently skipping.
+				unknown = append(unknown, common.GovActionId{})
+				continue
+			}
+			if !ls.GovActionExists(*actionId) {
```
chrisguiney, line 3066: Non-blocking, but worth considering before this lands, since it's cheaper to restructure now than after five rules exist.

There are now five separate passes over `tx.VotingProcedures()`, each doing its own per-vote ledger lookup: `GovActionExists` here at line 3058, then `GovActionById` at 3194, 3254, 3315, and 3386. In dingo each of those resolves to a SQL `SELECT` on `governance_proposal`, so a transaction with V voters × A actions now costs roughly 5×V×A queries where a single resolving pass would need A.

The duplication has already produced a behavioural divergence: this rule hard-rejects when an action isn't found, while the other four `continue` past the same condition. That's defensible individually, but it means the five rules disagree about what a missing action means, and nothing makes that deliberate.

One pass that resolves each distinct action id once and hands the resolved `*GovActionState` to the individual checks would fix the query count, collapse five copies of the same loop skeleton and nil-action-id handling into one, and make the missing-action policy a single explicit decision. Entirely reasonable as a follow-up if you'd rather keep this PR focused.

ledger/conway/rules.go
```diff
@@ -36,7 +38,12 @@ import (
 var UtxoValidationRules = []common.UtxoValidationRuleFunc{
 	UtxoValidateMetadata,
 	UtxoValidateProposalProcedures,
+	UtxoValidateGovActionWellFormedness,
+	UtxoValidateHardForkCanFollow,
+	UtxoValidateProposalAncestry,
+	UtxoValidateProposalDeposit,
```
chrisguiney, line 44: Heads-up for whoever does the dingo bump — not a defect in this PR, just cross-repo coordination that's easy to miss.

dingo pins skip-by-index constants against this exact slice: `conwayUtxoValidateFeeTooSmallRuleIndex = 19` and `conwayUtxoValidatePlutusScriptsRuleIndex = 38` (`ledger/eras/validation.go:113,115`). This PR inserts five rules before index 19, so both shift by five.

The good news is it fails loudly rather than silently: `validateUtxoValidationSkipIndex` compares function pointers and panics at package init if an index no longer resolves to the expected function. So there's no risk of dingo quietly skipping the wrong rule — but those two constants do need updating in the same bump, or dingo won't start.

ledger/conway/rules.go
```diff
@@ -376,6 +393,378 @@ func UtxoValidateProposalNetworkIds(
...
+
+// UtxoValidateProposalAncestry checks that a governance action's optional
+// PrevGovActionId, if present, references an existing governance action
+// belonging to the same purpose chain (ConwayGovPredFailure.InvalidPrevGovActionId).
+//
+// NOTE: the ledger spec additionally verifies that the referenced ancestor
+// is the *current* root of its purpose chain (i.e., no other pending action
+// of that purpose exists that isn't this proposal's ancestor). GovState in
+// this codebase does not expose per-purpose root tracking, so this rule is
+// limited to existence and purpose-type matching.
```
chrisguiney, line 624: Small addition to this `NOTE`, non-blocking.

Because the rule rejects when the ancestor isn't found, its safety depends on the state provider keeping *enacted* actions queryable — in `cardano-ledger` the current root of a purpose chain is normally an already-enacted action, and enacted actions are removed from the pending proposals map.

I checked dingo and it's fine today: `GetGovernanceProposal` filters only on `deleted_slot IS NULL`, so enacted rows stay visible and valid chained proposals aren't rejected. But that's an implicit dependency on a `LedgerState` implementation detail rather than something the interface promises, and a backend that pruned enacted proposals would start rejecting valid proposals. Worth one line here so the next person doesn't have to rediscover it.

chrisguiney (changes requested):
Thanks for taking this on — it's a big, careful piece of work, and the parts I could check against `cardano-ledger` hold up well. `protocolVersionCanFollow` faithfully reproduces `pvCanFollow`, the SPO restrictions match `votingStakePoolThreshold` (`TreasuryWithdrawals`/`NewConstitution` → `SNothing`), and I confirmed the bootstrap return-account exemption really is wrapped in `unless (hardforkConwayBootstrapPhase ...)` in `conwayGovTransition`, exactly as your `NOTE` says. The `NOTE` comments marking deliberately-deferred checks are genuinely useful, and scoping this as `Refs #1938` rather than `Closes` is the right call.

I'm requesting changes for one issue: `UtxoValidateVotingOnExpiredGovAction` depends on a `LedgerState` field that no production implementation currently populates, and the result is that every governance vote gets rejected. Details inline. It's a small fix and it follows a convention this PR already uses elsewhere.

**What I verified as clean:** `go build ./...`, `go test -race ./...` (all packages pass), `gofmt`, `golangci-lint` (0 issues), and all 10 CI checks green. No concurrency issues — every new rule is a pure function over the transaction plus read-only ledger queries, with no goroutines or shared state. No panic vectors either: `Voter.Hash` is `[28]byte`, so the `common.Blake2b224(voter.Hash)` conversions and every `[:8]` slice in the new error types are safe.

The three non-blocking notes inline are all things I'd be happy to see in a follow-up rather than here.

ledger/conway/rules.go
cryptodj413: Confirmed and fixed in ccc0a79 (pushed to this branch).

Root cause matches your analysis exactly: `GovActionState.ExpirySlot` had no prior consumer, so LedgerState implementations like dingo that don't populate it leave it at zero, and `slot > 0` was always true, rejecting every vote.

Fix: when `actionState.ExpirySlot == 0`, treat expiry as unmodeled by this LedgerState implementation and skip the check, rather than treating it as expired-at-slot-0. This matches the PR's existing convention of treating other unmodeled LedgerState surfaces (e.g. empty committee-members list) as permissive.

Added a test (`unset ExpirySlot is treated as not modeled` in `ledger/conway/rules_gov_transition_test.go`) alongside the existing `ExpirySlot:1000` pass/fail expiry-enforcement tests, which are left intact.

This was also independently verified by a code-reviewer pass that reverted the guard locally and confirmed the new test reproduces the exact symptom you reported before the fix was applied.

ledger/conway/rules.go
cryptodj413: Addressed in ccc0a79 — appended a line to the `UtxoValidateProposalAncestry` NOTE documenting the implicit LedgerState dependency you flagged (the rule's safety depends on enacted governance actions remaining queryable via `GovActionById` rather than being pruned once enacted). Doc-only, non-blocking, as noted.

ledger/conway/rules.go
cryptodj413: Agreed this is worth doing, but deferring it as a follow-up rather than restructuring in this PR — per your own framing ("non-blocking, but worth considering"), consolidating the five `VotingProcedures()` passes into a single resolving pass is a good candidate for a dedicated follow-up PR rather than expanding the diff here. Not fixed in this PR.

ledger/conway/rules.go
cryptodj413: Acknowledged — thanks for flagging this cross-repo coordination point. Not something to fix in this PR itself; leaving it as a heads-up for the dingo bump, as you suggested, so `conwayUtxoValidateFeeTooSmallRuleIndex`/`conwayUtxoValidatePlutusScriptsRuleIndex` get updated for the +5 shift when this lands there. Not fixed in this PR.

chrisguiney (approved):

PR #1980, opened 2026-08-20
fix(ledger): scope the PlutusV1 restriction to needed scripts

wolf31o2:
## Problem

`UtxoValidateInlineDatumsWithPlutusV1` is wrong in both directions at once. Conway and Dijkstra delegate rule 18 to this babbage implementation, so all three eras share both defects.

**Too broad.** It rejects whenever a PlutusV1 script is merely *reachable* — present in the witness set, or carried as a reference script on any spent or reference input — rather than when a script purpose actually requires one. An ordinary transaction that spends a UTxO carrying an unrelated PlutusV1 reference script is rejected. This halted a node's sync on a real Preview transaction.

**Too narrow.** The inline-datum scan covered only consumed inputs, missing a datum on a reference input or on one of the transaction's own outputs. It also detected datums through a `*BabbageTransactionOutput` type assertion, which misses outputs wrapped by a later era.

## Change

- New `script.NewTxScriptView`: resolves a transaction's inputs and reference inputs once, then reports both the scripts it makes `Available` and the `Needed` subset some purpose requires. The purpose walk covers spending, minting, certifying, rewarding, voting and proposing in canonical redeemer order; eras without the later purposes simply produce none.
- Rule 18 now keys on `Needed`, scans for inline datums across consumed inputs, reference inputs and produced outputs, and reads datum presence through the `TransactionOutput` interface. Datum-*hash* outputs stay correctly out of scope, since `Datum()` reports nil for them.
- A reference script on a produced output is also rejected — equally unrepresentable in the V1 context.
- Input-resolution failures are skipped rather than returned, matching the existing contract that `UtxoValidateBadInputsUtxo` reports them.

## What is deliberately not asserted

No restriction on the mere presence of a reference input. An earlier revision of this branch did assert one, on the reading that PlutusV1's `TxInfo` cannot represent reference inputs at all. It failed the Conway conformance vector `UTXOS/can use reference scripts` (tx 3), which expects such a transaction to succeed. The vector is authoritative and the restriction was removed; a test now pins the accepting behavior so it is not reintroduced.

## Tests

The rule had **no test coverage anywhere in the repository**, which is how the over-broad behavior survived. Six cases added in `ledger/babbage/rules_plutusv1_internal_test.go`.

Against the previous implementation, three fail — the ones that matter:

| Case | Previous behavior |
| --- | --- |
| unneeded PlutusV1 reference script on a spent input | rejected (the false positive) |
| inline datum on a produced output | accepted |
| reference script on a produced output | accepted |
| needed PlutusV1 spending script with inline datum | rejected — correct, unchanged |
| datum-hash output | accepted — correct, unchanged |
| unresolvable input | skipped — correct, unchanged |

Fixtures store outputs as `*BabbageTransactionOutput`, matching how `BabbageTransactionBody.Outputs()` represents them (`&b.TxOutputs[i]`). Using value types would dodge the previous implementation's type assertion and make the comparison measure the fixture rather than the rule.

The tests are in-package because `ouroboros-mock`'s ledger builder imports `ledger/babbage`, so importing it back is a cycle. The local `utxoOnlyLedgerState` embeds `common.LedgerState` as nil on purpose, so any other method this rule starts calling panics loudly rather than silently returning a zero value.

## Validation

| Check | Result |
| --- | --- |
| `go test ./...` | pass |
| `go test ./internal/test/conformance/` | pass |
| `go vet ./...` | clean |
| `golangci-lint v2.13.0 run ./ledger/...` | 1 pre-existing gofumpt finding in `ledger/common/script/purpose.go`, untouched by this change |

## Downstream

Dingo currently carries a local override of rule 18 for exactly this false positive (blinklabs-io/dingo#3240). Once this lands and dingo bumps, that override and its rule-index machinery can be removed.

`UtxoValidateConwayFeaturesWithPlutusV1V2` (rule 19) has the identical presence-not-neededness flaw and now has `TxScriptView` available to fix it the same way. Left out of this change to keep the conformance surface small; it fires on treasury values, proposal and voting procedures, and Conway certificates, so vote-delegation traffic hits it.

---
## Summary by cubic
Scopes the PlutusV1 restriction to scripts the transaction actually needs and scans consumed inputs, reference inputs, and produced outputs via the `TransactionOutput` interface. Previously it rejected on mere script reachability and only scanned consumed inputs; now it rejects only when a V1 script is required and an inline datum is present. Reference scripts and the mere presence of reference inputs are allowed.

- Introduces `script.NewTxScriptView` to resolve inputs once and track Available vs Needed scripts in canonical redeemer order; includes Plutus V4 witness scripts and reference scripts from all resolved inputs.
- `UtxoValidateInlineDatumsWithPlutusV1` now:
  - Fails only if a needed PlutusV1 script exists and an inline datum is on any consumed input, reference input, or produced output. Datum-hash outputs remain allowed.
  - Permits reference scripts on consumed inputs, reference inputs, and produced outputs; skips input-resolution errors (left to `UtxoValidateBadInputsUtxo`).
- Filters voting purposes to voters with script credentials only to avoid key-hash voters being treated as scripts.
- Tests assert the specific error type and cover acceptance of reference scripts, rejection of inline datums on reference inputs, a needed V1 script supplied only via a reference script, and PlutusV2 acceptance with inline datums. Minor `gofumpt` cleanups in `ledger/conway` and `ledger/mary`.

<sup>Written for commit 7f9fce84e569fdc3cfbd267598267b5613c6c937. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/1980?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Plutus V1 transaction validation by rejecting inline datums and reference scripts only when they are unsupported or required by the transaction.
  - Prevented unrelated reference inputs and unused scripts from causing unnecessary validation failures.
  - Preserved correct handling of datum hashes and unresolved inputs.

- **Tests**
  - Added coverage for accepted and rejected Plutus V1 transaction scenarios, including spending scripts, produced outputs, reference inputs, datum hashes, and unresolved inputs.

### DIFF
--- a/ledger/babbage/rules.go
+++ b/ledger/babbage/rules.go
@@ -22,6 +22,7 @@ import (
 	"github.com/blinklabs-io/gouroboros/ledger/allegra"
 	"github.com/blinklabs-io/gouroboros/ledger/alonzo"
 	"github.com/blinklabs-io/gouroboros/ledger/common"
+	"github.com/blinklabs-io/gouroboros/ledger/common/script"
 	"github.com/blinklabs-io/gouroboros/ledger/mary"
 	"github.com/blinklabs-io/gouroboros/ledger/shelley"
 	"github.com/blinklabs-io/plutigo/syn"
@@ -241,92 +242,82 @@ func UtxoValidateCostModelsPresent(
 
 // UtxoValidateInlineDatumsWithPlutusV1 rejects transactions that use inline datums with PlutusV1 scripts
 // Inline datums are a Babbage-era feature and are only supported with PlutusV2+
+// UtxoValidateInlineDatumsWithPlutusV1 rejects a transaction that requires a
+// PlutusV1 script while using a feature the PlutusV1 script context cannot
+// represent. PlutusV1 predates inline datums and reference scripts, so a
+// transaction that needs a V1 script and carries either is invalid.
+//
+// Two properties matter, and getting either wrong diverges from cardano-node:
+//
+//   - Only a *needed* script constrains the transaction. A V1 script that is
+//     merely reachable -- sitting in the witness set, or as a reference script
+//     on some UTxO being spent -- but required by no script purpose does not
+//     make the transaction invalid. Rejecting on availability turns an ordinary
+//     transaction that happens to spend a UTxO carrying an unrelated V1
+//     reference script into a permanent validation failure.
+//   - The disqualifying feature is transaction-wide. An inline datum counts
+//     wherever the V1 context would have to represent it: on a consumed input,
+//     on a reference input, or on one of this transaction's own outputs.
+//
+// Datum presence is read through the TransactionOutput interface rather than a
+// concrete Babbage type assertion, so outputs wrapped by a later era are still
+// inspected. Datum-*hash* outputs are correctly not treated as inline, because
+// Datum() reports nil for them.
 func UtxoValidateInlineDatumsWithPlutusV1(
 	tx common.Transaction,
 	slot uint64,
 	ls common.LedgerState,
 	pp common.ProtocolParameters,
 ) error {
-	// Check if transaction spends any UTxOs with inline datums
-	hasInlineDatums := false
-	for _, input := range tx.Inputs() {
-		utxo, err := ls.UtxoById(input)
-		if err != nil {
-			// Input not found in ledger state, skip
-			continue
+	view, err := script.NewTxScriptView(tx, ls)
+	if err != nil {
+		if errors.Is(err, common.ErrInputResolution) ||
+			errors.Is(err, common.ErrReferenceInputResolution) {
+			// UtxoValidateBadInputsUtxo reports an unresolvable input with the
+			// right error; reporting it here as well would make this rule a
+			// second, competing source of input-resolution failures.
+			return nil
 		}
-		babbageOutput, ok := utxo.Output.(*BabbageTransactionOutput)
-		if !ok {
-			continue
-		}
-		if babbageOutput.DatumOption != nil &&
-			babbageOutput.DatumOption.data != nil {
-			hasInlineDatums = true
-			break
-		}
+		return err
 	}
-
-	if !hasInlineDatums {
+	needsV1 := view.NeedsAny(func(s common.Script) bool {
+		_, ok := s.(common.PlutusV1Script)
+		return ok
+	})
+	if !needsV1 {
 		return nil
 	}
-
-	// Check if transaction uses PlutusV1 scripts in witnesses
-	witnesses := tx.Witnesses()
-	if witnesses != nil {
-		v1Scripts := witnesses.PlutusV1Scripts()
-		if len(v1Scripts) > 0 {
+	for _, utxo := range view.AllResolvedInputs() {
+		if utxo.Output != nil && utxo.Output.Datum() != nil {
 			return common.InlineDatumsNotSupportedError{
 				PlutusVersion: "PlutusV1",
 			}
 		}
 	}
-
-	// Check reference scripts on reference inputs
-	for _, refInput := range tx.ReferenceInputs() {
-		utxo, err := ls.UtxoById(refInput)
-		if err != nil {
-			return common.ReferenceInputResolutionError{
-				Input: refInput,
-				Err:   err,
+	for _, output := range tx.Outputs() {
+		if output == nil {
+			continue
+		}
+		if output.Datum() != nil {
+			return common.InlineDatumsNotSupportedError{
+				PlutusVersion: "PlutusV1",
 			}
 		}
-		if utxo.Output == nil {
-			continue
-		}
-		script := utxo.Output.ScriptRef()
-		if script == nil {
-			continue
-		}
-		switch script.(type) {
-		case common.PlutusV1Script:
+		// A reference script on a produced output is equally unrepresentable in
+		// the V1 context.
+		if output.ScriptRef() != nil {
 			return common.InlineDatumsNotSupportedError{
 				PlutusVersion: "PlutusV1",
 			}
 		}
 	}
-
-	// Per CIP-33, also check reference scripts on regular (spent) inputs
-	for _, input := range tx.Inputs() {
-		utxo, err := ls.UtxoById(input)
-		if err != nil {
-			// Skip errors - BadInputsUtxo will catch this
-			continue
-		}
-		if utxo.Output == nil {
-			continue
-		}
-		script := utxo.Output.ScriptRef()
-		if script == nil {
-			continue
-		}
-		switch script.(type) {
-		case common.PlutusV1Script:
-			return common.InlineDatumsNotSupportedError{
-				PlutusVersion: "PlutusV1",
-			}
-		}
-	}
-
+	// Deliberately no restriction on the mere presence of reference inputs.
+	// Asserting one here failed the Conway conformance vector
+	// "UTXOS/can use reference scripts" (tx 3), which expects success for a
+	// transaction that requires a PlutusV1 script and carries reference
+	// inputs. The vector is authoritative; a reading of the V1 context
+	// translation that forbids reference inputs outright is not what the
+	// reference implementation does at this protocol version.
 	return nil
 }
 
--- a/ledger/babbage/rules.go
+++ b/ledger/babbage/rules.go
@@ -240,14 +240,13 @@ func UtxoValidateCostModelsPresent(
 	return nil
 }
 
-// UtxoValidateInlineDatumsWithPlutusV1 rejects transactions that use inline datums with PlutusV1 scripts
-// Inline datums are a Babbage-era feature and are only supported with PlutusV2+
 // UtxoValidateInlineDatumsWithPlutusV1 rejects a transaction that requires a
-// PlutusV1 script while using a feature the PlutusV1 script context cannot
-// represent. PlutusV1 predates inline datums and reference scripts, so a
-// transaction that needs a V1 script and carries either is invalid.
+// PlutusV1 script while carrying an inline datum. Inline datums are a Babbage
+// feature the PlutusV1 script context cannot represent, so translating an output
+// that carries one into the V1 context fails.
 //
-// Two properties matter, and getting either wrong diverges from cardano-node:
+// Three properties matter, and getting any of them wrong diverges from
+// cardano-node:
 //
 //   - Only a *needed* script constrains the transaction. A V1 script that is
 //     merely reachable -- sitting in the witness set, or as a reference script
@@ -255,14 +254,32 @@ func UtxoValidateCostModelsPresent(
 //     make the transaction invalid. Rejecting on availability turns an ordinary
 //     transaction that happens to spend a UTxO carrying an unrelated V1
 //     reference script into a permanent validation failure.
-//   - The disqualifying feature is transaction-wide. An inline datum counts
-//     wherever the V1 context would have to represent it: on a consumed input,
-//     on a reference input, or on one of this transaction's own outputs.
+//   - The inline datum is disqualifying wherever the V1 context would have to
+//     represent it: on a consumed input, on a reference input, or on one of this
+//     transaction's own outputs. cardano-ledger's Conway PlutusV1 instance runs
+//     transTxOutV1 over all three -- inputs via transTxInInfoV1, reference
+//     inputs via the same function under mapM_, and outputs directly
+//     (eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs, instance
+//     EraPlutusTxInfo 'PlutusV1 ConwayEra).
+//   - A reference script is *not* disqualifying, anywhere. Conway's
+//     transTxOutV1 checks only the inline datum; unlike the Babbage-era
+//     transTxOutV1 it shadows, it has no ReferenceScriptsNotSupported branch.
+//     So neither a reference script on a consumed or reference input nor one on
+//     a produced output invalidates a V1-needing transaction.
 //
 // Datum presence is read through the TransactionOutput interface rather than a
 // concrete Babbage type assertion, so outputs wrapped by a later era are still
 // inspected. Datum-*hash* outputs are correctly not treated as inline, because
 // Datum() reports nil for them.
+//
+// Conway semantics are applied uniformly, including for the Babbage era that
+// also registers this rule and for Dijkstra, which delegates to it through
+// Conway. cardano-ledger's Babbage-era PlutusV1 instance is stricter than
+// Conway's on both counts above: its transTxOutV1 does carry a
+// ReferenceScriptsNotSupported branch, and it rejects a transaction that needs a
+// V1 script and carries any reference input at all. Splitting the rule per era
+// needs Babbage-era conformance vectors, which the suite does not contain, so
+// the divergence stays here rather than in untested era-specific code.
 func UtxoValidateInlineDatumsWithPlutusV1(
 	tx common.Transaction,
 	slot uint64,
@@ -287,6 +304,11 @@ func UtxoValidateInlineDatumsWithPlutusV1(
 	if !needsV1 {
 		return nil
 	}
+	// Datum() only, deliberately: ScriptRef() is not consulted on a resolved
+	// input, nor on a produced output below. The Conway conformance vector
+	// "UTXOS/can use regular inputs for reference" expects success for a
+	// transaction that needs a PlutusV1 script and spends a regular input
+	// carrying a reference script, so rejecting that shape fails the suite.
 	for _, utxo := range view.AllResolvedInputs() {
 		if utxo.Output != nil && utxo.Output.Datum() != nil {
 			return common.InlineDatumsNotSupportedError{
@@ -299,13 +321,6 @@ func UtxoValidateInlineDatumsWithPlutusV1(
 			continue
 		}
 		if output.Datum() != nil {
-			return common.InlineDatumsNotSupportedError{
-				PlutusVersion: "PlutusV1",
-			}
-		}
-		// A reference script on a produced output is equally unrepresentable in
-		// the V1 context.
-		if output.ScriptRef() != nil {
 			return common.InlineDatumsNotSupportedError{
 				PlutusVersion: "PlutusV1",
 			}
--- a/ledger/babbage/rules_plutusv1_internal_test.go
+++ b/ledger/babbage/rules_plutusv1_internal_test.go
@@ -323,9 +323,11 @@ func TestInlineDatumsPlutusV1_NeededScriptFromReferenceScriptRejected(
 		utxo(spend, inlineDatumOutput(scriptAddr(t, v1))),
 		utxo(ref, scriptRefOutput(keyAddr(t), v1)),
 	})
-	require.Error(
+	var want common.InlineDatumsNotSupportedError
+	require.ErrorAs(
 		t,
 		err,
+		&want,
 		"a needed PlutusV1 script supplied by a reference script must be detected",
 	)
 }
--- a/ledger/babbage/rules_plutusv1_internal_test.go
+++ b/ledger/babbage/rules_plutusv1_internal_test.go
@@ -1,0 +1,310 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package babbage
+
+import (
+	"errors"
+	"math/big"
+	"testing"
+
+	"github.com/blinklabs-io/gouroboros/cbor"
+
+	"github.com/blinklabs-io/gouroboros/ledger/common"
+	"github.com/blinklabs-io/gouroboros/ledger/mary"
+	"github.com/blinklabs-io/gouroboros/ledger/shelley"
+	"github.com/blinklabs-io/plutigo/data"
+)
+
+// UtxoValidateInlineDatumsWithPlutusV1 had no test coverage at all, which is
+// how it came to reject transactions cardano-node accepts. These pin both
+// directions: a PlutusV1 script that is merely reachable must not invalidate a
+// transaction, and one that is actually required must, wherever the
+// V1-incompatible feature sits.
+
+func testInput(b byte, idx uint32) shelley.ShelleyTransactionInput {
+	return shelley.NewShelleyTransactionInput(
+		hexRepeat(b),
+		int(idx), // #nosec G115 -- test index
+	)
+}
+
+func hexRepeat(b byte) string {
+	const hexDigits = "0123456789abcdef"
+	out := make([]byte, 64)
+	for i := range out {
+		out[i] = hexDigits[int(b)%16]
+	}
+	return string(out)
+}
+
+func inputSet(
+	inputs ...shelley.ShelleyTransactionInput,
+) shelley.ShelleyTransactionInputSet {
+	return shelley.NewShelleyTransactionInputSet(inputs)
+}
+
+func keyAddr(t *testing.T) common.Address {
+	t.Helper()
+	addr, err := common.NewAddressFromParts(
+		common.AddressTypeKeyNone,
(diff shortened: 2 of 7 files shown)


ledger/babbage/rules_plutusv1_internal_test.go
```diff
@@ -0,0 +1,310 @@
...
+		Body: BabbageTransactionBody{
+			TxInputs: inputSet(inlineIn, refScriptIn),
+		},
+	}
+	if err := runRule(t, tx, []common.Utxo{
+		utxo(inlineIn, inlineDatumOutput(key)),
+		utxo(refScriptIn, scriptRefOutput(key, v1)),
+	}); err != nil {
+		t.Fatalf("expected accept, got %v", err)
+	}
```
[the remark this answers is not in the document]
wolf31o2: Converted in `e14df6a`. The file now uses `require.NoError`, `require.ErrorAs` and `require.Error`, with no raw `t.Fatalf` or manual `errors.As` left (`grep -c` returns 0).

The one remaining `errors.New` is the local `utxoOnlyLedgerState` mock returning "utxo not found" for an unresolved input, which is production-shaped behavior rather than an assertion.

ledger/babbage/rules.go
```diff
@@ -241,92 +242,82 @@ func UtxoValidateCostModelsPresent(
...
 		return nil
 	}
-
-	// Check if transaction uses PlutusV1 scripts in witnesses
-	witnesses := tx.Witnesses()
-	if witnesses != nil {
-		v1Scripts := witnesses.PlutusV1Scripts()
-		if len(v1Scripts) > 0 {
+	for _, utxo := range view.AllResolvedInputs() {
+		if utxo.Output != nil && utxo.Output.Datum() != nil {
```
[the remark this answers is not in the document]
wolf31o2: Not reproducible — this case is already covered, and `e14df6a` adds a test pinning it.

`availableScripts` collects `utxo.Output.ScriptRef()` from **every** resolved input, consumed and reference alike, not only witness-set scripts:

```go
for _, utxo := range resolved {
    if utxo.Output == nil { continue }
    if s := utxo.Output.ScriptRef(); s != nil { out[s.Hash()] = s }
}
```

`AllResolvedInputs()` supplies both halves, and `neededScripts` resolves each purpose's hash against that same map, so a PlutusV1 script whose only copy is a reference script is found whenever a purpose requires it.

`TestInlineDatumsPlutusV1_NeededScriptFromReferenceScriptRejected` constructs exactly that shape: a UTxO locked by the V1 script carrying an inline datum, the script supplied only as a reference script on a reference input, and no witness scripts at all. The rule rejects it.

The distinction `Needed` exists to preserve is orthogonal and still holds — an unrelated reference script that no purpose requires does not invalidate the transaction, which `TestInlineDatumsPlutusV1_UnneededReferenceScriptAccepted` pins.

(Correcting this comment: it previously carried the reply meant for the testify thread, posted here by a scripting error on my side.)

ledger/common/script/needed.go
```diff
@@ -0,0 +1,224 @@
...
+				CredType:   lcommon.CredentialTypeScriptHash,
+				Credential: addr.StakeKeyHash(),
+			},
+		})
+	}
+	for voter := range tx.VotingProcedures() {
+		if voter == nil {
+			continue
+		}
+		keep(ScriptPurposeVoting{Voter: *voter})
```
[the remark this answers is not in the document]
wolf31o2: Correct on the mechanism; fixed in `e14df6a`.

`ScriptPurposeVoting.ScriptHash()` returns `lcommon.ScriptHash(lcommon.NewBlake2b224(s.Voter.Hash[:]))` unconditionally, with no check on `Voter.Type`, so a key-hash voter yielded its key hash typed as a script hash. `neededScripts` now filters to the two script voter types before constructing the purpose:

```go
case lcommon.VoterTypeConstitutionalCommitteeHotScriptHash,
    lcommon.VoterTypeDRepScriptHash:
```

One correction to the finding's scope: it cannot cause rule 18 to *skip* a needed PlutusV1 script. For script voters `Voter.Hash` already is the script hash, so those resolved correctly before this change. The reachable defect was the other direction — a key hash looked up in `Available` and, on a hash collision, entered as a requirement. Improbable rather than impossible; the filter closes it either way.

`ScriptPurposeCertifying.ScriptHash()` was checked for the same shape and does not need a filter: it returns `ScriptHash{}` unless `cred.CredType == lcommon.CredentialTypeScriptHash`, and `neededScripts` skips a zero hash.

(Correcting this comment: it previously carried the reply meant for the testify thread, posted here by a scripting error on my side.)

ledger/common/script/needed.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+		for _, s := range witnesses.PlutusV1Scripts() {
+			out[s.Hash()] = s
+		}
+		for _, s := range witnesses.PlutusV2Scripts() {
+			out[s.Hash()] = s
+		}
+		for _, s := range witnesses.PlutusV3Scripts() {
+			out[s.Hash()] = s
+		}
```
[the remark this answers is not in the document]
wolf31o2: Correct; fixed in `e14df6a` using `lcommon.PlutusV4ScriptsFromWitnessSet` as suggested.

Worth recording that it does not change rule 18's verdict — that rule asks only whether a *PlutusV1* script is needed, so an absent V4 entry cannot affect it. `TxScriptView` is a general-purpose view though, and a caller asking whether a V4 script is required would have got the wrong answer, so the omission was worth closing on its own terms.

(This thread auto-resolved when the push touched the surrounding lines; replying so the disposition is on the record rather than inferred from the resolve.)

ledger/babbage/rules_plutusv1_internal_test.go
```diff
@@ -0,0 +1,310 @@
+// Copyright 2026 Blink Labs Software
```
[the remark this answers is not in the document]
wolf31o2: Converted in `e14df6a`. The file now uses `require.NoError`, `require.ErrorAs` and `require.Error` throughout; `grep -c 't.Fatalf\|t.Fatal('` over the file returns 0.

The one remaining `errors.New` is the local `utxoOnlyLedgerState` mock returning "utxo not found" for an unresolved input. That is production-shaped behavior the rule under test consumes, not an assertion, so it stays.

(Auto-resolved by the push; replying for the record.)

ledger/babbage/rules_plutusv1_internal_test.go
```diff
@@ -0,0 +1,331 @@
...
+	err := runRule(t, tx, []common.Utxo{
+		// Spending a UTxO locked by the V1 script, carrying an inline datum.
+		utxo(spend, inlineDatumOutput(scriptAddr(t, v1))),
+		utxo(ref, scriptRefOutput(keyAddr(t), v1)),
+	})
+	require.Error(
+		t,
+		err,
+		"a needed PlutusV1 script supplied by a reference script must be detected",
+	)
```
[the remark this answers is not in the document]
wolf31o2: Correct; fixed in `8e479c1`.

```go
var want common.InlineDatumsNotSupportedError
require.ErrorAs(t, err, &want, "...")
```

The looser assertion would have passed on an input-resolution failure, or any other error the rule can return, so the test could stay green while the property it names — that a reference-script-supplied PlutusV1 script is detected *as a V1 violation* — silently regressed. Every other reject case in the file already asserted the concrete type; this one was the outlier because it started life as a probe to settle a different question.

(Auto-resolved by the push; replying so the disposition is recorded rather than inferred.)

ledger/babbage/rules.go
```diff
@@ -241,92 +242,82 @@ func UtxoValidateCostModelsPresent(
...
-			continue
-		}
-		switch script.(type) {
-		case common.PlutusV1Script:
+		// A reference script on a produced output is equally unrepresentable in
+		// the V1 context.
+		if output.ScriptRef() != nil {
 			return common.InlineDatumsNotSupportedError{
 				PlutusVersion: "PlutusV1",
 			}
```
chrisguiney: This is the one I would like to settle before merge: as far as I can tell this rejection is not backed by anything, and its stated justification is contradicted by the vectors.

In cardano-ledger an output and a resolved input go through the **same** `transTxOutV1` translation. So "equally unrepresentable in the V1 context" is precisely the claim that the conformance suite falsifies on the input side — I patched the rule to reject a reference script on a consumed input and got:

```
--- FAIL: TestRulesConformanceVectors/.../UTXOS/can_use_regular_inputs_for_reference
    harness.go:252: event 3 failed: expected success but got failure:
      (inline datums are not supported with PlutusV1 scripts ...)
```

I then removed *this* check and re-ran conformance: **still green**. Nothing in the vector set requires it; only the new unit test pins it.

I also instrumented `needsV1` across the whole suite. It fires 98 times, and in 8 of those the needed V1 script comes from a reference script with no witness copy at all (`witnessV1=0`). So V1-needed-alongside-reference-scripts is well represented in the vectors and consistently accepted.

That leaves this as a rejection added on reasoning alone, in the one rule whose false positive halted a node's sync — the same class of bug this PR exists to remove. My suggestion is to drop it and keep the PR to the datum fix plus the needed-scripts scoping, or to find a vector that demands it.

If it does stay, it needs its own error type. Reporting a reference-script violation as `InlineDatumsNotSupportedError` will mislead whoever debugs it, and cardano-ledger distinguishes the two; `common/errors.go` has no `ReferenceScriptsNotSupportedError` today.

Happy to be wrong here if you have a vector or a ledger reference I did not find — this is the one place I could not reconcile the code with the evidence.

ledger/babbage/rules.go
```diff
@@ -241,92 +242,82 @@ func UtxoValidateCostModelsPresent(
...
-	if witnesses != nil {
-		v1Scripts := witnesses.PlutusV1Scripts()
-		if len(v1Scripts) > 0 {
+	for _, utxo := range view.AllResolvedInputs() {
+		if utxo.Output != nil && utxo.Output.Datum() != nil {
 			return common.InlineDatumsNotSupportedError{
 				PlutusVersion: "PlutusV1",
 			}
 		}
 	}
```
chrisguiney, line 318: Small suggestion, and it would have preempted the P1 above: this loop checks `Datum()` on every resolved input but never `ScriptRef()`, while the output loop below checks both. That asymmetry is deliberate and correct — I verified it — but nothing here says so.

There is an excellent note further down explaining why reference-*input presence* is not asserted, naming its vector. One more line in the same spirit, naming `UTXOS/can_use_regular_inputs_for_reference` as the vector that requires accepting a reference script on a consumed input, would make the asymmetry self-explaining to the next reader — human or bot.

Separately: `AllResolvedInputs()` allocates a fresh N+M slice on every call, and it is called twice per invocation — once inside `NewTxScriptView` at `needed.go:106` and again here. Caching the combined slice on the view would drop one allocation per transaction.

ledger/common/script/needed.go
```diff
@@ -0,0 +1,246 @@
...
+		for _, s := range lcommon.PlutusV4ScriptsFromWitnessSet(witnesses) {
+			out[s.Hash()] = s
+		}
+	}
+	for _, utxo := range resolved {
+		if utxo.Output == nil {
+			continue
+		}
+		if s := utxo.Output.ScriptRef(); s != nil {
+			out[s.Hash()] = s
```
chrisguiney, line 158: Worth flagging for a follow-up: this is now the third copy of this map.

`conway/rules.go:2390` and `dijkstra/rules.go:416` already build the identical structure — witness V1/V2/V3/V4 plus reference scripts from resolved inputs — each alongside its own input resolution and `SortInputs`. `TxScriptView` is exactly the abstraction both of those want.

The drift that duplication causes is not hypothetical here: the V4 omission cubic caught in `e14df6a` is precisely it, appearing in a copy that was brand new. Three copies means the next language version has three places to remember.

I do not think it belongs in this PR — it is already 635 lines and the rule fix should land on its own. But an explicit follow-up issue to adopt `TxScriptView` at those two sites would be worth opening while the shape is fresh.

ledger/common/script/needed.go
```diff
@@ -0,0 +1,246 @@
...
+				Err:   err,
+			}
+		}
+		view.ResolvedReferenceInputs = append(
+			view.ResolvedReferenceInputs,
+			utxo,
+		)
+	}
+	view.Available = availableScripts(tx, view.AllResolvedInputs())
+	view.Needed = neededScripts(tx, view)
```
chrisguiney, line 125: Performance note for the hot path: the view resolves everything and walks every purpose unconditionally, before any caller gets a chance to gate on something cheap.

The previous rule could scan inputs for a datum and return. Now every transaction pays N+M `UtxoById` calls, two map builds, a sort, and a script-hash computation per purpose — including the overwhelming majority that have no PlutusV1 script anywhere. In dingo each `UtxoById` is a database read plus a full CBOR decode, and this runs per transaction for the whole chain during sync.

Since rule 18 only ever asks "is any *needed* script V1", the purpose walk can be skipped entirely when nothing in `Available` is a V1 script. Making `Needed` lazy, or exposing an `Available`-side pre-filter, would keep the general API while letting the common case exit after `availableScripts`.

Not blocking, and I have not measured it — the point is structural rather than a number I can quote.

ledger/common/script/needed.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+// neededScripts walks every script purpose the transaction requires and keeps
+// the available script each one resolves to.
+//
+// The walk order matches the canonical redeemer order per tag, so a caller that
+// also cares about redeemer indices sees the same sequence: spending inputs
+// (sorted), minting policies (sorted), certificates, withdrawals (sorted),
+// voters (sorted), proposal procedures. Eras without voting or proposing simply
+// produce none of those.
```
chrisguiney: Doc nit: this promises an order the code does not produce.

Inputs and mint policies genuinely are sorted (`SortInputs`, `slices.SortFunc`). But withdrawals (`:222`) and voting procedures iterate Go maps, so those two are in randomized order, not sorted.

It is harmless today because the result is a `map` and order cannot escape. What makes it worth fixing is the sentence explaining *why* the order matters — "a caller that also cares about redeemer indices sees the same sequence" — since that is a promise to a future caller that the code would quietly break the first time someone returns a slice instead. Either sort those two or drop the claim for them.

chrisguiney (dismissed):
This is a solid piece of work, and the framing in the description — wrong in both directions at once — is exactly right. The needed-vs-available distinction is the correct fix, and finding that the rule had *no* test coverage anywhere explains how the over-broad behavior survived this long.

A few things I checked and found sound, since they were the places I expected trouble:

- **`neededScripts` purpose resolution.** I walked each hash path rather than trusting the enumeration. `ScriptPurposeSpending.ScriptHash()` dereferences `s.Input.Output` unguarded (`purpose.go:88`), but the caller nil-checks first. `ScriptPurposeCertifying` self-filters on `CredentialTypeScriptHash`. And the new `voterUsesScriptCredential` filter is genuinely load-bearing, since `ScriptPurposeVoting.ScriptHash()` really is unconditional (`purpose.go:189-191`).
- **`StakeKeyHash()` for withdrawals.** I went looking for a bug here, expecting reward addresses to carry their credential in the payment slot and silently yield an empty hash. They do not — types `0b1110`/`0b1111` assert an *empty* payment payload and put the credential in `stakingPayload` (`address.go:197`, `:217-227`), so `ScriptPurposeRewarding` resolves correctly.
- **`slices.SortFunc` on `mint.Policies()`.** Safe — `Policies()` allocates a fresh slice per call (`common.go:446`), so the in-place sort cannot mutate transaction state.
- **The reference-input decision.** Well-founded, well-documented, and pinning it with a test so it cannot be reintroduced is the right instinct.
- `go test ./...`, `go test ./internal/test/conformance/`, and `go build` all pass locally; all 14 CI checks are green.

**On cubic's open P1 at `rules.go:291` — it is a false positive, and I confirmed it empirically rather than by reading.** I patched the rule to reject a reference script on a consumed input and ran the conformance suite: it fails `UTXOS/can_use_regular_inputs_for_reference` at event 3, "expected success but got failure". Your reply disputes it on discovery grounds, which is also correct; the vector settles it outright. I have posted the probe output on that thread.

One item I would like resolved before this merges, on `rules.go:306`. The rest are follow-ups and a doc nit.

ledger/babbage/rules.go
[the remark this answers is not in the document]
chrisguiney: Independent confirmation that this finding should not be acted on, from the other direction: rather than checking whether the needed script is *discovered*, I checked what happens if the rule does what the finding asks.

Patching `UtxoValidateInlineDatumsWithPlutusV1` to reject a reference script on a consumed input, then running the conformance suite:

```
--- FAIL: TestRulesConformanceVectors/eras/conway/impl/dump/Conway/Imp/
          ConwayImpSpec_-_Version_10/UTXOS/can_use_regular_inputs_for_reference
    harness.go:252: event 3 failed: expected success but got failure:
      transaction validation failed (inline datums are not supported with
      PlutusV1 scripts - inline datums are a Babbage feature only available
      for PlutusV2+)
```

So the vector set positively requires accepting the shape this finding asks to reject: a transaction that needs a PlutusV1 script and spends a regular input carrying a reference script.

Instrumenting `needsV1` across the whole suite is consistent with that — it fires 98 times, and 8 of those have `witnessV1=0`, meaning the needed V1 script's only copy is a reference script. Those all pass today.

Both readings therefore agree: the current behavior is correct as written. `@wolf31o2`'s explanation of *why* the script is still discovered is the complementary half of the same answer.

ledger/babbage/rules.go
wolf31o2: Settled, and you were right — removed in `7f9fce84`.

The decisive evidence is upstream: **Conway defines its own `transTxOutV1` and it shadows Babbage's.**

`eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs:324`
```haskell
transTxOutV1 txOutSource txOut = do
  when (isSJust (txOut ^. dataTxOutL)) $ do
    Left $ inject $ InlineDatumsNotSupported @era txOutSource
```

`eras/babbage/impl/src/Cardano/Ledger/Babbage/TxInfo.hs:119` has the same function *plus* a `ReferenceScriptsNotSupported` branch. Conway deleted it. So a reference script is not disqualifying anywhere in Conway — not on a produced output, and not on a consumed or reference input.

Both of your claims check out:

- **Unexercised.** Instrumented the rule and correlated probes against all 314 vectors: the produced-output reference-script branch fired **0** times, while `needsV1` fired 49 times per pass and the input-side reference-script path fired in 4 vectors. Restoring the check leaves conformance at exit 0 — so conformance could never have caught this, in either direction.
- **The input side already proves the semantics.** Verified empirically: patching an equivalent rejection onto resolved inputs, placed after the `needsV1` gate, fails two vectors — `UTXOS/can_use_reference_scripts` and `UTXOS/can_use_regular_inputs_for_reference`. (Worth recording that my first attempt at this probe landed unreachable and appeared to *disprove* it; placement after the gate is what makes it fire.)

One refinement to the stated cause. It is not that the input side falsifies the output side by symmetry through a shared `transTxOutV1` — within Conway both sides simply have no such check. The asymmetry that matters is between *eras*, not between inputs and outputs.

That era asymmetry is now documented rather than silently applied: this shared implementation is registered by babbage and reached by conway and dijkstra, and cardano-ledger's Babbage-era V1 instance is stricter on both counts (it keeps `ReferenceScriptsNotSupported`, and rejects any reference input at all). Applying Conway semantics uniformly is a deliberate choice, and splitting per era needs Babbage-era vectors the suite does not contain — it only carries Conway dumps.

`TestInlineDatumsPlutusV1_ScriptRefOnProducedOutputRejected` is inverted to `..._Accepted`. Restoring the check fails it while conformance stays green, so that unit test is the only thing holding this line either way.

ledger/babbage/rules.go
wolf31o2: Both halves taken, in `7f9fce84`.

The input loop now carries a comment naming `UTXOS/can_use_regular_inputs_for_reference` as the vector that pins it, and the doc gained a third bullet stating plainly that a reference script is not disqualifying anywhere, with the Conway `TxInfo.hs` citation. You are right that saying so up front would have preempted the P1 — the doc previously said "inline datums and reference scripts", which is what invited it.

On the traversal: `TxScriptView` now caches `allResolvedInputs`, populated once in `NewTxScriptView`, with `AllResolvedInputs()` falling back to building on demand for a hand-assembled view. `TestTxScriptViewAllResolvedInputsWithoutCache` covers that fallback and fails if the accessor returns the cache unconditionally.

ledger/common/script/needed.go
wolf31o2: Agreed, and deferred as you scoped it — no code change here.

Confirmed the count at current line numbers: `ledger/conway/rules.go:2393` and `ledger/dijkstra/rules.go:626` each rebuild the same witness-V1/V2/V3/V4-plus-resolved-input-reference-scripts map alongside their own resolution and `SortInputs`. With `availableScripts` that is three copies.

Collapsing them onto `TxScriptView` is the right end state but it touches two more rule paths, so it wants its own change and its own conformance run rather than riding along here. Tracked for follow-up.

ledger/common/script/needed.go
wolf31o2: Partly taken in `7f9fce84`, and I stopped short of the structural half deliberately.

Added the cheap, provably-equivalent guard: `neededScripts` returns an empty map when `len(view.Available) == 0`, since `keep` only ever inserts a hash already present in `Available`. That takes a script-free transaction — the overwhelming majority during sync — off `SortInputs`, the by-id map build, and one script-hash computation per purpose. Inverting the guard to `>= 0` fails 4 babbage tests and 2 script tests, so it is load-bearing rather than decorative.

Not done: making `Needed` lazy, or pre-filtering on the `Available` side. Both change the public shape of a type this PR is introducing, you marked the finding non-blocking and unmeasured, and the repo defines no benchmark on this path — so I would be reshaping an API on an argument neither of us has measured. Left as a follow-up where it can be benchmarked first.

ledger/common/script/needed.go
wolf31o2: Correct, and I dropped the claim rather than making it true.

Verified the drift: `Withdrawals() map[*Address]*big.Int` (`ledger/common/tx.go:61`) and `VotingProcedures map[*Voter]map[*GovActionId]VotingProcedure` (`ledger/common/gov.go:30`) are both Go maps, so those two loops iterate in randomized order while the doc promised sorted.

Sorting them is not a doc fix. Both maps are keyed by **pointer**, so ordering the keys cannot produce a canonical order — a correct fix has to derive a canonical key from each value, which is a real design decision. The doc now says inputs and mint policies are ordered, and that withdrawals and voters are not, naming why. Better to have the trap written down than papered over with a sort that only looks deterministic.

arepala-uml (approved):
Independent verification against current HEAD (7f9fce8):

- `go build ./...`, `go vet ./...`: clean
- `go test ./ledger/babbage/... ./ledger/common/script/...`: pass
- `go test -v ./internal/test/conformance/...`: 315/315 vectors pass
- `golangci-lint run ./ledger/...`: 0 issues

Confirmed the babbage -> conway -> dijkstra delegation actually reaches this rule, and independently re-derived the reference-script/output design decision against Conway's transTxOutV1 semantics and the two conformance vectors it depends on (UTXOS/can use reference scripts, UTXOS/can use regular inputs for reference).

CodeRabbit and cubic findings: all have a corresponding fix or evidence-backed rebuttal in the current code. Two cubic threads (rules.go:313, needed.go:265) remain formally unresolved on GitHub despite the fix already landing in e14df6a - thread hygiene only, no outstanding code issue.

chrisguiney's five review comments: all addressed in code, verified against current HEAD.
- Reference script on a produced/consumed output not disqualifying: removed in 7f9fce84, backed by conformance run and upstream transTxOutV1 citation.
- Asymmetry comment + AllResolvedInputs caching: comment added naming the pinning vector; allResolvedInputs is now cached on TxScriptView (needed.go), confirmed in current code.
- Triplicated available-scripts map across babbage/conway/dijkstra: deferred, tracked as issue #1982.
- Needed-map laziness/performance: cheap len(Available)==0 short-circuit added; full laziness deferred, tracked as issue #1983.
- Doc claim that all purposes are walked in canonical order: corrected to name withdrawals and voting procedures as unordered (Go map iteration).

None of chrisguiney's five threads are marked resolved on GitHub yet, and no re-review has been posted since 7f9fce84 landed, though a re-review request is already pending on his account.

No merge-blocking issues found in this pass.

PR #1991, opened 2026-08-22
refactor(ledger): collapse per-era Plutus-script availability maps

### LINKED ISSUE #1982: three copies of the available-scripts map across era rule paths
`availableScripts` in `ledger/common/script/needed.go` (added in #1980) is the third copy of the same map construction. The other two:

- `ledger/conway/rules.go:2393`
- `ledger/dijkstra/rules.go:626`

Each builds the witness PlutusV1/V2/V3/V4 scripts plus reference scripts from resolved inputs, keyed by script hash, alongside its own input resolution and `SortInputs`.

Three copies of "which scripts does this transaction make available" will drift, and a divergence is a validation gap rather than a visible failure — the V4 omission #1980 had to fix in its own copy is exactly that failure mode.

## Suggested direction

Collapse the two rule-side copies onto `script.NewTxScriptView`, which already resolves inputs once and exposes `Available` and `Needed`. That also removes duplicate `UtxoById` calls: `common.LedgerState.UtxoById` is uncached, so each call is a database read plus a full CBOR decode.

## Why it is not in #1980

It reaches into two more rule paths in different eras, so it wants its own change and its own conformance run rather than riding along with a rule-18 fix. Raised by a reviewer there and deferred by agreement.

## Caution

`NewTxScriptView` resolves consumed *and* reference inputs eagerly. A caller that today resolves only one set would pay for both, so measure before swapping it into a hot path — the repository defines no benchmark on this path, and #1980 declined a related restructuring for exactly that reason.

cryptodj413:
## Problem

`ledger/common/script/needed.go`'s `TxScriptView` (added in #1980) is the
third copy of "which scripts does this transaction make available." The
other two:

- `ledger/conway/rules.go` (`UtxoValidatePlutusScripts`)
- `ledger/dijkstra/rules.go` (`validateGuardingPlutusScripts` +
  `resolvedInputsForGuardingPlutus`)

Each builds the witness PlutusV1/V2/V3/V4 scripts plus reference scripts from
resolved inputs, keyed by script hash, alongside its own input resolution.
Three independent copies drift, and a divergence is a validation gap rather
than a visible failure — the V4 omission #1980 had to fix was exactly that
failure mode.

## Change

- Export `script.ResolveTxInputs` and `script.ConcatResolvedInputs` from
  `needed.go`: the shared consumed/reference input resolution `NewTxScriptView`
  already did, now reused by Conway and Dijkstra instead of re-implemented.
  Each call site still makes exactly one `UtxoById` call per input.
- Export `script.PlutusWitnessScripts` and `script.AvailablePlutusScripts`:
  the shared "witness Plutus scripts + Plutus reference scripts" map, reused
  by `conway.UtxoValidatePlutusScripts` and Dijkstra's guarding-redeemer
  execution (including its per-sub-transaction witness merge).

## Native-script exclusion

`AvailablePlutusScripts` excludes native scripts, matching Dijkstra's
existing filtered behavior rather than Conway's previously-unfiltered one.
This is deliberate: Dijkstra's native-script guard fallback
(`TestUtxoValidateGuardingRedeemerRejectsNativeScriptGuard`) depends on a
native script being *absent* from the availability map so it falls into the
"missing script" branch and returns `ExtraRedeemerError`; a native script
present-but-unexecutable-by-the-Plutus-switch would silently pass instead.
Conway has no such branch, so applying the same exclusion there is a no-op
(a native script hits `default: continue` either way).

Verified by temporarily reverting the exclusion locally: both the new test
and the pre-existing Dijkstra test fail exactly as expected.

## Tests

New in `ledger/common/script/needed_test.go`:
- `TestAvailablePlutusScriptsExcludesNativeScripts` — pins the exclusion.
- `TestPlutusWitnessScriptsNilWitnessSet`
- `TestResolveTxInputsOrderAndErrors` — order and
  `InputResolutionError`/`ReferenceInputResolutionError` contract.

## Validation

- `make test` (root + all example modules)
- `make lint`
- `go run go.uber.org/nilaway/cmd/nilaway@latest -include-pkgs=github.com/blinklabs-io/gouroboros ./...`
- `go test -v ./internal/test/conformance/...` — 315/315 vectors pass, unchanged
- `go test -race ./...` — full suite green

Fixes #1982

---
## Summary by cubic
Collapses per-era Plutus-script availability and input resolution into shared helpers to remove drift and prevent validation gaps. Old: Conway and Dijkstra each built their own maps/resolvers; new: both use `script.PlutusWitnessScripts`, `script.AvailablePlutusScripts`, `script.ResolveTxInputs`, and `script.ConcatResolvedInputs`. Also removes duplicate `UtxoById` calls in Dijkstra’s native-script guard fallback by reusing already-resolved inputs.

- Replaces per-era loops with exports from `ledger/common/script`; still exactly one `UtxoById` call per input.
- `AvailablePlutusScripts` excludes native scripts (matches Dijkstra); in Conway this is behavior-neutral.
- Dijkstra merges sub-transaction witness scripts via `script.PlutusWitnessScripts` and checks native-guard fallback using resolved inputs; the non-resolved code path remains for other callers.
- Adds tests pinning native-script exclusion, the resolver’s order/error-type contract, and Dijkstra’s native reference-script guard rejection; conformance results stay unchanged.

<sup>Written for commit d8953249fb214b974f708ca5605750182f89ea38. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/1991?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Plutus script validation for transactions using both consumed and reference inputs.
  * Ensured reference scripts are correctly discovered while native scripts are excluded.
  * Preserved clear, typed errors when transaction inputs cannot be resolved.
  * Improved validation consistency across supported ledger eras, including Dijkstra subtransactions.
* **Tests**
  * Added coverage for input resolution order, reference inputs, missing inputs, native-script exclusion, and transactions without witness sets.

### DIFF
--- a/ledger/common/script/needed_test.go
+++ b/ledger/common/script/needed_test.go
@@ -157,10 +157,13 @@ func refOutput(s common.Script) common.TransactionOutput {
 // AvailablePlutusScripts backs conway.UtxoValidatePlutusScripts' and
 // dijkstra's guarding-redeemer execution's script-availability maps. Both
 // key an unmatched redeemer's "no script" branch on a hash's absence here,
-// and Dijkstra's native-script guard fallback (see rules_test.go's
-// TestUtxoValidateGuardingRedeemerRejectsNativeScriptGuard) depends on a
-// native script -- witness or reference -- staying absent rather than
-// present-but-unexecutable.
+// and Dijkstra's native-script guard fallback depends on a native script --
+// witness or reference -- staying absent rather than
+// present-but-unexecutable. dijkstra/rules_test.go's
+// TestUtxoValidateGuardingRedeemerRejectsNativeScriptGuard pins the witness
+// case at the rule level;
+// TestUtxoValidateGuardingRedeemerRejectsNativeReferenceScriptGuard pins the
+// reference case there.
 func TestAvailablePlutusScriptsExcludesNativeScripts(t *testing.T) {
 	v1 := common.PlutusV1Script([]byte{0x01})
 	v2 := common.PlutusV2Script([]byte{0x02})
--- a/ledger/common/script/needed_test.go
+++ b/ledger/common/script/needed_test.go
@@ -15,12 +15,16 @@
 package script_test
 
 import (
+	"errors"
 	"testing"
 
 	"github.com/blinklabs-io/gouroboros/cbor"
+	"github.com/blinklabs-io/gouroboros/ledger/babbage"
 	"github.com/blinklabs-io/gouroboros/ledger/common"
 	"github.com/blinklabs-io/gouroboros/ledger/common/script"
 	"github.com/blinklabs-io/gouroboros/ledger/conway"
+	"github.com/blinklabs-io/gouroboros/ledger/shelley"
+	mockledger "github.com/blinklabs-io/ouroboros-mock/ledger"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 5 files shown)


wolf31o2 (changes requested):
The shared helpers preserve the previous Conway and Dijkstra script-selection behavior, and the focused race suites pass. One merge blocker remains: validateGuardingPlutusScripts resolves all inputs, then its native-script fallback calls dijkstraGuardResolvesToNativeScript and reads those same UTxOs again. That breaks this change’s one-read and same-snapshot contract. Reuse the already-resolved inputs in that fallback, as the open CodeRabbit thread describes.

ledger/dijkstra/rules.go
```diff
@@ -413,28 +413,16 @@ func validateGuardingPlutusScripts(
...
-			continue
-		}
-		if _, ok := common.PlutusScriptVersion(scriptRef); ok {
-			availableScripts[scriptRef.Hash()] = scriptRef
+	availableScripts := script.AvailablePlutusScripts(tx, resolvedInputs)
+	if dijkstraTx, ok := tx.(*DijkstraTransaction); ok {
+		for _, subTx := range dijkstraTx.Body.TxSubTransactions.Items() {
+			for hash, s := range script.PlutusWitnessScripts(subTx.WitnessSet) {
+				availableScripts[hash] = s
+			}
```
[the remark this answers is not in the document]
cryptodj413: Fixed in 72af760: `dijkstraGuardResolvesToNativeScript` is split into shared `dijkstraGuardScriptHash`/`dijkstraGuardWitnessesHaveNativeScript` helpers plus a new `dijkstraGuardResolvesToNativeScriptFromResolved` that takes the `resolvedInputs` `validateGuardingPlutusScripts` already resolved instead of re-querying `ls.UtxoById`. The other caller (`dijkstraGuardNeedsPlutusRedeemer`, which has no pre-resolved inputs) is untouched.

cryptodj413:
@wolf31o2 Fixed in 72af760. `dijkstraGuardResolvesToNativeScript` is now split into shared `dijkstraGuardScriptHash`/`dijkstraGuardWitnessesHaveNativeScript` helpers plus a new `dijkstraGuardResolvesToNativeScriptFromResolved` that checks the `resolvedInputs` `validateGuardingPlutusScripts` already resolved instead of calling `ls.UtxoById` again. The other caller of the original function, `dijkstraGuardNeedsPlutusRedeemer`, has no pre-resolved inputs and is unchanged.

Validated: `go build ./...`, `gofmt -l .`, `golangci-lint run ./...`, `go test -race ./...` (full suite green, including `TestUtxoValidateGuardingRedeemerRejectsNativeScriptGuard`), `go test -v ./internal/test/conformance/...` (315/315), nilaway clean. CodeRabbit has re-reviewed the new commit and marked its thread on this resolved. Requesting another review.

ledger/common/script/needed_test.go
```diff
@@ -121,3 +125,191 @@ func TestTxScriptViewAllResolvedInputsWithoutCache(t *testing.T) {
...
+		TxOutScriptRef: &common.ScriptRef{Type: refType, Script: s},
+	}
+}
+
+// AvailablePlutusScripts backs conway.UtxoValidatePlutusScripts' and
+// dijkstra's guarding-redeemer execution's script-availability maps. Both
+// key an unmatched redeemer's "no script" branch on a hash's absence here,
+// and Dijkstra's native-script guard fallback (see rules_test.go's
+// TestUtxoValidateGuardingRedeemerRejectsNativeScriptGuard) depends on a
+// native script -- witness or reference -- staying absent rather than
```
wolf31o2: **[nit, non-blocking] The cited test does not exercise the reference half of this claim**

`TestUtxoValidateGuardingRedeemerRejectsNativeScriptGuard` supplies its native script through `WsNativeScripts` only (`ledger/dijkstra/rules_test.go:113-117`), and a witness-set native script was never a candidate for this map — `PlutusWitnessScripts` reads only the four Plutus witness fields. So that test is insensitive to the reference-script filter.

Verified by deleting the `PlutusScriptVersion` guard from `AvailablePlutusScripts` on `72af760c` (compiling tree, `go vet` exit 0):

```
--- FAIL: TestAvailablePlutusScriptsExcludesNativeScripts
    native reference script must not be an available Plutus script
ok      github.com/blinklabs-io/gouroboros/ledger/dijkstra   0.018s   <-- still passes
```

The assertion in this test at line ~210 does cover the reference case at the unit level, so the function itself is not untested. What is missing is a Dijkstra rule-level case where the guard's native script arrives as a *reference* script, and the parenthetical points a future reader at a test that cannot fail if the filter is removed. Worth either dropping "or reference" from the sentence or adding the rule-level case, so the next person auditing whether this guard is safe to simplify does not trust the wrong signal.

wolf31o2 (dismissed):

ledger/common/script/needed_test.go
cryptodj413: Fixed in d895324: added `TestUtxoValidateGuardingRedeemerRejectsNativeReferenceScriptGuard`, which puts the guard's native script on a resolved reference input (via `WithUtxoById`) instead of the witness set, and asserts `UtxoValidatePlutusScripts` still returns `ExtraRedeemerError`. Verified it fails without `AvailablePlutusScripts`' `PlutusScriptVersion` filter. Also corrected the doc comment on `TestAvailablePlutusScriptsExcludesNativeScripts` to cite each rule-level test for the case it actually covers.

cryptodj413:
Addressed the non-blocking nit in d895324: added `TestUtxoValidateGuardingRedeemerRejectsNativeReferenceScriptGuard` (rule-level, native script via reference input) and corrected the doc comment on `TestAvailablePlutusScriptsExcludesNativeScripts` to cite each test for the case it actually covers.

Validated: `go build ./...`, `gofmt -l .`, `golangci-lint run ./...`, `go test -race ./...` (full suite green), `go test -v ./internal/test/conformance/...` (315/315), nilaway clean. Confirmed the new test fails without `AvailablePlutusScripts`' `PlutusScriptVersion` filter.

wolf31o2 (approved):

PR #2030, opened 2026-08-24
fix(ledger): preserve wire CBOR for Shelley/Allegra tx bodies and witness sets

### LINKED ISSUE #1990: ShelleyTransactionBody and AllegraTransactionBody lack MarshalCBOR, inflating rebuilt transaction sizes
## Summary

`MaryTransactionBody` implements `MarshalCBOR` and returns its preserved bytes (`ledger/mary/mary.go:241`):

```go
func (b *MaryTransactionBody) MarshalCBOR() ([]byte, error) {
	if b.Cbor() != nil {
		return b.Cbor(), nil
	}
	return cbor.EncodeGeneric(b)
}
```

`ShelleyTransactionBody` and `AllegraTransactionBody` do not. Both store their decoded bytes in `UnmarshalCBOR` via `SetCbor(cborData)`, but with no `MarshalCBOR` the generic encoder re-encodes the decoded fields whenever the body is encoded as part of a containing value.

## Effect

`ShelleyBlock.Transactions` and `AllegraBlock.Transactions` construct each transaction from the block's parallel body, witness-set and metadata arrays, so the resulting transaction carries no stored transaction CBOR. `Cbor()` falls through to `MarshalCBOR`, which encodes `t.Body` generically.

`ShelleyProtocolParameterUpdate` (`ledger/shelley/pparams.go:154`) tags none of its 16 optional fields `omitempty`, and `AllegraProtocolParameterUpdate` is an alias for it, so that re-encoding emits an explicit CBOR null for every absent field.

Measured for preprod transaction `a00696a0c2d70c381a265a845e43c55e1d00f96b27c06defc015dc92eb206240` (epoch 4, the update proposing protocol version 3.0), whose seven `protocol_param_update` maps each hold a single entry:

| decode path | `len(tx.Cbor())` |
| --- | --- |
| `NewShelleyTransactionFromCbor` | 1156 (wire) |
| rebuilt the way `ShelleyBlock.Transactions` does | 1366 |
| `NewAllegraTransactionFromCbor` | 1156 |
| rebuilt the way `AllegraBlock.Transactions` does | 1366 |
| `NewMaryTransactionFromCbor` | 1156 |
| rebuilt the way `MaryBlock.Transactions` does | 1156 |

`common.TxSizeForFee` returns `len(tx.Cbor())` for pre-Alonzo transactions, so `shelley.MinFeeTx` computes 215485 rather than 206245 for a transaction whose declared fee is exactly `44 * 1156 + 155381 = 206245`, and `shelley.UtxoValidateFeeTooSmallUtxo` rejects it. `shelley.UtxoValidateMaxTxSizeUtxo` measures the same inflated 1366 bytes.

## Reproduction

In `ledger/shelley`, with `preprodUpdateTxHex` the 1156-byte wire encoding of the transaction above:

```go
raw, _ := hex.DecodeString(preprodUpdateTxHex)
wireTx, _ := NewShelleyTransactionFromCbor(raw)
rebuilt := &ShelleyTransaction{Body: wireTx.Body, WitnessSet: wireTx.WitnessSet}
// len(raw) == 1156, len(rebuilt.Cbor()) == 1366, both hash to the same tx id
```

Present at `v0.193.3` and at `7f9fce84e569`.

## Fix

Add the `MaryTransactionBody.MarshalCBOR` shape to `ShelleyTransactionBody` and `AllegraTransactionBody`. Verified locally on `v0.193.3`: `rebuilt.Cbor()` becomes byte-identical to the 1156-byte wire encoding, and `go test ./ledger/...` passes.

`ShelleyTransactionWitnessSet` has the same gap, and Allegra and Mary both embed it: it stores its bytes in `UnmarshalCBOR` and has no `MarshalCBOR`, so a rebuilt witness set is also re-encoded from decoded fields. Adding the method there as well keeps all three eras on preserved witness bytes; the same local run passes with all three methods added.

## Downstream

blinklabs-io/dingo#3255 works around this in `ledger/eras` by rebuilding the wire size from preserved component bytes and replacing the upstream Shelley and Allegra fee and max-size rules. That workaround can be deleted once this ships.

cryptodj413:
## Summary

`ShelleyTransactionBody` and `AllegraTransactionBody` store their decoded CBOR via `SetCbor()` but had no `MarshalCBOR()`, so re-encoding a decoded body (as `ShelleyBlock.Transactions()` / `AllegraBlock.Transactions()` do when constructing a transaction from a block's parallel body/witness/metadata arrays) fell through to the generic encoder instead of returning the preserved bytes. `ShelleyTransactionWitnessSet` had the same gap and is shared by Allegra and Mary.

The generic re-encode inflates size because `ShelleyProtocolParameterUpdate` has no `omitempty` tags on its 16 fields, so re-encoding an update proposal emits an explicit CBOR null for every unset field. This overstates `len(tx.Cbor())`, which `common.TxSizeForFee` uses for pre-Alonzo fee/size checks, so `shelley.UtxoValidateFeeTooSmallUtxo` and `UtxoValidateMaxTxSizeUtxo` can reject a transaction whose declared fee/size were actually correct.

## Fix

Add the same `Cbor()`-first `MarshalCBOR()` shape already used by `MaryTransactionBody` to:
- `ShelleyTransactionBody`
- `AllegraTransactionBody`
- `ShelleyTransactionWitnessSet` (shared by Shelley, Allegra, and Mary)

## Testing

- Added `TestShelleyTransactionBody_MarshalCBOR_PreservesWireBytes`, which builds minimal wire bytes for a body carrying a protocol-parameter-update proposal, decodes it, and asserts `MarshalCBOR()` returns the preserved bytes verbatim and that a transaction rebuilt from the decoded body/witness set (the same pattern `ShelleyBlock.Transactions()` uses) round-trips without inflation.
  - Confirmed this test fails to compile against the pre-fix code (`ShelleyTransactionBody` had no `MarshalCBOR` method), then passes with the fix.
- `go build ./...`
- `go test ./...` (all packages pass, including `internal/test/conformance`)
- `gofmt -l` clean

Fixes #1990

---
## Summary by cubic
Preserves wire CBOR for Shelley and Allegra transaction bodies and the shared witness set to prevent size inflation that breaks pre-Alonzo fee/size validation. Before: generic re-encoding emitted CBOR nulls for unset protocol-parameter-update fields; after: MarshalCBOR returns preserved bytes when present.

- Add MarshalCBOR to ShelleyTransactionBody, AllegraTransactionBody, and ShelleyTransactionWitnessSet (shared by Shelley/Allegra/Mary); prefer preserved Cbor() bytes, else fall back to generic encoding.
- Keep rebuilt transaction size equal to wire size, restoring accurate len(tx.Cbor()) for fee and max-size checks.
- Add regression tests for Shelley and Allegra bodies and a witness-set fixture using an indefinite-length VkeyWitnesses array to guarantee the preserved-bytes path is exercised; update tests to use `testify` require/assert.

<sup>Written for commit c4719f71145ce36b5c257fd2f3ea442ec2970f6b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2030?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction re-encoding to preserve original CBOR data when available.
  * Fixed handling of Shelley transactions with sparse protocol parameter updates.
  * Ensured rebuilding transactions with empty witness sets does not alter encoded data.

* **Tests**
  * Added regression coverage for preserving transaction wire bytes during re-marshaling.

### DIFF
--- a/ledger/shelley/shelley.go
+++ b/ledger/shelley/shelley.go
@@ -318,6 +318,13 @@ func (b *ShelleyTransactionBody) UnmarshalCBOR(cborData []byte) error {
 	return nil
 }
 
+func (b *ShelleyTransactionBody) MarshalCBOR() ([]byte, error) {
+	if b.Cbor() != nil {
+		return b.Cbor(), nil
+	}
+	return cbor.EncodeGeneric(b)
+}
+
 func (b *ShelleyTransactionBody) Inputs() []common.TransactionInput {
 	items := b.TxInputs.Items()
 	ret := make([]common.TransactionInput, len(items))
@@ -610,6 +617,13 @@ func (w *ShelleyTransactionWitnessSet) UnmarshalCBOR(cborData []byte) error {
 	*w = ShelleyTransactionWitnessSet(tmp)
 	w.SetCbor(cborData)
 	return nil
+}
+
+func (w *ShelleyTransactionWitnessSet) MarshalCBOR() ([]byte, error) {
+	if w.Cbor() != nil {
+		return w.Cbor(), nil
+	}
+	return cbor.EncodeGeneric(w)
 }
 
 func (w ShelleyTransactionWitnessSet) Vkey() []common.VkeyWitness {
--- a/ledger/shelley/tx_test.go
+++ b/ledger/shelley/tx_test.go
@@ -15,11 +15,14 @@
 package shelley_test
 
 import (
+	"bytes"
 	"encoding/hex"
 	"math/big"
 	"reflect"
 	"testing"
 
+	"github.com/blinklabs-io/gouroboros/cbor"
+	"github.com/blinklabs-io/gouroboros/ledger/common"
 	"github.com/blinklabs-io/gouroboros/ledger/shelley"
 	"github.com/blinklabs-io/plutigo/data"
 )
@@ -46,6 +49,98 @@ func TestShelleyTransactionInputToPlutusData(t *testing.T) {
 			"did not get expected PlutusData\n     got: %#v\n  wanted: %#v",
 			tmpData,
 			expectedData,
+		)
+	}
+}
+
+// TestShelleyTransactionBody_MarshalCBOR_PreservesWireBytes reproduces
+// https://github.com/blinklabs-io/gouroboros/issues/1990: a decoded
+// ShelleyTransactionBody carrying a protocol parameter update must
+// re-marshal to its preserved wire bytes instead of falling through to
+// the generic encoder, which would emit an explicit CBOR null for each
+// of the 15 unset ShelleyProtocolParameterUpdate fields.
+func TestShelleyTransactionBody_MarshalCBOR_PreservesWireBytes(t *testing.T) {
+	genesisHash := common.NewBlake2b224(bytes.Repeat([]byte{0xAB}, 28))
+	// Minimal wire-format body: a fee, a TTL, and a protocol param
+	// update proposal with only MinFeeA (key 0) populated, as a real
+	// node would emit it.
+	rawBody := map[uint64]any{
+		2: uint64(206245),  // fee
+		3: uint64(5000000), // ttl
+		6: []any{
+			map[common.Blake2b224]any{
+				genesisHash: map[uint64]any{0: uint64(44)}, // MinFeeA only
+			},
+			uint64(4), // epoch
+		},
+	}
+	wireBytes, err := cbor.Encode(rawBody)
+	if err != nil {
+		t.Fatalf("unexpected error encoding wire bytes: %s", err)
+	}
+
+	body, err := shelley.NewShelleyTransactionBodyFromCbor(wireBytes)
+	if err != nil {
+		t.Fatalf("unexpected error decoding transaction body: %s", err)
+	}
+	if body.Update == nil || len(body.Update.ProtocolParamUpdates) != 1 {
+		t.Fatalf(
+			"expected a single protocol param update, got %#v",
+			body.Update,
+		)
+	}
+	update, ok := body.Update.ProtocolParamUpdates[genesisHash]
+	if !ok {
+		t.Fatalf("expected update entry for genesis hash %s", genesisHash)
+	}
+	if update.MinFeeA == nil || *update.MinFeeA != 44 {
+		t.Fatalf("expected MinFeeA == 44, got %#v", update.MinFeeA)
+	}
+	if update.MinFeeB != nil {
+		t.Fatalf("expected MinFeeB to remain unset, got %#v", update.MinFeeB)
+	}
+
+	marshaled, err := body.MarshalCBOR()
+	if err != nil {
+		t.Fatalf("unexpected error marshaling body: %s", err)
+	}
+	if !bytes.Equal(marshaled, wireBytes) {
+		t.Fatalf(
+			"MarshalCBOR() did not return preserved wire bytes: got %d bytes, wanted %d bytes",
+			len(marshaled),
+			len(wireBytes),
+		)
+	}
+
+	// Rebuilding a transaction from the decoded body and an empty
+	// witness set -- the same pattern ShelleyBlock.Transactions() uses
+	// -- must not inflate the encoded transaction size.
+	wsBytes, err := cbor.Encode(map[uint64]any{})
+	if err != nil {
+		t.Fatalf("unexpected error encoding witness set: %s", err)
+	}
+	var witnessSet shelley.ShelleyTransactionWitnessSet
+	if _, err := cbor.Decode(wsBytes, &witnessSet); err != nil {
+		t.Fatalf("unexpected error decoding witness set: %s", err)
+	}
+
+	wireTxBytes, err := cbor.Encode(
+		[]any{cbor.RawMessage(wireBytes), cbor.RawMessage(wsBytes), nil},
+	)
+	if err != nil {
+		t.Fatalf("unexpected error encoding wire transaction: %s", err)
+	}
+
+	rebuilt := &shelley.ShelleyTransaction{
+		Body:       *body,
+		WitnessSet: witnessSet,
+	}
+	rebuiltBytes := rebuilt.Cbor()
+	if !bytes.Equal(rebuiltBytes, wireTxBytes) {
+		t.Fatalf(
+			"rebuilt transaction CBOR does not match wire bytes: got %d bytes, wanted %d bytes",
+			len(rebuiltBytes),
+			len(wireTxBytes),
 		)
 	}
 }
--- a/ledger/shelley/tx_test.go
+++ b/ledger/shelley/tx_test.go
@@ -21,6 +21,9 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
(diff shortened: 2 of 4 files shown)


ledger/shelley/tx_test.go
```diff
@@ -49,3 +52,95 @@ func TestShelleyTransactionInputToPlutusData(t *testing.T) {
...
+		Body:       *body,
+		WitnessSet: witnessSet,
+	}
+	rebuiltBytes := rebuilt.Cbor()
+	if !bytes.Equal(rebuiltBytes, wireTxBytes) {
+		t.Fatalf(
+			"rebuilt transaction CBOR does not match wire bytes: got %d bytes, wanted %d bytes",
+			len(rebuiltBytes),
+			len(wireTxBytes),
+		)
```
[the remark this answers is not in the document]
cryptodj413: Addressed in d7bda8b: switched to `require`/`assert` from testify, matching convention used elsewhere in `ledger/shelley` tests.

ledger/shelley/tx_test.go
```diff
@@ -49,3 +55,78 @@ func TestShelleyTransactionInputToPlutusData(t *testing.T) {
...
+		t,
+		wireBytes,
+		marshaled,
+		"MarshalCBOR() did not return preserved wire bytes",
+	)
+
+	// Rebuilding a transaction from the decoded body and an empty
+	// witness set -- the same pattern ShelleyBlock.Transactions() uses
+	// -- must not inflate the encoded transaction size.
+	wsBytes, err := cbor.Encode(map[uint64]any{})
```
wolf31o2: This test still passes if both `AllegraTransactionBody.MarshalCBOR` and `ShelleyTransactionWitnessSet.MarshalCBOR` are removed: it never exercises an Allegra body, and canonical `a0` is reproduced by generic witness-set encoding. Add an Allegra reconstruction case and a witness-set fixture whose generic encoding differs from its wire bytes, then assert both reconstructed transactions preserve the original bytes.

wolf31o2 (changes requested):

ledger/shelley/tx_test.go
cryptodj413: Fixed in c4719f7:
- Added `TestAllegraTransactionBody_MarshalCBOR_PreservesWireBytes`, mirroring the Shelley body/reconstruction test for Allegra.
- Replaced the empty witness set with `newIndefLengthWitnessSetFixture`, whose VkeyWitnesses entry is wire-encoded as an indefinite-length CBOR array. Generic re-encoding of a decoded `[]common.VkeyWitness` always produces a definite-length array, so the fixture's generic encoding provably differs from its wire bytes (asserted directly in the fixture).

Verified both the new Allegra test and the updated Shelley test fail when `AllegraTransactionBody.MarshalCBOR`/`ShelleyTransactionWitnessSet.MarshalCBOR`'s preserved-bytes check is removed, and pass with it restored.

cryptodj413:
Addressed @wolf31o2's review feedback in c4719f7:
- Added `TestAllegraTransactionBody_MarshalCBOR_PreservesWireBytes`, exercising `AllegraTransactionBody.MarshalCBOR` directly.
- Replaced the empty witness-set fixture with one whose `VkeyWitnesses` entry is wire-encoded as an indefinite-length CBOR array, so its generic re-encoding provably differs from wire bytes, making `ShelleyTransactionWitnessSet.MarshalCBOR` load-bearing for both tests.

Verified both tests fail when the relevant `MarshalCBOR`'s preserved-bytes check is removed, and pass with it restored. CI green; re-review requested.

ledger/shelley/shelley.go
```diff
@@ -612,6 +619,13 @@ func (w *ShelleyTransactionWitnessSet) UnmarshalCBOR(cborData []byte) error {
 	return nil
 }
 
+func (w *ShelleyTransactionWitnessSet) MarshalCBOR() ([]byte, error) {
```
wolf31o2, line 622: Out of scope for #1990, recording it because this is where the class question gets settled.

Dijkstra has the identical defect, unfixed. `DijkstraTransaction.MarshalCBOR` (`ledger/dijkstra/dijkstra.go:1408`) ends in `return cbor.Encode([]any{t.Body, t.WitnessSet, aux})` - the same generic-encode fallback as Shelley and Allegra, not the `ReassembleTransactionCbor` path Alonzo, Babbage, and Conway take. And neither `DijkstraTransactionBody` nor `DijkstraTransactionWitnessSet` defines `MarshalCBOR`; the full inventory for that file has no body or witness-set entry.

Both types do store the wire bytes: the body embeds `common.TransactionBodyBase` and the witness set embeds `cbor.DecodeStoreCbor` directly, so `Cbor()` is populated on decode and simply never read back on marshal. That is precisely the shape this PR fixes two eras down.

Dijkstra is not on mainnet, so this is a follow-up rather than a merge concern - but it is the newest era and will inherit the bug this issue was opened for.

wolf31o2 (approved):
Approving. The previous finding is addressed at c4719f71, and each of the three methods is now load-bearing against a compiling failing run:

- `ShelleyTransactionBody.MarshalCBOR` - replacing its body with a bare `cbor.EncodeGeneric` fails `TestShelleyTransactionBody_MarshalCBOR_PreservesWireBytes`, with `ledger/allegra` still green.
- `AllegraTransactionBody.MarshalCBOR` - same substitution fails `TestAllegraTransactionBody_MarshalCBOR_PreservesWireBytes`, with `ledger/shelley` still green. This is the era case the old test never reached.
- `ShelleyTransactionWitnessSet.MarshalCBOR` - deleting it outright fails *both* era tests. The indefinite-length fixture is doing real work: generic re-encoding of a decoded `[]common.VkeyWitness` always emits a definite-length array, so the fixture's wire bytes cannot be reproduced by accident the way canonical `a0` was.

Note the substitution mattered for the two body methods: deleting them breaks the build, because the tests call them directly. Only neutering the preserved-bytes branch actually exercises the behavior.

**The class is complete for the eras that need it, and the fix reaches further than the title says.** The two preservation mechanisms in this package split cleanly by era:

| Era | Tx-level marshal | Covered by |
| --- | --- | --- |
| Shelley, Allegra, Mary | generic `cbor.Encode(tmpObj)` | body + witness-set `MarshalCBOR` |
| Alonzo, Babbage, Conway | `ReassembleTransactionCbor` from `Body.Cbor()`/`WitnessSet.Cbor()` | tx level |

Mary reaches the generic path too and holds `shelley.ShelleyTransactionWitnessSet` (`mary.go:326`), as Allegra does (`allegra.go:315`). Its body method already existed, but its witness set had none - so this change fixes Mary's witness-set preservation as well, without being named in the title or the issue.

Checks at c4719f71 in a clean worktree: `go test ./ledger/...` all green, `golangci-lint run` on both changed packages 0 issues, `gofmt -l ledger/` empty, `go vet ./ledger/...` clean.

One out-of-scope gap inline - not for this PR.

PR #2047, opened 2026-08-24
fix(ledger): validate exact reward withdrawals

### LINKED ISSUE #1998: ledger: validate exact reward-account withdrawal amounts
## Summary

The shared withdrawal validation rule checks reward-account registration but
does not compare each requested amount with the current account balance, even
though `RewardState` exposes `RewardAccountBalance`.

For protocol versions before PV12, a valid withdrawal must equal the full
reward balance. A registered zero-balance account may withdraw zero. Partial
and excessive amounts must be rejected. PV12 partial-withdrawal behavior must
remain separate.

## Acceptance criteria

- Exact full-balance withdrawals are accepted before PV12.
- Zero from a registered zero-balance account is accepted.
- Partial and excessive amounts return a typed validation error.
- Balance lookup failures are propagated.
- Existing unregistered-account and phase-2-invalid behavior is preserved.
- PV12 behavior is covered and not accidentally constrained by the pre-PV12 rule.
- Focused era tests and relevant ledger checks pass.

wolf31o2:
## Summary

- require exact reward withdrawals before Dijkstra and for Dijkstra transactions using Plutus V1-V3
- reject Dijkstra withdrawals above the current reward balance while permitting other partial withdrawals
- include subtransaction witnesses when resolving top-level scripts
- use the released `ouroboros-mock` v0.16.1 fixtures

## Validation

- `make test`
- `make lint`
- `go test -v -count=1 ./internal/test/conformance/...` (315/315)
- `go vet ./...`
- `go build ./...`

Fixes #1998


ledger/shelley/rules.go
```diff
@@ -636,6 +645,30 @@ func UtxoValidateWithdrawals(
 				RewardAddress: *addr,
 			}
 		}
+		if !requireExactAmount {
+			continue
+		}
+		balance, err := ls.RewardAccountBalance(cred)
+		if err != nil {
+			return err
```
chrisguiney, line 656: Blocking, though the problem is downstream rather than anything wrong with this code: this is the first production call to `RewardAccountBalance` anywhere in the rule set, and Dingo's implementation is still a stub.

```go
// dingo ledger/view.go:394
// TODO: implement reward account balance retrieval. Requires per-account reward
// balance tracking which is not yet stored in the database.
func (lv *LedgerView) RewardAccountBalance(cred lcommon.Credential) (*uint64, error) {
	return nil, ErrNotImplemented
}
```

The path is live: `ledger/state.go:4980` -> `ValidateTxConway` -> `conwayValidationRules` (`ledger/eras/conway.go:196`) -> `conway.UtxoValidateWithdrawals` -> here, with a `*LedgerView` as `ls`. On PV10/PV11 `requireExactAmount` is true, so this `return err` fires for *every* transaction containing any withdrawal, and `ValidateTxConway` joins it into a hard failure.

Dingo pins `gouroboros v0.193.3` today so nothing breaks this instant, but it lands the moment that pin moves. Could we get the balance tracking into Dingo first, or coordinate the two so the bump doesn't strand mainnet validation?

ledger/shelley/rules.go
```diff
@@ -636,6 +645,30 @@ func UtxoValidateWithdrawals(
 				RewardAddress: *addr,
 			}
 		}
+		if !requireExactAmount {
+			continue
+		}
```
chrisguiney: The Dijkstra path currently gets no amount check at all, which I think is more relaxed than intended. A quick probe against this branch:

- balance `1_000_000`, withdrawal `999_999_999_999`, PV12 -> accepted, `nil` error

Because `ledger/conway/rules.go:1817` and `ledger/shelley/rules.go:240` add the *declared* withdrawal amount to `consumedValue`, an over-withdrawal inflates the consumed side and the transaction balances against ADA that does not exist.

Upstream relaxes the rule rather than dropping it. From `eras/dijkstra/impl/src/Cardano/Ledger/Dijkstra/Rules/Entities.hs`:

```haskell
validateWithdrawals legacyMode network withdrawals accounts = do
  missingWithdrawals <-
    if legacyMode
      then do ... withdrawalsThatDoNotDrainAccounts ...
              failOnNonEmptyMap incompleteWithdrawals IncompleteWithdrawals
      else do ... withdrawalsThatExceedAccountBalance ...
              failOnNonEmptyMap exceededWithdrawals ExceededBalancesInWithdrawals
```

The Dijkstra changelog matches: *"Add `WithdrawalsExceedAccountBalance` to `DijkstraLedgerPredFailure`"*. So Dijkstra wants `amount <= balance`, not "no check" — which does mean fetching the balance on this path too.

ledger/shelley/rules.go
```diff
@@ -612,7 +612,8 @@ func UtxoValidateDelegation(
 }
 
 // UtxoValidateWithdrawals validates withdrawals against ledger state.
-// It checks that reward accounts are registered.
+// Before Dijkstra, every withdrawal must drain a registered reward account's
+// current balance exactly. Dijkstra permits partial withdrawals.
```
chrisguiney: Small correction to the doc, related to the comment below: upstream gates partial withdrawals on whether the transaction uses legacy Plutus, not on the era alone. From `eras/dijkstra/impl/src/Cardano/Ledger/Dijkstra.hs`:

```haskell
dsattPlutusLegacyMode = not $ Set.null $ Set.filter (<= PlutusV3) languagesUsed
```

`legacyMode = True` selects the exact-drain rule, so a PV12 transaction that uses any PlutusV1-V3 script still has to drain the account exactly. This repo already models `PlutusV4Script` and `TxSubTransactions`, so the primitives are here — but if implementing that is out of scope for this PR, could the comment read something like "Dijkstra permits partial withdrawals for transactions that use no legacy Plutus scripts" with a follow-up issue? As written it reads as settled, and the next person to touch this will trust it.

ledger/shelley/rules.go
```diff
@@ -636,6 +645,30 @@ func UtxoValidateWithdrawals(
...
+		}
+		balance, err := ls.RewardAccountBalance(cred)
+		if err != nil {
+			return err
+		}
+		if balance == nil {
+			return WithdrawalFromUnregisteredRewardAccountError{
+				RewardAddress: *addr,
+			}
+		}
```
chrisguiney, line 662: Non-blocking. This duplicates the `WithdrawalFromUnregisteredRewardAccountError` a few lines up, and it makes the `IsRewardAccountRegistered` call at line 643 redundant on this path — `RewardAccountBalance`'s documented contract (`ledger/common/state.go:81-94`) already says `(nil, nil)` means unregistered.

There is a real cost in Dingo, where `IsRewardAccountRegistered` is a `GetAccountByCredential` DB read (`ledger/view.go:364`): pre-Dijkstra now pays two state reads per withdrawal where one would do.

That said, keeping both is defensible if an implementation could hold a stale balance for a deregistered account. If that is the intent, a one-line comment saying so would stop the next reader from "simplifying" it away.

ledger/shelley/rules.go
```diff
@@ -624,7 +625,15 @@ func UtxoValidateWithdrawals(
...
 	}
 
-	for addr := range withdrawals {
+	requireExactAmount := true
+	if versionedPparams, ok := pp.(interface {
+		ProtocolMajorVersion() uint
+	}); ok {
+		requireExactAmount = versionedPparams.ProtocolMajorVersion() <
+			common.ProtocolVersionDijkstra
+	}
```
chrisguiney, line 646: Non-blocking: this is byte-for-byte the same anonymous-interface assertion as `ledger/conway/rules.go:3190-3196`. Might be worth pulling into a small helper in `common` — something like `ProtocolMajorVersionOf(pp) (uint, bool)` — so the two cannot drift apart later.

Defaulting to strict when the assertion fails is the right call either way, and `*DijkstraProtocolParameters` picks the method up through the embedded Conway params, so the gate does reach the Dijkstra rule set correctly.

ledger/shelley/rules.go
```diff
@@ -636,6 +645,30 @@ func UtxoValidateWithdrawals(
...
+		balance, err := ls.RewardAccountBalance(cred)
+		if err != nil {
+			return err
+		}
+		if balance == nil {
+			return WithdrawalFromUnregisteredRewardAccountError{
+				RewardAddress: *addr,
+			}
+		}
+		expected := new(big.Int).SetUint64(*balance)
```
chrisguiney: Minor, entirely optional: this heap-allocates a `big.Int` per withdrawal on the happy path. Since every `Withdrawals()` value originates from a `uint64`, the comparison can be done without allocating:

```go
if amount == nil || !amount.IsUint64() || amount.Uint64() != *balance {
```

`IsUint64()` also correctly rejects a negative, so the guard stays equivalent.

ledger/shelley/rules.go
```diff
@@ -624,7 +625,15 @@ func UtxoValidateWithdrawals(
...
-	for addr := range withdrawals {
+	requireExactAmount := true
+	if versionedPparams, ok := pp.(interface {
+		ProtocolMajorVersion() uint
+	}); ok {
+		requireExactAmount = versionedPparams.ProtocolMajorVersion() <
+			common.ProtocolVersionDijkstra
+	}
+
+	for addr, amount := range withdrawals {
```
chrisguiney, line 648: Not introduced here and no need to act in this PR — but now that this loop can return two different error types, it is worth noting that ranging a `map[*common.Address]*big.Int` makes *which* invalid withdrawal gets reported vary between runs. Same pattern as the DRep loop in `ledger/conway/rules.go`, so it is probably an issue rather than a change. Flagging in case you think it is worth filing.

ledger/shelley/rules_test.go
```diff
@@ -1006,36 +1018,89 @@ func TestUtxoValidateWithdrawals(t *testing.T) {
...
-		rewardAddr := makeRewardAddress(t, keyHash)
+	keyHash := common.Blake2b224Hash([]byte("test-stake-key"))
+	rewardAddr := makeRewardAddress(t, keyHash)
+
+	for _, tc := range []struct {
+		name       string
+		balance    uint64
+		withdrawal uint64
+		wantError  bool
+	}{
```
chrisguiney, line 1029: Coverage gap worth closing: every case in this table passes `pp = nil`, which only exercises the `ok == false` fallback in the type assertion. Nothing here asserts that a partial withdrawal is *rejected* at PV10/PV11 with a real `*conway.ConwayProtocolParameters`.

Since the version gate is the substance of the change, it would be good to pin both directions explicitly. The conway side covers "PV12 permits", but nothing covers "PV11 rejects".

go.sum
```diff
@@ -4,6 +4,8 @@ github.com/bits-and-blooms/bitset v1.24.4 h1:95H15Og1clikBrKr/DuzMXkQzECs1M6hhoG
 github.com/bits-and-blooms/bitset v1.24.4/go.mod h1:7hO7Gc7Pp1vODcmWvKMRA9BNmbv6a/7QIWpPxHddWR8=
 github.com/blinklabs-io/ouroboros-mock v0.16.0 h1:uAPoFE+WOXX9hct1Kb725gYS/8SKzATrwCF7XKOZ0Xs=
 github.com/blinklabs-io/ouroboros-mock v0.16.0/go.mod h1:ysGU8dMjG8ShMQco/aarR1xIRH7GPCIdhrnkM+J4IVA=
```
chrisguiney: Nit: `go mod tidy` drops both `v0.16.0` lines here. CI's `go test ./...` will not notice, but `make test` depends on `mod-tidy`, so it shows up as a dirty tree locally.

chrisguiney (changes requested):
Thanks for this — the pre-Dijkstra rule reads cleanly and the table test is a nice tightening of the old coverage.

I reproduced the validation claims locally against `8be4632`: `go build ./...`, `go vet ./...`, `golangci-lint run ./ledger/...` (0 issues), `go test ./ledger/...`, the `-race` run on both packages, and the conformance suite (315/315) all pass. The single `go test ./...` failure I saw was `pipeline.TestValidateStage_Eta0ProviderError`, a pre-existing timing flake that passes at `-count=5` and is unrelated to this change.

No concurrency concerns — nothing added introduces goroutines, locks, or shared mutable state.

Two things I'd like to work through before approving, both left inline on `ledger/shelley/rules.go`:

1. This becomes the first production caller of `RewardAccountBalance`, which is still a `not implemented` stub in Dingo — so every withdrawal transaction would fail phase-1 validation there once the pin moves.
2. The Dijkstra branch skips amount validation entirely. Upstream relaxes "drain exactly" to "must not exceed balance" rather than removing the bound, and it gates that on legacy-Plutus usage rather than on the era alone.

The rest of the inline notes are non-blocking — duplication, a doubled state read, an allocation, an untidy `go.sum`, and one coverage gap. Happy to be wrong on any of these, especially the Dijkstra reading.

ledger/shelley/rules.go
wolf31o2: Dingo #3384 backs RewardAccountBalance with persisted account rewards, including key/script credentials and inactive, missing, and zero balances.

ledger/shelley/rules.go
wolf31o2: 3595238 rejects Dijkstra withdrawals above the current balance.

ledger/shelley/rules.go
wolf31o2: 3595238 keeps exact-drain semantics when a needed top-level purpose resolves to Plutus V1-V3, including a script supplied by a subtransaction.

ledger/shelley/rules.go
wolf31o2: 3595238 uses RewardAccountBalance as the single registration and balance lookup.

ledger/shelley/rules.go
wolf31o2: Kept the local protocol-version assertion; script-mode resolution remains local to this rule.

ledger/shelley/rules.go
wolf31o2: 3595238 compares uint64 values on the valid path and only copies the big.Int for an error.

ledger/shelley/rules.go
wolf31o2: No change; deterministic multi-withdrawal error selection predates this fix.

ledger/shelley/rules_test.go
wolf31o2: 3595238 adds explicit PV11 rejection alongside PV12 acceptance.

go.sum
wolf31o2: 3595238 tidies the root and all example module sums.

wolf31o2:
3595238 enforces PV11 and legacy-Plutus exact drains, Dijkstra nonlegacy upper bounds, and top-level script resolution through subtransaction witnesses. Coverage includes key, native, Plutus V1-V4, PV11/PV12, subtransaction, exact, partial, and excessive cases. Dingo #3384 supplies the persisted reward balances. Full race tests, lint, vet, build, and 315/315 conformance vectors pass.

chrisguiney (approved):

PR #2049, opened 2026-08-24
fix(ledger): validate block witness counts

### LINKED ISSUE #2045: fix: reject mismatched block witness counts
## Summary

Block decoding must require equal transaction-body and witness-set counts.
Without that invariant, Shelley through Conway blocks with fewer witness sets
than bodies can reach `Transactions()` and panic on an out-of-range access.

## Acceptance criteria

- Reject mismatched transaction-body and witness-set counts during block
  decoding in every affected era from Shelley through Conway, including when
  body-hash validation is skipped.
- Defensively prevent partial or out-of-range transaction access for malformed
  in-memory blocks. Preserve the existing `Transactions()` API; because it has
  no error return, return an empty transaction slice for that defensive path.
- Keep Dijkstra unchanged unless an equivalent parallel-slice invariant is
  demonstrated there.
- Add per-era regression tests for missing and extra witness sets, matching
  counts, and the relevant validation configurations.

wolf31o2:
Closes #2045.

## Summary

- Reject transaction-body and witness-set count mismatches during unconditional block decoding across all eras.
- Prevent partial or out-of-range transaction access for malformed in-memory blocks.
- Preserve valid block decoding and the existing public API.

`ExtractAndSetTransactionCbor` performs the count check on every decode, including `SkipBodyHashValidation`. Because `Transactions()` has no error channel, malformed in-memory blocks return zero transactions; constructor paths reject malformed blocks during decoding.

## Summary by Cubic

Reject mismatched transaction-body and witness-set counts during CBOR decode across all eras. Previously a mismatch could decode and expose partial transactions; now constructors error on decode and era `Transactions()` methods return an empty slice for in-memory mismatches.

- Move the count check into `ExtractAndSetTransactionCbor` and remove it from `common.ValidateBlockBodyHash`.
- Run the check in both default and `SkipBodyHashValidation` paths and remove redundant work that added up to ~46% overhead.
- Keep guard clauses in Shelley, Allegra, Mary, Alonzo, Babbage, and Conway `Transactions()` implementations.
- Add tests in `ledger/witness_count_test.go`.

## Summary by CodeRabbit

- Prevent mismatched transaction body and witness-set counts during block parsing.
- Return a clear validation error for malformed blocks.
- Return an empty transaction list safely across supported ledger eras.
- Add coverage for missing, extra, and matching witness sets.


wolf31o2:
CodeRabbit was rate-limited and Cubic did not post a review for this head. Independent review found no actionable findings. Focused per-era tests, the full test suite, lint, build, conformance, and nilaway passed locally; CI is green at `c9fda3478c45576a137eccecffcef9c2369929f6`.

ledger/common/verify_config.go
```diff
@@ -203,6 +205,9 @@ func ValidateBlockBodyHash(
 			nil,
 		)
 	}
+	if err := validateBlockTransactionWitnessCounts(raw, eraName); err != nil {
```
chrisguiney: This layer can't fire in practice, and it costs a fair amount on the default path, so it seems worth reconsidering before merge.

**It's unreachable.** `ValidateBlockBodyHash` runs *after* `cbor.Decode`, which already ran `UnmarshalCBOR` -> `ExtractAndSetTransactionCbor` -> rejected the mismatch. Probed with body-hash validation **enabled** (the production default):

```
validation ENABLED -> decode Conway block error: transaction body and witness set counts do not match: 2 bodies, 1 witness sets
validation SKIPPED -> (identical)
```

The `ValidationError` built below is never produced through any constructor. `VerifyBlock` (`verify_block.go:731`) is the same story — it receives an already-decoded block.

**It isn't load-bearing.** Removing this call and running the entire suite breaks only `TestValidateBlockBodyHashTransactionWitnessCounts`. The constructor tests in `ledger/witness_count_test.go` still pass, via the cheap check in `ExtractAndSetTransactionCbor`.

**It's expensive.** `cbor.Decode(raw[1], &transactionBodies)` copies every transaction body and every witness set. Benchmarking `ValidateBlockBodyHash` on Conway-shaped blocks, main vs this branch:

| txs | block size | main | branch | delta | main allocs | branch allocs |
|---|---|---|---|---|---|---|
| 1 | 2.5 KB | 7517 ns | 7953 ns | +5.8% | 16 | 31 |
| 10 | 19.7 KB | 27105 ns | 32648 ns | +20.5% | 19 | 58 |
| 50 | 95.9 KB | 93307 ns | 136481 ns | +46% | 21 | 145 |

362 KB/op -> 785 KB/op at 50 txs, on every block that isn't `SkipBodyHashValidation`.

Two ways out, whichever you prefer:

**(a) Drop this call.** `ExtractAndSetTransactionCbor` covers every decode path unconditionally and the `Transactions()` guards cover in-memory blocks, so no coverage is lost. The test can point at a constructor instead.

**(b) Keep it as defense-in-depth for direct callers of the exported function, but count with `cborArrayInfo`** (already in this package) rather than decoding. I tried this locally: the existing test still passes and the benchmark returns to main's numbers exactly (93329 ns/op, 21 allocs at 50 txs). One caveat if you go this way — `cborArrayInfo` reports indefinite-length arrays as count 0, so it would need the same indefinite fallback `setArrayItemCbor` already models, or the check quietly no-ops on them.

ledger/common/verify_config.go
```diff
@@ -226,6 +231,59 @@ func ValidateBlockBodyHash(
 	return nil
 }
 
+func validateBlockTransactionWitnessCounts(
+	raw []cbor.RawMessage,
+	eraName string,
+) error {
+	if len(raw) < 3 {
```
chrisguiney: Minor: this is unreachable. `ValidateBlockBodyHash` already returned above if `len(raw) < minRawLength`, and both call sites pass 4 or 5.

ledger/witness_count_test.go
```diff
@@ -0,0 +1,324 @@
...
+			constructor: func(
+				data []byte,
+				config ...common.VerifyConfig,
+			) (common.Block, error) {
+				return NewConwayBlockFromCbor(data, config...)
+			},
+			minRawLength: 5,
+		},
+	}
+	config := common.VerifyConfig{SkipBodyHashValidation: true}
```
chrisguiney: Small coverage gap: this table only exercises `SkipBodyHashValidation: true`, so the default configuration — the one production uses — isn't covered at the constructor level. A default-config case per era would be cheap, and it would also make visible that both configs reject through the same layer (see the note on `ledger/common/verify_config.go`).

ledger/shelley/shelley.go
```diff
@@ -151,9 +151,12 @@ func (b *ShelleyBlock) Era() common.Era {
 }
 
 func (b *ShelleyBlock) Transactions() []common.Transaction {
+	if len(b.TransactionBodies) != len(b.TransactionWitnessSets) {
+		return []common.Transaction{}
+	}
 	ret := make([]common.Transaction, len(b.TransactionBodies))
 	// #nosec G115
-	for idx := range b.TransactionBodies {
+	for idx := range ret {
```
chrisguiney: `ret` is made with `len(b.TransactionBodies)`, so `range ret` and `range b.TransactionBodies` are equivalent here — this looks like a no-op in all six era files, with the early return above providing the actual safety.

Happy to defer if it's deliberate signalling; just flagging that six files are touched for no behavior change.

ledger/shelley/shelley.go
```diff
@@ -151,9 +151,12 @@ func (b *ShelleyBlock) Era() common.Era {
 }
 
 func (b *ShelleyBlock) Transactions() []common.Transaction {
+	if len(b.TransactionBodies) != len(b.TransactionWitnessSets) {
```
chrisguiney, line 154: Agreed this is the right call given `Transactions()` has no error channel — #2045 asked for an error, and an empty slice is the closest the signature allows.

Might be worth a line in the PR description, though: a malformed *in-memory* block now reports zero transactions silently. It's mostly theoretical for dingo, since stored blocks get re-decoded through these constructors (`database/block_indexer.go`, `ledger/replay_recovery.go`) and would hit the decode error first — but a consumer like `ledger/state.go` applying an empty tx list would advance the chain rather than surface the corruption.

ledger/witness_count_test.go
```diff
@@ -0,0 +1,324 @@
...
+						err,
+						"transaction body and witness set counts do not match",
+					)
+				})
+			}
+		})
+	}
+}
+
+func blockCborWithTransactionCounts(
```
chrisguiney, line 312: Nit: this is ~95% identical to `blockWithTransactionCounts` in `ledger/common/verify_config_test.go` (only the returned hash differs). The package split makes sharing awkward, so this may just be the cost of testing both layers — flagging in case an internal test helper appeals.

ledger/common/common.go
```diff
@@ -860,6 +860,15 @@ func ExtractAndSetTransactionCbor(
 	expectedBodies int,
 	expectedWitnesses int,
 ) error {
+	if expectedBodies != expectedWitnesses {
+		return fmt.Errorf(
+			"transaction body and witness set counts do not match: "+
+				"%d bodies, %d witness sets",
```
chrisguiney: Tiny nit: always plural, so a single mismatch reads "1 witness sets".

chrisguiney (changes requested):
Thanks for this — the invariant is real and the main fix lands in the right place. Verified locally against `c9fda347`: both new test files fail on `origin/main`, and `Shelley/missing_witness` reproduces the exact crash from #2045 (`index out of range [1] with length 1`). `extra_witness` doesn't panic on main but silently drops a witness set, so that's a good catch too.

Also checked: real block fixtures for all six affected eras still decode cleanly through the new guard; Dijkstra and Byron are correctly left alone (single transaction slice, no parallel-slice invariant, and neither reaches `ValidateBlockBodyHash`); `Transactions()` is the only place a shared index reaches into both slices. Full suite, `-race` on `./ledger/...`, `go vet`, and `gofmt` are green here as well as in CI, and the branch merges cleanly onto current main.

One blocking item: the `ValidateBlockBodyHash` layer is provably unreachable and measurably expensive on the default validation path — details inline on `ledger/common/verify_config.go`. Everything else below is non-blocking.

Requesting changes for that one item only — happy to re-review as soon as it's sorted.

ledger/common/verify_config.go
wolf31o2: Removed in c978350 — took option (a). Reproduced the unreachability independently: a 2-body/1-witness block through `NewShelleyBlockFromCbor` and `NewConwayBlockFromCbor` returns `transaction body and witness set counts do not match: 2 bodies, 1 witness sets` from `ExtractAndSetTransactionCbor` under both the default config and `SkipBodyHashValidation: true`, so the `ValidationError` never surfaced.

Chose (a) over (b) because `ExtractAndSetTransactionCbor` runs on every decode including when `SkipBodyHashValidation` is set, which the removed layer never covered — so the cheap check is strictly broader, not just cheaper. Option (b) would also have needed the indefinite-length stream walk `setArrayItemCbor` models, and a guard that silently no-ops on indefinite arrays is worse than no guard. There are no direct callers of the exported function in the workspace; the two `dingo` hits are comments.

`ledger/common/verify_config.go` and `verify_config_test.go` are now byte-identical to `main`, so the benchmark delta is zero by construction.

ledger/common/verify_config.go
wolf31o2: Resolved by the same change — `validateBlockTransactionWitnessCounts` and its `len(raw) < 3` guard are gone in c978350.

ledger/witness_count_test.go
wolf31o2: Fixed in c978350. The mismatch cases now run under both the default config and `SkipBodyHashValidation: true` via a `configs` table, which makes the shared rejection layer visible. The positive case still uses the skipping config only, because these synthetic blocks carry no header or real body hash — noted in a comment above the table. Renamed the test to `TestBlockConstructorsRejectMismatchedWitnessCounts`, since the old name claimed narrower coverage than it now has.

ledger/shelley/shelley.go
wolf31o2: Not deliberate — reverted to `range b.TransactionBodies` in all six era files in c978350. Those files now carry only the three-line guard.

ledger/shelley/shelley.go
wolf31o2: Added to the PR description.

ledger/witness_count_test.go
wolf31o2: Resolved as a side effect: `blockWithTransactionCounts` and `TestValidateBlockBodyHashTransactionWitnessCounts` were removed with the layer they covered, so only `blockCborWithTransactionCounts` in `ledger/witness_count_test.go` remains. No shared helper needed.

ledger/common/common.go
wolf31o2: Fixed in c978350 — now `body count %d, witness set count %d`, which reads correctly at 1.

wolf31o2:
All seven items addressed in c978350; branch merged up to current `main`.

The blocking one: took option (a). The unreachability reproduces independently — a 2-body/1-witness block through `NewShelleyBlockFromCbor` and `NewConwayBlockFromCbor` is rejected by `ExtractAndSetTransactionCbor` under both the default config and `SkipBodyHashValidation: true`, so the `ValidationError` never surfaced. Option (a) over (b) because the cheap check is strictly broader, not just cheaper: it runs even when body-hash validation is skipped, which the removed layer never covered. `ledger/common/verify_config.go` and `verify_config_test.go` are now byte-identical to `main`, so the benchmark delta is zero by construction.

Also: reverted the no-op `range` change in all six era files, added default-config cases at the constructor level, renamed that test since it no longer only covers the skipped path, and fixed the `1 witness sets` grammar.

Checks: `go vet`, `gofmt`, and `golangci-lint` clean on `./ledger/...`; `go test -race ./ledger/...` green. Removing the `ExtractAndSetTransactionCbor` guard makes the constructor tests fail — the skipped-validation cases return no error at all — confirming the remaining check is what holds the invariant.

chrisguiney (approved):

Issue #2048, opened 2026-08-24
protocol: keep optional responders stable when disabled


wolf31o2:
## Summary

Optional node-to-node server responders should remain protocol-safe when their
feature is disabled or no application callback is configured. Peer sharing,
Leios fetch, and Leios votes currently use inconsistent fallback behavior,
while Leios notify already keeps an unconfigured request non-fatal.

## Scope

- Define a state-machine-valid fallback for disabled peer-sharing requests.
- Define non-fatal fallbacks for unconfigured Leios fetch request types.
- Keep an unconfigured Leios votes responder stable while it has server agency.
- Preserve error propagation for configured callback and transport failures.
- Document the fallback behavior.

## Acceptance criteria

- Disabled peer sharing completes the request without invoking its callback.
- Unconfigured Leios fetch and Leios votes handlers do not emit a connection-level
  protocol error.
- Each fallback respects the mini-protocol's valid state transitions and wire
  messages.
- Regression coverage exercises the handlers through the protocol/muxer path and
  proves the shared connection remains usable.

PR #2052, opened 2026-08-24
fix(protocol): keep optional responders non-fatal

wolf31o2:
Relates to #2048.

## Summary

- return valid empty arrays for disabled or unconfigured peer sharing and unconfigured Leios fetch vote requests
- preserve shared-bearer liveness for those safe fallbacks while retaining observable errors where the wire protocol has no valid absence message
- keep the Leios votes busy timeout bounded and reject unconfigured requests instead of stalling the shared bearer
- document the exact fallback and error behavior

Issue #2048 remains open because Leios votes and Leios fetch block, block-transaction, and block-range requests have no safe absence messages. Making those paths non-fatal requires a protocol or registration design rather than placeholder or null payloads.

## Validation

- `make test`
- `make lint`
- `go build ./...`
- `go vet ./protocol/leiosfetch ./protocol/leiosvotes ./protocol/peersharing`
- `go test -race -count=1 ./protocol/leiosfetch ./protocol/leiosvotes ./protocol/peersharing`

### DIFF
--- a/protocol/leiosfetch/README.md
+++ b/protocol/leiosfetch/README.md
@@ -188,13 +188,13 @@ whole node-to-node connection.
 
 ## Optional server responders
 
-Leios fetch is optional, but its server responder is available on every
-node-to-node connection that supports the mini-protocol. If a callback is not
-configured, the server declines each request with a state-machine-valid empty
-response: `NoBlock`, `NoBlockTxs`, an empty `Votes`, or an empty
-`LastBlockAndTxsInRange`, respectively. These responses return the protocol to
-`Idle` without producing a connection-level error. Errors from configured
-callbacks and transport failures are still propagated.
+Leios fetch is optional, but not every request has a safe empty response. An
+unconfigured `VotesRequestFunc` returns `Votes` with an empty CBOR array,
+returning the protocol to `Idle` without a connection-level error. The block,
+block-transactions, and block-range requests require their callbacks: their
+absence replies are either placeholder wire IDs or ambiguous with a real
+response. Errors from configured callbacks and transport failures are still
+propagated.
 
 ## Notes
 
--- a/protocol/leiosfetch/README.md
+++ b/protocol/leiosfetch/README.md
@@ -186,6 +186,16 @@ whole node-to-node connection.
   error to the caller, who can distinguish "not available" from a real protocol
   error with `errors.Is`.
 
+## Optional server responders
+
+Leios fetch is optional, but its server responder is available on every
+node-to-node connection that supports the mini-protocol. If a callback is not
+configured, the server declines each request with a state-machine-valid empty
+response: `NoBlock`, `NoBlockTxs`, an empty `Votes`, or an empty
+`LastBlockAndTxsInRange`, respectively. These responses return the protocol to
+`Idle` without producing a connection-level error. Errors from configured
+callbacks and transport failures are still propagated.
+
 ## Notes
 
 - Part of the experimental Leios protocol suite
--- a/protocol/leiosfetch/server.go
+++ b/protocol/leiosfetch/server.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/blinklabs-io/gouroboros/cbor"
 	"github.com/blinklabs-io/gouroboros/protocol"
 )
 
@@ -102,10 +103,9 @@ func (s *Server) handleBlockRequest(msg protocol.Message) error {
 			"connection_id", s.callbackContext.ConnectionId.String(),
 		)
 	if s.config == nil || s.config.BlockRequestFunc == nil {
-		// Leios is optional and its server responder is registered for every
-		// node-to-node connection. An unconfigured responder declines the
-		// request without treating it as a bearer-level protocol error.
-		return s.SendMessage(NewMsgNoBlock())
+		return errors.New(
+			"received leios-fetch BlockRequest message but no callback function is defined",
+		)
 	}
 	msgBlockRequest := msg.(*MsgBlockRequest)
 	resp, err := s.config.BlockRequestFunc(
@@ -152,7 +152,9 @@ func (s *Server) handleBlockTxsRequest(msg protocol.Message) error {
 			"connection_id", s.callbackContext.ConnectionId.String(),
 		)
 	if s.config == nil || s.config.BlockTxsRequestFunc == nil {
-		return s.SendMessage(NewMsgNoBlockTxs())
+		return errors.New(
+			"received leios-fetch BlockTxsRequest message but no callback function is defined",
+		)
 	}
 	msgBlockTxsRequest := msg.(*MsgBlockTxsRequest)
 	resp, err := s.config.BlockTxsRequestFunc(
@@ -200,7 +202,7 @@ func (s *Server) handleVotesRequest(msg protocol.Message) error {
 			"connection_id", s.callbackContext.ConnectionId.String(),
 		)
 	if s.config == nil || s.config.VotesRequestFunc == nil {
-		return s.SendMessage(NewMsgVotes(nil))
+		return s.SendMessage(NewMsgVotes([]cbor.RawMessage{}))
 	}
 	msgVotesRequest := msg.(*MsgVotesRequest)
 	resp, err := s.config.VotesRequestFunc(
@@ -230,7 +232,9 @@ func (s *Server) handleBlockRangeRequest(msg protocol.Message) error {
 			"connection_id", s.callbackContext.ConnectionId.String(),
 		)
 	if s.config == nil || s.config.BlockRangeRequestFunc == nil {
-		return s.SendMessage(NewMsgLastBlockAndTxsInRange(nil, nil))
+		return errors.New(
+			"received leios-fetch BlockRangeRequest message but no callback function is defined",
+		)
 	}
 	msgBlockRangeRequest := msg.(*MsgBlockRangeRequest)
 	err := s.config.BlockRangeRequestFunc(
--- a/protocol/leiosfetch/server.go
+++ b/protocol/leiosfetch/server.go
@@ -102,9 +102,10 @@ func (s *Server) handleBlockRequest(msg protocol.Message) error {
 			"connection_id", s.callbackContext.ConnectionId.String(),
 		)
 	if s.config == nil || s.config.BlockRequestFunc == nil {
-		return errors.New(
-			"received leios-fetch BlockRequest message but no callback function is defined",
-		)
+		// Leios is optional and its server responder is registered for every
+		// node-to-node connection. An unconfigured responder declines the
+		// request without treating it as a bearer-level protocol error.
+		return s.SendMessage(NewMsgNoBlock())
 	}
 	msgBlockRequest := msg.(*MsgBlockRequest)
 	resp, err := s.config.BlockRequestFunc(
@@ -151,9 +152,7 @@ func (s *Server) handleBlockTxsRequest(msg protocol.Message) error {
 			"connection_id", s.callbackContext.ConnectionId.String(),
 		)
 	if s.config == nil || s.config.BlockTxsRequestFunc == nil {
-		return errors.New(
-			"received leios-fetch BlockTxsRequest message but no callback function is defined",
-		)
+		return s.SendMessage(NewMsgNoBlockTxs())
 	}
 	msgBlockTxsRequest := msg.(*MsgBlockTxsRequest)
 	resp, err := s.config.BlockTxsRequestFunc(
@@ -201,9 +200,7 @@ func (s *Server) handleVotesRequest(msg protocol.Message) error {
 			"connection_id", s.callbackContext.ConnectionId.String(),
 		)
 	if s.config == nil || s.config.VotesRequestFunc == nil {
-		return errors.New(
-			"received leios-fetch VotesRequest message but no callback function is defined",
-		)
+		return s.SendMessage(NewMsgVotes(nil))
 	}
 	msgVotesRequest := msg.(*MsgVotesRequest)
 	resp, err := s.config.VotesRequestFunc(
@@ -233,9 +230,7 @@ func (s *Server) handleBlockRangeRequest(msg protocol.Message) error {
 			"connection_id", s.callbackContext.ConnectionId.String(),
 		)
 	if s.config == nil || s.config.BlockRangeRequestFunc == nil {
-		return errors.New(
-			"received leios-fetch BlockRangeRequest message but no callback function is defined",
-		)
+		return s.SendMessage(NewMsgLastBlockAndTxsInRange(nil, nil))
 	}
 	msgBlockRangeRequest := msg.(*MsgBlockRangeRequest)
 	err := s.config.BlockRangeRequestFunc(
--- a/protocol/leiosfetch/server_test.go
+++ b/protocol/leiosfetch/server_test.go
@@ -29,9 +29,17 @@ import (
 	"github.com/blinklabs-io/gouroboros/muxer"
 	"github.com/blinklabs-io/gouroboros/protocol"
 	pcommon "github.com/blinklabs-io/gouroboros/protocol/common"
+	"github.com/blinklabs-io/gouroboros/protocol/peersharing"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+func testLeiosFetchConnectionId() connection.ConnectionId {
+	return connection.ConnectionId{
+		LocalAddr:  &net.TCPAddr{IP: net.IPv4(127, 0, 0, 1), Port: 0},
+		RemoteAddr: &net.TCPAddr{IP: net.IPv4(127, 0, 0, 1), Port: 0},
+	}
+}
 
 func readLeiosFetchTestSegment(
 	t *testing.T,
@@ -75,7 +83,7 @@ func sendNotFoundTest(
 	t *testing.T,
 	cfg Config,
 	request protocol.Message,
-) uint {
+) (uint, []byte) {
 	t.Helper()
 	connId := connection.ConnectionId{
 		LocalAddr:  &net.TCPAddr{IP: net.IPv4(127, 0, 0, 1), Port: 0},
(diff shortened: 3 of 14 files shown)


protocol/leiosfetch/server_test.go
```diff
@@ -170,42 +172,22 @@ func TestHandleBlockRequest_CallbackIsCalled(t *testing.T) {
...
-	err := server.handleBlockRequest(msg)
-
-	assert.Error(t, err)
-	assert.Contains(t, err.Error(), "no callback function is defined")
+	msgType := sendNotFoundTest(
+		t,
+		cfg,
+		NewMsgBlockRequest(pcommon.NewPoint(12345, []byte{0x01, 0x02})),
+	)
+	assert.Equal(t, uint(MessageTypeNoBlock), msgType)
```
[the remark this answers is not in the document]
wolf31o2: Updated the five terminal assertions to use require.Equal in e72d541.

protocol/peersharing/README.md
```diff
@@ -78,7 +78,7 @@ disables peer sharing is `NoPeerSharing`.
 | Local advertised | Remote advertised | Client may send `ShareRequest`? | Server honours incoming `ShareRequest`? |
 |---|---|---|---|
 | any | NoPeerSharing | no — `ErrRemotePeerSharingDisabled` | yes (if local advertised active) |
-| NoPeerSharing | active | no — would be a self-protocol violation; client should not be calling `GetPeers` | no — `ErrLocalPeerSharingDisabled` |
+| NoPeerSharing | active | no — would be a self-protocol violation; client should not be calling `GetPeers` | empty `SharePeers` response |
```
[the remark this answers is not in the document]
wolf31o2: Updated the handshake overview to distinguish client-side negotiation gating from the server-side empty fallback in e72d541.

protocol/leiosvotes/server.go
```diff
@@ -114,9 +119,10 @@ func (s *Server) handleRequestNext(msg protocol.Message) error {
 		)
 	}
 	if s.config == nil || s.config.RequestNextFunc == nil {
-		return errors.New(
-			"received leios-votes VotesRequestNext message but no callback function is defined",
-		)
+		// Leios vote diffusion is optional and has no empty response message.
+		// Keep the request in Busy (with no timeout) so an unconfigured
+		// responder does not turn an optional request into a bearer error.
+		return nil
```
cryptodj413: **Merge blocker.** `StateBusy` (leiosvotes.go) has `Agency: AgencyServer` and only `MessageTypeVote` transitions — not even `MsgDone`. Returning `nil` here without sending anything leaves the protocol in `Busy` with no legal way out, and (per `initProtocol` above) with no timeout at all when unconfigured, which is the default path per `connection.go`.

Because the state never transitions, `recvReadyChan` is never re-signaled and `recvLoop` stops consuming (`protocol/protocol.go`). `readLoop` keeps draining the muxer into `recvQueueChan` (cap 55) until it fills, then blocks on the muxer's per-protocol send channel (cap 10) — and since the muxer has one `readLoop`, that stalls every other mini-protocol sharing the connection (chainsync/blockfetch/txsubmission included). Reproduced against a two-protocol muxer topology built from this PR's own test helpers: after 67 extra segments (55+10+in-flight) the write blocks and never recovers.

Before this PR, the first unconfigured request produced a clean `SendError` and closed the connection — diagnosable. This trades that for a silent, permanent, remotely-triggerable stall with no log, metric, or timeout.

leios-notify isn't a valid precedent: its `StateBusy` has no timeout in the *base* state map because a notify server legitimately holds agency until it has something to announce. leios-votes deliberately installs a 60s timeout and its own client expects exactly `count` votes back, so "pending forever" is a new state, not an existing one.

Suggest: keep a bounded `Busy` timeout even when unconfigured (restores the old, diagnosable close), or stop registering/advertising the leios-votes responder at all when unconfigured so the request is rejected at the muxer instead of buffered indefinitely.

protocol/peersharing/server.go
```diff
@@ -101,9 +103,9 @@ func (s *Server) handleShareRequest(msg protocol.Message) error {
 			"connection_id", s.callbackContext.ConnectionId.String(),
 		)
 	if s.config == nil || s.config.ShareRequestFunc == nil {
-		return errors.New(
-			"received peer-sharing ShareRequest message but no callback function is defined",
-		)
+		// Peer sharing is optional. Keep the protocol in its normal request /
+		// response cycle when the application has no peer source configured.
+		return s.SendMessage(NewMsgSharePeers(nil))
```
cryptodj413: **Merge blocker.** `NewMsgSharePeers(nil)` sets `PeerAddresses: nil`, which CBOR-encodes as `[1, null]` (`8201f6`), not `[1, []]` (`820180`) — confirmed by encoding both and diffing against this package's own wire fixture in `messages_test.go` (`[1, [...]]`). `[1, null]` isn't a member of `[ *peerAddress ]`; gouroboros' own decoder happens to accept `null` as a nil slice, which is why the new tests (which decode before asserting) pass on the wrong bytes.

Unlike leios (experimental), peer sharing is spoken to real Haskell `cardano-node` peers. If its decoder is stricter about `null` vs an empty array here, this reintroduces the exact failure mode (bearer reset) the PR is trying to prevent — couldn't confirm against a real node since `internal/test/peersharing-interop` needs Docker + two synced preview nodes (unavailable here), but the deviation from the documented/fixture shape is confirmed.

Fix: `NewMsgSharePeers([]PeerAddress{})` here and at line 96 — verified to produce `820180`.

protocol/leiosfetch/server.go
```diff
@@ -233,9 +230,7 @@ func (s *Server) handleBlockRangeRequest(msg protocol.Message) error {
 			"connection_id", s.callbackContext.ConnectionId.String(),
 		)
 	if s.config == nil || s.config.BlockRangeRequestFunc == nil {
-		return errors.New(
-			"received leios-fetch BlockRangeRequest message but no callback function is defined",
-		)
+		return s.SendMessage(NewMsgLastBlockAndTxsInRange(nil, nil))
```
cryptodj413: **Merge blocker.** `NewMsgLastBlockAndTxsInRange(nil, nil)` sets `BlockRaw: nil` (a `cbor.RawMessage`), which encodes as `[7, null, null]`. Unlike `BlockRequest`/`BlockTxsRequest`, this sub-protocol has no `NoBlock`/`NoBlockTxs`-style sentinel for "nothing in range" — there's no empty representation for `BlockRaw` even with an empty slice, since the field itself goes to `null`.

`Client.BlockRangeRequest` will hand its caller a real-looking `MsgLastBlockAndTxsInRange` whose block field is the single byte `f6`, with no way to distinguish "empty range" from "here is a block." This is the same gap the PR correctly identifies and works around for leios-votes ("no empty-batch response in this wire protocol") but doesn't address here.

Either add a real "no data" message for this sub-protocol (or a documented sentinel `BlockRaw` value the client checks for), or leave this path erroring until one exists — silently fabricating a null block is worse than the current error.

protocol/leiosfetch/server.go
```diff
@@ -201,9 +200,7 @@ func (s *Server) handleVotesRequest(msg protocol.Message) error {
 			"connection_id", s.callbackContext.ConnectionId.String(),
 		)
 	if s.config == nil || s.config.VotesRequestFunc == nil {
-		return errors.New(
-			"received leios-fetch VotesRequest message but no callback function is defined",
-		)
+		return s.SendMessage(NewMsgVotes(nil))
```
cryptodj413: Same class as the `SharePeers`/`LastBlockAndTxsInRange` issue: `NewMsgVotes(nil)` encodes as `[5, null]`, not `[5, []]`. Lower severity since leios is experimental, but a one-line fix: `NewMsgVotes([]cbor.RawMessage{})`.

protocol/peersharing/server.go
```diff
@@ -91,7 +90,10 @@ func (s *Server) handleMessage(msg protocol.Message) error {
 
 func (s *Server) handleShareRequest(msg protocol.Message) error {
 	if s.config != nil && s.config.LocalDisabled {
-		return ErrLocalPeerSharingDisabled
+		// A peer that sends ShareRequest after we advertised NoPeerSharing is
+		// not following the handshake negotiation. Still answer with an empty
+		// response so an invalid optional request cannot tear down the bearer.
+		return s.SendMessage(NewMsgSharePeers(nil))
```
cryptodj413: This `LocalDisabled` branch returns before the `Debug("share request", ...)` log below, so a peer that ignores the `NoPeerSharing` handshake negotiation now produces no error, no log, and no metric — previously it surfaced `ErrLocalPeerSharingDisabled` on the error channel. Worth at least a `Debug`/`Warn` log here so operators keep visibility into peers violating the negotiation.

protocol/peersharing/peersharing.go
```diff
@@ -30,9 +30,10 @@ var ErrRemotePeerSharingDisabled = errors.New(
...
 )
 
-// ErrLocalPeerSharingDisabled is returned by the server when a peer sends a
-// ShareRequest but we advertised NoPeerSharing during the handshake. A
-// spec-compliant peer must not send ShareRequest in that case.
+// ErrLocalPeerSharingDisabled is retained for compatibility with callers that
+// classified the previous server-side behavior. The server now answers an
+// unexpected ShareRequest with an empty SharePeers message so it cannot tear
+// down the shared bearer.
 var ErrLocalPeerSharingDisabled = errors.New(
```
cryptodj413, line 38: With the `LocalDisabled` branch in server.go no longer returning this error, `ErrLocalPeerSharingDisabled` has no remaining caller in the repo (checked). Since no caller can receive it again, consider marking it `// Deprecated:` with a pointer to the new behavior, or removing it.

protocol/leiosfetch/server_test.go
```diff
@@ -95,26 +95,28 @@ func sendNotFoundTest(
 	defer server.Protocol.Stop()
 	m.Start()
```
cryptodj413, line 123: The `protocol.ProtocolOptions{ConnectionId: connId, Muxer: m}` this helper builds a few lines above has no `ErrorChan`. `Protocol.SendError` selects on a nil channel with a `default:` branch, so any protocol-level error here is silently dropped rather than asserted absent — these tests only prove a response segment arrived, not that no connection-level error was raised, which is the issue's actual acceptance criterion ("do not emit a connection-level protocol error"). `leiosvotes/server_test.go`'s `TestUnconfiguredRequestDoesNotTearDownBearer` does this correctly with a buffered `ErrorChan` + `require.Never`; worth mirroring that pattern here.

protocol/leiosfetch/README.md
```diff
@@ -186,6 +186,16 @@ whole node-to-node connection.
   error to the caller, who can distinguish "not available" from a real protocol
   error with `errors.Is`.
 
+## Optional server responders
+
+Leios fetch is optional, but its server responder is available on every
+node-to-node connection that supports the mini-protocol. If a callback is not
+configured, the server declines each request with a state-machine-valid empty
```
cryptodj413: "empty `Votes`" and "empty `LastBlockAndTxsInRange`" aren't accurate as written: both currently encode with `null` fields (`[5, null]` and `[7, null, null]`), not an empty array/absent value. Once the code-side fixes above land this can describe the corrected wire shape.

protocol/peersharing/README.md
```diff
@@ -78,7 +79,7 @@ disables peer sharing is `NoPeerSharing`.
...
```
cryptodj413, line 83: Same doc/code mismatch as above: the "empty `SharePeers` response" is currently `[1, null]` on the wire, not `[1, []]`. Update once `NewMsgSharePeers(nil)` → `NewMsgSharePeers([]PeerAddress{})` is fixed.

protocol/leiosvotes/README.md
```diff
@@ -44,6 +44,17 @@ Busy --Vote, tokens = 1--> Idle
 Idle --Done--> Done
 ```
 
+## Optional server responder
+
+Leios votes is optional, but its server responder is available on every
+node-to-node connection that supports the mini-protocol. If `RequestNextFunc`
+is not configured, a request is accepted and remains in the server's `Busy`
```
cryptodj413: Locally accurate but incomplete: it doesn't mention that (a) the protocol can never be used again on that connection once it enters this state, (b) a peer that keeps sending on the pending protocol can wedge the shared bearer (see the server.go review comment), and (c) a gouroboros *client* talking to an unconfigured gouroboros server still hits its own 60s timeout and closes the connection (`leiosvotes/client.go:51`). Worth revising once the underlying behavior is fixed rather than documenting the current gap as if it were the intended end state.

cryptodj413 (changes requested):
Reviewed the diff directly (state maps, CBOR encoding of the fallback messages, and the reachability path from `connection.go`) rather than relying on the PR description alone. The direction is right — an optional mini-protocol shouldn't tear down a shared bearer — but three of the four fallbacks don't hold up under the mini-protocol's own state machine or wire format, and the READMEs describe behavior the code doesn't implement.

Both CodeRabbit findings are addressed in e72d541, except `leiosfetch/server_test.go:362` which is still `assert.Equal` (cosmetic, last statement in the test either way — not blocking).

Summary, worst first (see inline comments for the specifics):

- **Merge blocker** — `leiosvotes` `StateBusy` has no exit transition and, when unconfigured, no timeout, so the unconfigured fallback leaves the protocol permanently stuck and, under load, fills the shared `recvQueueChan`/muxer buffers and stalls every other mini-protocol on the connection. Reachable by default: `connection.go` passes `c.leiosVotesConfig` (nil unless a caller opts in via `WithLeiosVotesConfig`) straight into `leiosvotes.New`.
- **Merge blocker** — `NewMsgSharePeers(nil)` and `NewMsgLastBlockAndTxsInRange(nil, nil)` encode as CBOR `null` in the relevant field(s), not `[]`/absent, per this package's own wire fixture in `messages_test.go`. Peer sharing is spoken to real `cardano-node` peers, so a `null` where an array is expected is a real interop risk. `LastBlockAndTxsInRange` additionally has no "no data" sentinel at all, so a client can't tell a null block apart from a real one.
- Non-blocking — `NewMsgVotes(nil)` has the same null-vs-empty issue on the (experimental) leios-fetch votes path; one-line fix, same as the two above.
- Non-blocking — `protocol/leiosfetch/messages.go:30-33` carries its own `NOTE:` flagging `MessageTypeNoBlock`/`MessageTypeNoBlockTxs` (10/11) as placeholder IDs pending spec confirmation. Before this PR they were only emitted by a node that opted in via a configured callback; after `handleBlockRequest`/`handleBlockTxsRequest`'s new fallbacks, every default node-to-node responder now emits them to any peer that asks, configured or not. If the ratified ID differs, an unconfigured node sends an unknown message type to a real Leios relay by default.
- Non-blocking — the leios-votes fix is one-sided: `leiosvotes/client.go:51` unconditionally sets the client's own `Busy` timeout, so a gouroboros client requesting votes from an unconfigured gouroboros server still hits its 60s timeout and closes the connection on its side, even though the server no longer errors. Worth noting in the README rather than presenting "doesn't close the bearer" as achieved for both ends.
- A few test-coverage and doc-drift items inline; not blocking on their own but worth folding into the same pass.

Not run: full `make test` (scoped to the changed packages + `protocol`/`muxer` instead), `internal/test/peersharing-interop` (needs Docker + two synced preview `cardano-node` containers — unavailable here, and it's the harness that would settle the `SharePeers` null-vs-array question against a real Haskell decoder), devnet/conformance/Antithesis, and `make nilaway`.

protocol/leiosfetch/server.go
wolf31o2: Fixed in 7e3368f8. NewMsgVotes normalizes nil at the constructor boundary, with exact CBOR coverage for 820580. The mux-backed fallback test sends two requests, exercises a sibling protocol, and verifies no connection-level error.

protocol/leiosvotes/server.go
wolf31o2: Fixed in 7e3368f8. The unconfigured Leios votes path again reports a callback-missing protocol error, and StateBusy always keeps a bounded timeout. A real mux regression with a buffered ErrorChan proves the error is observable instead of leaving the receive loop stalled.

protocol/leiosfetch/server.go
wolf31o2: Fixed in 7e3368f8. The unconfigured block-range path again returns the callback-missing error and no longer fabricates [7,null,null]. The regression asserts that handler disposition directly.

protocol/leiosfetch/README.md
wolf31o2: Fixed in 7e3368f8. The README now states that only the Votes request has a safe empty-array response; block, block-transaction, and block-range requests retain observable errors because their absence representations are placeholder or ambiguous.

protocol/peersharing/server.go
wolf31o2: Fixed in 7e3368f8. NewMsgSharePeers now normalizes nil to a non-nil empty slice, both fallback call sites pass empty slices, and exact-wire coverage requires 820180 rather than 8201f6. The mux-backed responder tests also keep the error channel empty.

protocol/leiosvotes/README.md
wolf31o2: Fixed in 7e3368f8. The unconfigured Leios votes path no longer remains pending: it reports a protocol error, preserves the bounded Busy timeout, and the README now documents that behavior without claiming both ends keep the bearer open.

protocol/leiosfetch/server_test.go
wolf31o2: Fixed in 7e3368f8. The mux-backed helper now supplies a buffered ErrorChan, asserts it remains empty, sends each safe fallback twice, and verifies a peer-sharing responder still works on the same bearer.

protocol/peersharing/README.md
wolf31o2: Fixed in 7e3368f8. The peer-sharing documentation now records the exact valid empty wire shape [1, []], matching the constructor and responder regression.

protocol/peersharing/server.go
wolf31o2: Fixed in 7e3368f8. Disabled requests are logged before the fallback with local_disabled=true, and the branch emits an explicit negotiated-mode-violation message. The regression checks the message, protocol, role, and disabled context.

protocol/peersharing/peersharing.go
wolf31o2: Fixed in 7e3368f8. The exported sentinel remains for source compatibility and now carries a Go-standard Deprecated comment pointing callers to the empty-response behavior.

arepala-uml (approved):

cryptodj413 (approved):

PR #2058, opened 2026-08-24
fix(muxer): unblock protocol unregistration

### LINKED ISSUE #2056: muxer: unblock protocol restart with queued input
## Summary

Mini-protocol unregistration can block behind an in-progress muxer delivery.
The delivery path holds the receiver mutex while waiting for channel capacity,
while unregistration needs the same mutex before it can close the receiver.
This can stall the shared connection when a protocol restarts with queued input.

## Acceptance criteria

- Unregistration wakes a blocked receiver delivery before waiting to close the
  receiver channel.
- Repeated peer-sharing termination messages cannot stall other protocols on
  the same connection.
- Receiver closure and protocol re-registration semantics remain intact.
- Deterministic race-enabled regression coverage exercises the muxer and
  peer-sharing boundary.

wolf31o2:
## Summary

- wake capacity-blocked mux deliveries when a mini-protocol unregisters
- preserve stopped receivers during restart gaps so queued stale segments do not close the shared connection
- require the peer-sharing liveness regression to deliver traffic on the other protocol

## Testing

- `go test -race ./protocol/peersharing -run '^TestRepeatedDoneDoesNotStallSharedMuxer$' -count=100`
- `go test -race ./muxer -run '^(TestUnregisterProtocolWakesBlockedDelivery|TestUnregisterProtocolBehavior|TestMuxerErrorHandling)$' -count=50`
- `go test -race ./muxer ./protocol/peersharing`
- `make test`
- `go test -v ./internal/test/conformance/...`
- `make lint`
- `make nilaway`
- `go vet ./...`
- `go build ./...`

Closes #2056

### DIFF
--- a/muxer/muxer.go
+++ b/muxer/muxer.go
@@ -76,8 +76,26 @@ type Muxer struct {
 }
 
 type segmentChannel struct {
-	mu sync.Mutex
-	ch chan *Segment
+	mu       sync.Mutex
+	ch       chan *Segment
+	done     chan struct{}
+	onceStop sync.Once
+}
+
+func (s *segmentChannel) stop() {
+	// Wake a delivery that is waiting for channel capacity before taking the
+	// mutex needed to close the receiver. Otherwise unregistration can wait
+	// forever behind that delivery while the protocol waits for
+	// unregistration to finish.
+	s.onceStop.Do(func() {
+		close(s.done)
+	})
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.ch != nil {
+		close(s.ch)
+		s.ch = nil
+	}
 }
 
 type ConnectionClosedError struct {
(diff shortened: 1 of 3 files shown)


wolf31o2:
Addressed in `452d430`.

The repeated peer-sharing messages after `MsgDone` target protocol 10 after its receiver has intentionally unregistered, so connection closure with the exact `unknown protocol ID 10` error is a valid non-stalling outcome; the error is not for the registered probe protocol. The regression now rejects every other mux/protocol error, requires the expected probe ID when delivery wins, and always drains and checks the writer result. The focused race tests pass 100 repetitions.

wolf31o2:
Addressed in 7087e67. The regression now succeeds only after probe delivery and always checks the writer result. The stricter test exposed a restart-gap failure, so stopped receivers now remain as tombstones until re-registration and stale segments are discarded without closing the shared connection. The 100-iteration race regression, full race suite, conformance suite, lint, NilAway, vet, and build pass locally.

@coderabbitai review

arepala-uml (approved):

PR #2065, opened 2026-08-25
fix(ledger): apply governance rules in Dijkstra

### LINKED ISSUE #2063: ledger: apply governance validation rules to Dijkstra transactions
## Problem

Dijkstra registers its own transaction-validation rule list but omits governance predicates that remain applicable after Conway. This allows invalid Dijkstra proposals and votes to bypass proposal structure, purpose-chain, deposit, return-account, voter, action-existence, expiry, and voting-restriction checks.

## Scope

- Register the applicable governance rules in Dijkstra validation order.
- Adapt Dijkstra protocol parameters where shared rules require Conway-compatible values.
- Ensure Dijkstra parameter-change actions participate in shared ancestry and voting classification.
- Add regression coverage for rule registration and representative invalid Dijkstra proposals and votes.
- Preserve Dijkstra-specific proposal, voting, and reference-script rules.

No wire-format change is expected.

wolf31o2:
## Summary

- register the applicable governance validation rules in Dijkstra transaction order
- share parameter-change ancestry and voting classification across eras while preserving Conway constructor boundaries
- classify the Dijkstra reference-script parameters in their security voting group
- reject typed-nil governance actions before shared validation dispatch
- cover ordered registration and representative invalid proposals and votes

## Validation

- exact-base regression proof
- focused Dijkstra race tests
- full ledger race tests
- conformance race tests
- `go test ./...`
- golangci-lint
- NilAway
- `go vet`

Closes #2063

### DIFF
--- a/ledger/conway/rules.go
+++ b/ledger/conway/rules.go
@@ -209,7 +209,7 @@ func UtxoValidateProposalProcedures(
 ) error {
 	for _, proposal := range tx.ProposalProcedures() {
 		govAction := proposal.GovAction()
-		if govAction == nil {
+		if isNilGovAction(govAction) {
 			continue
 		}
 
@@ -238,12 +238,13 @@ func UtxoValidateEmptyTreasuryWithdrawals(
 ) error {
 	for _, proposal := range tx.ProposalProcedures() {
 		govAction := proposal.GovAction()
-		if govAction == nil {
+		if isNilGovAction(govAction) {
 			continue
 		}
 
 		// Check if this is a TreasuryWithdrawalGovAction with empty withdrawals
-		if twAction, ok := govAction.(*common.TreasuryWithdrawalGovAction); ok {
+		if twAction, ok := govAction.(*common.TreasuryWithdrawalGovAction); ok &&
+			twAction != nil {
 			if len(twAction.Withdrawals) == 0 {
 				return EmptyTreasuryWithdrawalsError{}
 			}
@@ -283,7 +284,7 @@ func UtxoValidateBootstrapAllowedGovActions(
 	}
 	for _, proposal := range tx.ProposalProcedures() {
 		govAction := proposal.GovAction()
-		if govAction == nil {
+		if isNilGovAction(govAction) {
 			continue
 		}
 		// NOTE: closed-set type-switch over the 7 GovActionType constants in
@@ -336,7 +337,7 @@ func UtxoValidateBootstrapParameterGroups(
 	}
 	for _, proposal := range tx.ProposalProcedures() {
 		govAction := proposal.GovAction()
-		if govAction == nil {
+		if isNilGovAction(govAction) {
 			continue
 		}
 		paramChange, ok := govAction.(*ConwayParameterChangeGovAction)
@@ -371,12 +372,13 @@ func UtxoValidateProposalNetworkIds(
 
 		// Check addresses within governance actions
 		govAction := proposal.GovAction()
-		if govAction == nil {
+		if isNilGovAction(govAction) {
 			continue
 		}
 
 		// TreasuryWithdrawalGovAction contains withdrawal addresses
-		if twAction, ok := govAction.(*common.TreasuryWithdrawalGovAction); ok {
+		if twAction, ok := govAction.(*common.TreasuryWithdrawalGovAction); ok &&
+			twAction != nil {
 			for addr := range twAction.Withdrawals {
 				if addr.NetworkId() != networkId {
 					badAddrs = append(badAddrs, *addr)
@@ -565,7 +567,7 @@ func (r *govActionResolver) resolve(
 	actionId common.GovActionId,
 ) (actionType common.GovActionType, action common.GovAction, ok bool) {
 	if proposal, found := r.txProposal(actionId); found {
-		if proposal.action == nil {
+		if isNilGovAction(proposal.action) {
 			return 0, nil, false
 		}
 		resolvedType, ok := govActionValidationType(proposal.action)
@@ -615,7 +617,7 @@ func hardForkProposedVersion(
 	action common.GovAction,
 ) (major, minor uint, ok bool) {
 	hf, isHardFork := action.(*common.HardForkInitiationGovAction)
-	if !isHardFork {
+	if !isHardFork || hf == nil {
 		return 0, 0, false
 	}
 	return hf.ProtocolVersion.Major, hf.ProtocolVersion.Minor, true
@@ -794,7 +796,7 @@ func UtxoValidateHardForkCanFollow(
 	txProposalsLoaded := false
 	for idx, proposal := range tx.ProposalProcedures() {
 		hf, ok := proposal.GovAction().(*common.HardForkInitiationGovAction)
-		if !ok {
+		if !ok || hf == nil {
 			continue
 		}
 		newPV := hf.ProtocolVersion
@@ -827,7 +829,7 @@ func UtxoValidateHardForkCanFollow(
 				}
 				ancestorAction = ancestorState.Action
 			}
-			if ancestorAction == nil {
+			if isNilGovAction(ancestorAction) {
 				continue
 			}
 			major, minor, isHardFork := hardForkProposedVersion(ancestorAction)
@@ -895,7 +897,7 @@ func UtxoValidateProposalAncestry(
 	txProposals := txProposalActions(tx)
 	for idx, proposal := range proposals {
 		govAction := proposal.GovAction()
-		if govAction == nil {
+		if isNilGovAction(govAction) {
 			continue
 		}
 		ancestorId, purpose, hasPurpose := govActionAncestor(govAction)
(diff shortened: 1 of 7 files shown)


ledger/conway/rules.go
```diff
@@ -416,8 +416,8 @@ func govActionAncestor(
 	ga common.GovAction,
 ) (ancestor *common.GovActionId, purpose govActionPurpose, ok bool) {
 	switch a := ga.(type) {
-	case *ConwayParameterChangeGovAction:
-		return a.ActionId, govPurposePParamUpdate, true
+	case common.ParameterChangeGovAction:
+		return a.PreviousGovActionId(), govPurposePParamUpdate, true
```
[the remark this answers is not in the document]
wolf31o2: Fixed in 2956482. Shared governance validation now rejects typed-nil actions before dispatch, and the parameter-change interface methods are nil-safe. Added an exact-base panic regression; focused and full ledger race tests, vet, lint, and NilAway pass.

cryptodj413 (commented):
Verified the two claims in the PR summary independently: the governance-rule-parity work is correct and well-tested; the typed-nil claim ("reject typed-nil governance actions before shared validation dispatch") is only partly true and the summary overstates it. No merge blockers, but I'd want the pipeline-ordering gap below addressed before calling the panic fully closed — see inline comments for specifics.

**Rule parity (the actual point of the PR) — confirmed correct.** Diffed the new Dijkstra rule list against `origin/main` (pure insertion, nothing removed/reordered) and against Conway's list with `conway.` prefixes stripped (identical order apart from one Dijkstra-only rule appended at the end). All eight predicates the issue names (proposal structure, purpose-chain, deposit, return-account, voter, action-existence, expiry, voting-restriction) are registered in Conway's order; the Dijkstra-specific rules this PR needed to preserve are all still present at their original relative positions. Confirmed two independent regressions genuinely require the fix: reverting just the 10 new registrations fails 5 subtests, and reverting just the two `isNilGovAction` guards makes the typed-nil test panic instead of pass.

**Typed-nil — real fix, incomplete claim.** `isNilGovAction` (a `reflect`-based check, the correct approach for this Go footgun) is applied at 5 of the ~11 dispatch sites where a `common.GovAction`-family value gets type-asserted then a method called. It's missing at `ledger/dijkstra/rules.go`'s pre-existing `UtxoValidateProposalProcedures` (rule index 1) and at `ledger/conway/rules.go:830`'s plain-interface `ancestorAction == nil` check inside `UtxoValidateProposalAncestry` (which does NOT catch a typed-nil pointer) — see inline comments. Driving the full Dijkstra pipeline with a typed-nil parameter-change action panics at index 1, before the new guard at index 2 ever runs. Not wire-reachable today (both eras' `UnmarshalCBOR` always produce a non-nil concrete action or an error, confirmed by reading `ledger/dijkstra/gov.go` and `ledger/conway/gov.go`), so this is a programmatic-construction risk rather than a live-network DoS — hence non-blocking rather than a blocker. Still, `ledger/conway/rules.go:830`'s site matters more than it looks: this PR newly registers `UtxoValidateHardForkCanFollow` for Dijkstra, so a `LedgerState` implementation that ever returns a typed-nil `Action` on a `GovActionState` now crashes Dijkstra validation too (Dingo's current `GovActionState` never populates `Action`, so this is currently dormant there, but the exposure is new).

Dingo context, since this is more than academic for Dijkstra: Dingo already has a live, wired-up call into `gouroboros/ledger/dijkstra.UtxoValidationRules` (`ledger/eras/dijkstra.go` → `ValidateTxFunc`), executed on real blocks when Dijkstra is enabled. That's gated off by default (`EnableDijkstra` requires the Musashi prototype network, `RunMode=leios`, or an explicit `StartEra=dijkstra` opt-in — never reachable from preview/preprod/mainnet), but it's not dormant scaffolding — a panic in this code path would take down a Musashi/leios-mode node today. Worth keeping in mind for how urgently the pipeline-ordering gap gets closed, even though it isn't wire-reachable yet.

A few smaller, non-blocking items worth a look, not called out inline: `ledger/dijkstra/gov.go:137` and `ledger/conway/gov.go:143`'s `GetPolicyHash` have no nil-receiver guard, unlike the two methods added right below them — currently protected only by the caller's own `isNilGovAction` check, so low risk, but inconsistent with the pattern this PR otherwise establishes. `conway.UtxoValidateBootstrapVotingRestrictions` (line 91) asserts `pp.(*ConwayProtocolParameters)` and will always fail against Dijkstra's own pparams type, making it a permanent no-op under Dijkstra even though `TestDijkstraGovernanceValidationRules` asserts it's registered — no mainnet impact since Dijkstra is past the relevant protocol version, but the test coverage reads as stronger than it is. `make nilaway` wasn't run against this PR specifically (network fetch required) — worth running in CI given three of the findings here are in the nil-dereference class.

Not run: `make test` (ran the equivalent `go test -race ./ledger/...` plus full-repo directly, to avoid `go mod tidy` touching the worktree), `make nilaway`, devnet/Antithesis/conformance runs (none available here — a devnet replay of real Dijkstra governance traffic against Dingo would be the check that settles whether the newly-registered ancestry rule ever false-rejects a valid proposal, given Dingo doesn't implement the optional `GovPurposeRootsState` fast-path interface).

ledger/dijkstra/rules.go
```diff
@@ -39,7 +39,12 @@ const minUtxoOverheadBytes = 160
 var UtxoValidationRules = []common.UtxoValidationRuleFunc{
 	conway.UtxoValidateMetadata,
 	UtxoValidateProposalProcedures,
+	conway.UtxoValidateGovActionWellFormedness,
```
cryptodj413, line 42: This new entry adds the typed-nil guard (`conway.UtxoValidateGovActionWellFormedness` → `isNilGovAction`) at pipeline index 2, but `UtxoValidateProposalProcedures` right above it at index 1 is unguarded and unchanged by this PR. Driving it with a typed-nil `*DijkstraParameterChangeGovAction` panics there — confirmed with a throwaway test:

```
paramChange:  rule idx 1 (dijkstra.UtxoValidateProposalProcedures) PANICKED: nil pointer dereference
```

(dereference happens a few lines down, in `UtxoValidateProposalProcedures`'s `&paramChangeAction.ParamUpdate` after the `govAction.(*DijkstraParameterChangeGovAction)` type assertion.) Not wire-reachable today since `UnmarshalCBOR` never produces a typed-nil action, so non-blocking — but worth either moving the new guard ahead of this rule in the list, or adding the same `isNilGovAction`-style check inside `UtxoValidateProposalProcedures` itself so ordering doesn't matter.

ledger/dijkstra/rules.go
```diff
@@ -80,6 +85,11 @@ var UtxoValidationRules = []common.UtxoValidationRuleFunc{
 	conway.UtxoValidateDelegation,
 	conway.UtxoValidateWithdrawals,
 	conway.UtxoValidateCommitteeCertificates,
+	conway.UtxoValidateUnknownVoters,
+	conway.UtxoValidateUnknownGovActionIds,
+	conway.UtxoValidateVotingOnExpiredGovAction,
+	conway.UtxoValidateBootstrapVotingRestrictions,
+	conway.UtxoValidateStakePoolVotingRestrictions,
```
cryptodj413, line 92: `conway.UtxoValidateBootstrapVotingRestrictions` (line 91, just above) does `pp.(*ConwayProtocolParameters)` internally and will never match Dijkstra's own pparams type — it's a permanent no-op when registered here, unlike this line's `UtxoValidateStakePoolVotingRestrictions`/the deposit/hard-fork rules elsewhere in this PR, which correctly go through `conwayPparams()` to survive the assertion. No consensus impact (Dijkstra is past the relevant bootstrap protocol version), but `TestDijkstraGovernanceValidationRules` asserting this rule is *registered* reads as coverage it doesn't actually provide. Worth either wrapping it through `conwayPparams()` like its neighbors, or a comment noting it's intentionally inert for Dijkstra.

ledger/dijkstra/rules.go
wolf31o2: Fixed in 280b674. `UtxoValidateProposalProcedures` now detects the typed-nil action before dereferencing the Dijkstra parameter update. The regression drives `common.VerifyTransaction` through the full ordered rule list, confirms there is no panic, and asserts the existing malformed-action error from the next validation rule.

ledger/dijkstra/rules.go
wolf31o2: Fixed in 280b674. Dijkstra now registers its own wrapper, converts the protocol parameters through `conwayPparams`, and then calls the Conway bootstrap-voting rule. The regression proves a PV9 DRep vote on a parameter change is rejected while an Info action remains allowed.

wolf31o2:
Addressed the full human-review set in 280b674 and 1255fb0: the ordered Dijkstra pipeline no longer dereferences typed-nil actions, the bootstrap-voting rule now receives converted parameters and is behavior-tested, typed-nil Conway ancestors are handled without panic, related dispatch sites are guarded consistently, and both parameter-change policy accessors are nil-safe. Focused race tests, full ledger tests, 315 conformance vectors, lint, NilAway, and build pass locally.

cryptodj413 (approved):

PR #2068, opened 2026-08-25
fix(ledger): exempt script withdrawals from DRep gate

wolf31o2:
## Summary

- apply the Conway PV10/PV11 withdrawal DRep check only to key-hash reward credentials
- check every key-hash credential present in withdrawals, including zero amounts
- preserve reward-account registration and amount validation for script credentials
- retain undelegated key-hash rejection and PV12 behavior

## Tests

- `make test`
- `make lint`
- `GOFLAGS=-buildvcs=false make build`

Related: blinklabs-io/dingo#3376

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -44,7 +44,7 @@ grep -cE "^    --- FAIL: TestRulesConformanceVectors/" /tmp/conformance.txt
 
 ## Rules Claude gets wrong without being told
 
-1. Era delegation is not universal — grep the era's function body before describing behavior; don't assume it forwards to the previous era. E.g. `conway.UtxoValidateWithdrawals` (`ledger/conway/rules.go`) skips phase-2-invalid txs, checks reward-account registration, requires key-hash reward credentials to have a DRep vote delegation at PV10/PV11, exempts script-hash reward credentials, and removes the requirement at PV12 per CIP-181.
+1. Era delegation is not universal — grep the era's function body before describing behavior; don't assume it forwards to the previous era. E.g. `conway.UtxoValidateWithdrawals` (`ledger/conway/rules.go`) skips phase-2-invalid txs, checks reward-account registration, requires every key-hash reward credential present in withdrawals (including zero amounts) to have a DRep vote delegation at PV10/PV11, exempts script-hash reward credentials, and removes the requirement at PV12 per CIP-181.
 2. `DecodeStoreCbor` requires a custom `UnmarshalCBOR` that calls `SetCbor(cborData)`. Missing → `Cbor()` returns nil → hashing breaks.
 3. Hash from preserved `.Cbor()` bytes. Never re-encode and hash.
 4. CBOR `EncMode`/`DecMode` are globally cached via `sync.Once` in `cbor/{encode,decode}.go`. Don't construct per call.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -44,7 +44,13 @@ grep -cE "^    --- FAIL: TestRulesConformanceVectors/" /tmp/conformance.txt
 
 ## Rules Claude gets wrong without being told
 
-1. Era delegation is not universal — grep the era's function body before describing behavior; don't assume it forwards to the previous era. E.g. `conway.UtxoValidateWithdrawals` (`ledger/conway/rules.go`) skips phase-2-invalid txs, checks reward-account registration, requires every key-hash reward credential present in withdrawals (including zero amounts) to have a DRep vote delegation at PV10/PV11, exempts script-hash reward credentials, and removes the requirement at PV12 per CIP-181.
+1. Era delegation is not universal — grep the era's function body before
+   describing behavior; don't assume it forwards to the previous era. E.g.
+   `conway.UtxoValidateWithdrawals` (`ledger/conway/rules.go`) skips
+   phase-2-invalid txs, checks reward-account registration, requires every
+   key-hash reward credential present in withdrawals (including zero amounts)
+   to have a DRep vote delegation at PV10/PV11, exempts script-hash reward
+   credentials, and removes the requirement at PV12 per CIP-181.
 2. `DecodeStoreCbor` requires a custom `UnmarshalCBOR` that calls `SetCbor(cborData)`. Missing → `Cbor()` returns nil → hashing breaks.
 3. Hash from preserved `.Cbor()` bytes. Never re-encode and hash.
 4. CBOR `EncMode`/`DecMode` are globally cached via `sync.Once` in `cbor/{encode,decode}.go`. Don't construct per call.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -44,7 +44,7 @@ grep -cE "^    --- FAIL: TestRulesConformanceVectors/" /tmp/conformance.txt
 
 ## Rules Claude gets wrong without being told
 
-1. Era delegation is not universal — grep the era's function body before describing behavior; don't assume it forwards to the previous era. E.g. `conway.UtxoValidateWithdrawals` (`ledger/conway/rules.go`) skips phase-2-invalid txs, checks reward-account registration, requires a DRep vote delegation at PV10/PV11, and removes that requirement at PV12 per CIP-181.
+1. Era delegation is not universal — grep the era's function body before describing behavior; don't assume it forwards to the previous era. E.g. `conway.UtxoValidateWithdrawals` (`ledger/conway/rules.go`) skips phase-2-invalid txs, checks reward-account registration, requires key-hash reward credentials to have a DRep vote delegation at PV10/PV11, exempts script-hash reward credentials, and removes the requirement at PV12 per CIP-181.
 2. `DecodeStoreCbor` requires a custom `UnmarshalCBOR` that calls `SetCbor(cborData)`. Missing → `Cbor()` returns nil → hashing breaks.
 3. Hash from preserved `.Cbor()` bytes. Never re-encode and hash.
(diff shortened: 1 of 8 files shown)


CLAUDE.md
```diff
@@ -44,7 +44,7 @@ grep -cE "^    --- FAIL: TestRulesConformanceVectors/" /tmp/conformance.txt
 
 ## Rules Claude gets wrong without being told
 
-1. Era delegation is not universal — grep the era's function body before describing behavior; don't assume it forwards to the previous era. E.g. `conway.UtxoValidateWithdrawals` (`ledger/conway/rules.go`) skips phase-2-invalid txs, checks reward-account registration, requires a DRep vote delegation at PV10/PV11, and removes that requirement at PV12 per CIP-181.
+1. Era delegation is not universal — grep the era's function body before describing behavior; don't assume it forwards to the previous era. E.g. `conway.UtxoValidateWithdrawals` (`ledger/conway/rules.go`) skips phase-2-invalid txs, checks reward-account registration, requires every key-hash reward credential present in withdrawals (including zero amounts) to have a DRep vote delegation at PV10/PV11, exempts script-hash reward credentials, and removes the requirement at PV12 per CIP-181.
```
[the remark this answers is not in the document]
wolf31o2: Wrapped the guidance to the 80-column limit in d8a463aa.

arepala-uml (dismissed):
Verified against the actual PR branch (build, vet, and the full `ledger/...` test suite all pass).

- This fixes a real mainnet incident, not a theoretical case: blinklabs-io/dingo#3376 shows a canonical mainnet transaction (block 13,844,577, slot 195889689, confirmed via Koios and a reference `cardano-node`) was wrongly rejected by rule 46 because a script-hash reward credential was subjected to the PV10/11 key-hash-only DRep delegation gate. The new tests use that exact reward address, confirmed to decode as a script-hash credential.
- Proved the regression tests are meaningful: reverted `ledger/conway/rules.go` to its pre-fix state and reran — the new/changed subtests (`PV10/11 script credential`, `PV10/11 undelegated zero amount`, `PV10 mixed amounts`) fail exactly as expected, then pass again with the fix restored.
- The credential-type check (`CredType != CredentialTypeAddrKeyHash`) is idiomatic — the same pattern gates script exemptions elsewhere in `ledger/common/rules.go` and `gov.go`.
- Dropping `amount` from the loop is correct: this function only gates on delegation (amount validation lives in `shelley.UtxoValidateWithdrawals`, unchanged here), and now correctly includes zero-amount withdrawals since the delegation gate is amount-independent.
- Doc updates (AGENTS.md/CLAUDE.md/ARCHITECTURE.md/README.md/state.go/errors.go) checked against current code — accurate; the "exact reward balance" language in AGENTS.md reflects an already-merged prior commit, not new behavior from this PR.

CodeRabbit's one finding (80-column wrap) is already fixed and verified on the current head. Cubic: no actionable findings.

ledger/conway/rules_test.go
```diff
@@ -229,30 +321,32 @@ func TestUtxoValidateWithdrawals_DRepDelegationProtocolGate(t *testing.T) {
...
-		require.ErrorAs(t, err, &target)
-		assert.Equal(t, otherRewardAddr, target.RewardAddress)
+		))
+		assert.Equal(t, map[common.Blake2b224]int{
+			stakeKeyHash:      1,
+			otherStakeKeyHash: 1,
+		}, lookups)
 	})
 
 	t.Run("PV10 requires delegation lookup support", func(t *testing.T) {
```
cryptodj413, line 389: Nice regression coverage overall. Two gaps I checked by hand (scratch tests, `-count=20` to defeat map-order luck, removed after) that aren't locked in by the suite:

1. **`DRepDelegationStateUnavailableError` is no longer returned for script-only withdrawals.** Before this PR, `delegationState` was resolved before the credential type was checked, so a script-only withdrawal against a ledger state that doesn't implement `DRepDelegationState` used to error. Now it returns `nil` — the correct new contract (it relaxes what a `LedgerState` must implement), but nothing here exercises it. Worth a case: script-only withdrawal + a `LedgerState` that doesn't implement `DRepDelegationState`, asserting `NoError`.
2. **Mixed script + undelegated key-hash in one withdrawal map.** The existing "mixed amounts" case (line 314) uses two key-hash credentials, both delegated — the adversarial mix (script credential exempt, key-hash credential rejected, same tx) is untested. It's the case most likely to regress if the `continue` at line 3205 is later refactored.

ledger/conway/rules.go
```diff
@@ -3199,8 +3200,9 @@ func UtxoValidateWithdrawals(
 		return nil
 	}
 	var delegationState common.DRepDelegationState
-	for addr, amount := range withdrawals {
-		if amount == nil || amount.Sign() == 0 {
+	for addr := range withdrawals {
+		credential, ok := addr.StakeCredential()
+		if !ok || credential.CredType != common.CredentialTypeAddrKeyHash {
```
cryptodj413, line 3205: Non-blocking, pre-existing (not introduced by this PR): the registration/amount check (, called a few lines above at line 3184) still runs before this DRep gate, while upstream () runs the DRep check first. Both reject an unregistered key-hash reward account at PV10/PV11, so accept/reject consensus is unaffected, but the surfaced error differs from a reference  ( here vs.  upstream). Only matters if Dingo/downstream ever surfaces these error strings for comparison against a reference node.

cryptodj413 (commented):
Verified this against upstream, not just against the PR description. No merge blockers found — leaving this as a comment review rather than an approval given the stakes (this fixes a live mainnet consensus rejection, blinklabs-io/dingo#3376), so a maintainer's own sign-off should be the approving review.

**Correctness — confirmed against upstream `cardano-ledger`.** Fetched and read `Conway/Rules/Ledger.hs`, `Conway/Era.hs`, `Dijkstra/Rules/Ledger.hs`, and `Shelley/Rules/Ledger.hs` from `IntersectMBO/cardano-ledger`. Three independent points of agreement: (1) the DRep-delegation requirement is genuinely key-hash-only in the spec — `credKeyHash` in `validateWithdrawalsDelegated` silently drops script credentials from the list comprehension, and the failure type `ConwayWdrlNotDelegatedToDRep (NonEmpty (KeyHash Staking))` can't even carry a script hash; (2) the check is amount-independent (`Map.toList wdrls` discards amount), so removing the `amount.Sign() == 0` skip is correct, not an over-reach; (3) the phase-2-valid gating (`rules.go:3181`) matches `Ledger.hs`'s `isPhase2ValidTxL` guard.

**Regression test is genuinely load-bearing.** `rules_test.go:72-` uses the literal mainnet reward account from dingo#3376, asserts `CredType == CredentialTypeScriptHash`, and asserts zero DRep lookups occurred (an absence assertion, not just happy-path). Reverted `ledger/conway/rules.go` to `origin/main` with these tests in place and reproduced the mainnet rejection string character-for-character, including the stake address — this is testing the real scenario, not a synthetic stand-in.

**Retained behavior confirmed:** undelegated key-hash still rejected at PV10/PV11 (`rules_test.go:94-131`), PV12/Dijkstra removal untouched (version gate at `rules.go:3197-3200` unchanged by the diff).

**Worth naming explicitly (not a defect):** commit 2a45cbc changes a *zero-amount* key-hash withdrawal from accepted to rejected-if-undelegated — a real behavior tightening on mainnet, in the opposite direction from the bug being fixed. It's correct per the upstream source (no amount filter there either) and reachable in practice (pre-Dijkstra Shelley's exact-drain requirement means a zero withdrawal implies a registered zero-balance account that may legitimately lack DRep delegation), but it hasn't been checked against a live `cardano-node` — only against the fetched Haskell source. Worth a maintainer's explicit sign-off on that point given it's a new rejection path on mainnet, not just a new acceptance path.

Dependency note: traced Dingo's consumption of this rule (`ledger/eras/conway.go` wraps `conway.UtxoValidationRules` unmodified, no local reimplementation or workaround) — Dingo needs nothing beyond the eventual `gouroboros` version bump once this merges and is tagged; issue #3376 already documents that plan.

See inline comments for two coverage gaps and a couple of non-blocking notes.

Not run: full `make test` (ran the equivalent `go test -race ./ledger/...` plus full-repo and conformance suites directly, to avoid `go mod tidy` touching the worktree), `make nilaway`, live `cardano-node`/devnet/Antithesis cross-validation of the zero-amount tightening, and direct on-chain confirmation of the referenced tx (relied on the byte-exact reproduction of the reported error string and stake address instead).

ledger/conway/rules.go
wolf31o2: Confirmed: this is pre-existing error precedence only; accept/reject behavior is unchanged. No production change in this scope.

ledger/conway/rules_test.go
wolf31o2: Covered in `54dc16c`. A script-only withdrawal now passes with a ledger-state wrapper that does not expose `DRepDelegationState`. The mixed case first proves a script credential is never sent to the DRep lookup, then proves an undelegated key credential in the same map is rejected.

Both subtests fail against the pre-fix rule and pass repeatedly with the current rule under `-race`.

wolf31o2:
Added the requested withdrawal-boundary coverage in `54dc16c`:

- script-only withdrawals do not require the optional DRep delegation interface;
- mixed script/key withdrawals exempt the script credential and still reject an undelegated key credential.

The focused regression test fails against the pre-fix rule, passes with the fix at `-race -count=10`, and the full Conway race package and focused vet pass.

cryptodj413 (dismissed):

cryptodj413 (approved):

PR #2092, opened 2026-08-25
fix(ledger): reject duplicate stake registrations

### LINKED ISSUE #2069: ledger: reject duplicate Conway stake registrations
## Summary

Conway delegation validation records stake-registration certificates in
transaction-local state without first rejecting a credential that is already
registered. This allows an invalid registration to pass the delegation rule.

## Acceptance criteria

- Reject plain and registration-with-delegation certificate variants when the
  stake credential is already registered in ledger state.
- Reject a second registration of the same credential within one transaction.
- Preserve valid deregistration followed by registration in one transaction.
- Return `StakeCredentialAlreadyRegisteredError` for duplicate registrations.
- Add regression coverage for the affected certificate paths.

wolf31o2:
## Summary

- reject duplicate stake registration across all Conway registration certificate variants
- track transaction-local registration and deregistration by full credential identity
- preserve valid deregistration followed by registration in the same transaction

## Checks

- focused duplicate-registration regressions under `-race`
- full ledger race suite
- race-enabled conformance suite
- repository and example-module race tests
- lint, vet, build, NilAway, module verification, and formatting

Closes #2069

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate stake registrations within a transaction.
  * Correctly handles stake credentials by type, including key-hash and script-hash credentials.
  * Allows re-registration after a same-transaction deregistration.
* **Tests**
  * Added coverage for registration, deregistration, and delegation certificate validation scenarios.

### DIFF
--- a/ledger/conway/rules.go
+++ b/ledger/conway/rules.go
@@ -2939,8 +2939,20 @@ func UtxoValidateDelegation(
 	ls common.LedgerState,
 	pp common.ProtocolParameters,
 ) error {
-	// Track credentials/pools/DReps registered within this transaction
-	inTxStakeRegs := make(map[common.Blake2b224]bool)
+	// Track credential registration state changes within this transaction.
+	// The bool records both registrations and deregistrations so later
+	// certificates observe the state produced by earlier certificates.
+	type stakeCredentialKey struct {
+		credType uint
+		hash     common.Blake2b224
+	}
+	stakeKey := func(cred common.Credential) stakeCredentialKey {
+		return stakeCredentialKey{
+			credType: cred.CredType,
+			hash:     cred.Credential,
+		}
+	}
+	inTxStakeState := make(map[stakeCredentialKey]bool)
 	inTxPoolRegs := make(map[common.PoolKeyHash]bool)
 	inTxDRepRegs := make(map[common.Blake2b224]bool)
 	// Track VRF keys seen in this transaction (for PV11+ duplicate detection)
@@ -2948,8 +2960,20 @@ func UtxoValidateDelegation(
 
 	// Helper to check if stake credential is registered (in state or in-tx)
 	isStakeRegistered := func(cred common.Credential) bool {
-		return ls.IsStakeCredentialRegistered(cred) ||
-			inTxStakeRegs[cred.Credential]
+		if registered, ok := inTxStakeState[stakeKey(cred)]; ok {
+			return registered
+		}
+		return ls.IsStakeCredentialRegistered(cred)
+	}
+
+	registerStakeCredential := func(cred common.Credential) error {
+		if isStakeRegistered(cred) {
+			return StakeCredentialAlreadyRegisteredError{
+				Credential: cred,
+			}
+		}
+		inTxStakeState[stakeKey(cred)] = true
+		return nil
 	}
 
 	// Helper to check if pool is registered (in state or in-tx)
@@ -3009,10 +3033,14 @@ func UtxoValidateDelegation(
 		switch c := cert.(type) {
 		// Track registrations for in-tx state
 		case *common.RegistrationCertificate:
-			inTxStakeRegs[c.StakeCredential.Credential] = true
+			if err := registerStakeCredential(c.StakeCredential); err != nil {
+				return err
+			}
 
 		case *common.StakeRegistrationCertificate:
-			inTxStakeRegs[c.StakeCredential.Credential] = true
+			if err := registerStakeCredential(c.StakeCredential); err != nil {
+				return err
+			}
 
 		case *common.PoolRegistrationCertificate:
 			inTxPoolRegs[c.Operator] = true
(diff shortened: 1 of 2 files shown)


wolf31o2:
Rebuilt the branch on current main and narrowed it to the delegation-rule change and regression tests. Revalidated the focused regression under race, the complete Conway package, vet, and diff checks. Current-head CI and bot review are running.

cryptodj413:
@cubic-dev-ai review this PR

ledger/conway/rules.go
```diff
@@ -3043,10 +3071,10 @@ func UtxoValidateDelegation(
 
 		// Track deregistrations for in-tx state
 		case *common.StakeDeregistrationCertificate:
-			delete(inTxStakeRegs, c.StakeCredential.Credential)
+			inTxStakeState[stakeKey(c.StakeCredential)] = false
 
 		case *common.DeregistrationCertificate:
-			delete(inTxStakeRegs, c.StakeCredential.Credential)
+			inTxStakeState[stakeKey(c.StakeCredential)] = false
```
cryptodj413, line 3077: Beyond rejecting duplicate registrations, writing an explicit `false` here (rather than deleting the map entry) combined with `isStakeRegistered` now preferring in-tx state over ledger state (line ~2962) also tightens `StakeDelegationCertificate`/`VoteDelegationCertificate`/`StakeVoteDelegationCertificate`: a credential deregistered earlier in the same transaction is now correctly rejected for a same-tx delegation, where it was previously silently accepted. Confirmed with a throwaway test (`StakeDeregistrationCertificate(X)` then `StakeDelegationCertificate(X, pool)` in one tx, X registered in ledger state): pre-fix `err == nil`, post-fix `DelegateUnregisteredStakeCredentialError`. This matches `Deleg.hs`'s `StakeKeyNotRegisteredDELEG` and is correct, but it's a second, unstated behavior change — worth a line in the commit message and a regression test alongside the duplicate-registration ones.

ledger/conway/rules_test.go
```diff
@@ -3203,6 +3203,193 @@ func TestPoolValidateVrfKeyUniqueness(t *testing.T) {
...
+		},
+		{
+			name: "registration",
+			cert: func() common.Certificate {
+				return &common.RegistrationCertificate{
+					StakeCredential: stakeCred,
+				}
+			},
+		},
+		{
```
cryptodj413, line 3258: None of the four registration+delegation subtests actually pin the check order (already-registered vs. unregistered-pool/DRep), because `poolKeyHash` is always registered via `WithPools` in `newLedgerState`. Reordering `registerStakeCredential` to run *after* the pool/DRep check in `rules.go` would still pass every case here. Consider adding a case with an unregistered pool (or DRep) plus an already-registered credential, asserting `StakeCredentialAlreadyRegisteredError` specifically — matching `Deleg.hs`'s deposit → already-registered → delegatee check order.

cryptodj413 (commented):
Reviewed current head 7fc2643 (build, `go vet ./ledger/...`, `go test -race ./ledger/...`, `go test ./internal/test/conformance/...`, `golangci-lint run ./ledger/conway/...`, `gofmt -s -l ledger/conway/` all clean; whole-repo `go test ./...` also green). No merge blocker.

Verified against upstream `cardano-ledger` (`Cardano.Ledger.Conway.Rules.Deleg`): rejecting duplicate registrations for all 5 certificate types (Reg, StakeReg, StakeRegDeleg, VoteRegDeleg, StakeVoteRegDeleg) matches `checkAccountAlreadyRegistered`/`DelegAccountAlreadyRegistered`, and it isn't gated by the Conway bootstrap-phase flag, so the fix applies unconditionally as written. Certificate list order is preserved end-to-end, so the `registerStakeCredential` short-circuit before the pool/DRep check matches the spec's check order.

Proved the new regression test actually catches the bug: reverted `rules.go` to the pre-fix commit while keeping the new test — 6 of 8 duplicate-detection subtests fail as expected; the two negative-case subtests (unregistered, registered-then-deregistered) correctly still pass pre-fix.

Two non-blocking notes left inline. Also confirmed, but out of scope for this PR: Shelley's `UtxoValidateDelegation` (reused by Allegra/Mary/Alonzo/Babbage) has the identical duplicate-registration gap and never rejects a repeat `StakeRegistrationCertificate` — probed directly, `err == nil` in both Shelley and Babbage. Worth its own follow-up issue; a fix there needs care since falsely rejecting historical registrations during a from-genesis replay would be a regression.

Bot review: CodeRabbit completed on this head with no actionable comments. Cubic could not run — org monthly review-line quota exhausted, resumes 2026-09-01 (documented in-thread). Team review already requested from blinklabs-io/core.

ledger/conway/rules.go
wolf31o2: Covered by #2102 with regression cases for stake, vote, and combined delegation after same-transaction deregistration.

ledger/conway/rules_test.go
wolf31o2: Covered by #2102 with duplicate-registration ordering cases against unregistered pool and DRep targets.

PR #2110, opened 2026-08-27
fix(ledger): accept null origin prev_hash in Shelley-family headers

wolf31o2:
`270c74c` accepted a null `prev_hash` in `BabbageBlockHeaderBody` but left `ShelleyBlockHeaderBody` strict. That body serves Shelley, Allegra, Mary, and Alonzo, so a chain whose first block is in any of those eras still fails to decode:

```
decode Alonzo block error: decode blake2b-256 hash: expected CBOR byte string
```

## Cause

The CDDL declares the field as `prev_hash : hash32 / nil` (`ledger/dijkstra/testdata/dijkstra.cddl:55`); an origin header has no predecessor to reference. `a9f0980` gave `Blake2b256` a strict `UnmarshalCBOR` that rejects any non-byte-string input, which is correct for every other `hash32` field — `prev_hash` is the only nullable one in the CDDL. So the tolerance belongs on the field, not on the hash type.

This mirrors the Babbage wire-struct approach exactly rather than relaxing `Blake2b256`.

## Evidence

Decoding 400 real Alonzo blocks taken from Dingo's immutable-DB fixture, unchanged across all runs:

| gouroboros | result |
|:---|:---|
| `v0.197.0` | 400/400 decode |
| `760e96d` (parent of `a9f0980`) | 400/400 decode |
| `a9f0980` | 399/400 — block 0 fails |
| `v0.201.1` | 399/400 |
| `origin/main` (with `270c74c`) | 399/400 |
| this branch | 400/400 |

The one failing block is the chain's first: `block_number=0`, `slot=0`, `prev_hash=null`. `v0.197.0` decoded it to the all-zero hash, which is the value this change restores.

## Scope

Both header body types now handle it; `DijkstraBlockHeader` reuses `BabbageBlockHeaderBody`, and Byron has no `Blake2b256` prev-hash field, so the class is complete.

`TestShelleyBlockHeaderPreviousHashDecoding` mirrors the Babbage cases: origin null, exact hash, short hash, non-byte value, and an assertion that bare `Blake2b256` still rejects null. Reverting the `shelley.go` change fails the `origin null` case.

`go test -race ./...`, `go vet ./...`, `gofmt`, and `golangci-lint run ./ledger/...` all clean.

## Downstream

Dingo's `fix/3376-gouroboros-v0199` (blinklabs-io/dingo#3539) bumps to `v0.201.1` and sees 121 test failures across 5 packages from this single block. It stays blocked until this lands and is released.

---
## Summary by cubic
Accepts a null `prev_hash` in `ShelleyBlockHeaderBody` so chains whose first block is in the Shelley, Allegra, Mary, or Alonzo eras decode successfully. A previous fix covered only `BabbageBlockHeaderBody`; this extends the same tolerance to the Shelley body.

- Origin headers encode `prev_hash` as null because they have no predecessor; the CDDL declares it `hash32 / nil`.
- Null `prev_hash` decodes to the zero hash, restoring `v0.197.0` behavior.
- Non-null hashes and the global `Blake2b256` decoder stay strict.
- New tests cover origin null, exact hash, short hash, and non-byte values.

<sup>Written for commit 9810eb22673a98946e6effdfa4116130fb211043. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2110?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed decoding of Shelley origin block headers when the previous hash is null.
  * Preserved strict validation for valid, malformed, short, and non-byte previous-hash values.

* **Tests**
  * Added coverage for origin headers, valid hash round-trips, invalid hash lengths, and unsupported value types.

### DIFF
--- a/ledger/shelley/shelley.go
+++ b/ledger/shelley/shelley.go
@@ -234,13 +234,56 @@ type ShelleyBlockHeaderBody struct {
 	ProtoMinorVersion    uint64
 }
 
+type shelleyBlockHeaderBodyWire struct {
+	cbor.StructAsArray
+	BlockNumber          uint64
+	Slot                 uint64
+	PrevHash             cbor.RawMessage
+	IssuerVkey           common.IssuerVkey
+	VrfKey               []byte
+	NonceVrf             common.VrfResult
+	LeaderVrf            common.VrfResult
+	BlockBodySize        uint64
+	BlockBodyHash        common.Blake2b256
+	OpCertHotVkey        []byte
+	OpCertSequenceNumber uint32
+	OpCertKesPeriod      uint32
+	OpCertSignature      []byte
+	ProtoMajorVersion    uint64
+	ProtoMinorVersion    uint64
+}
+
 func (b *ShelleyBlockHeaderBody) UnmarshalCBOR(cborData []byte) error {
-	type tShelleyBlockHeaderBody ShelleyBlockHeaderBody
-	var tmp tShelleyBlockHeaderBody
+	var tmp shelleyBlockHeaderBodyWire
 	if _, err := cbor.Decode(cborData, &tmp); err != nil {
 		return err
 	}
-	*b = ShelleyBlockHeaderBody(tmp)
+	var prevHash common.Blake2b256
+	// Origin headers encode prev_hash as null. Keep that compatibility scoped
+	// to this field while non-null hashes use the globally strict decoder.
+	// Mirrors BabbageBlockHeaderBody; this body serves Shelley through Alonzo.
+	if len(tmp.PrevHash) != 1 || tmp.PrevHash[0] != 0xf6 {
+		if _, err := cbor.Decode(tmp.PrevHash, &prevHash); err != nil {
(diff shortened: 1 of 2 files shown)


wolf31o2:
Local exact-head review at `9810eb22673a98946e6effdfa4116130fb211043`: no blocking findings.

Passed: `go test ./ledger/shelley`, `gofmt` on changed files, and `git diff --check`.

Not completed: `go vet ./ledger/shelley`, because SumDB/network access was unavailable. CodeRabbit: success. Cubic: neutral. Human review remains required.

arepala-uml (approved):

PR #2111, opened 2026-08-27
fix(byron): validate payloads without mutating inputs

### LINKED ISSUE #2100: ledger/byron: validate payloads without mutating inputs
Byron validation and conversion paths do not consistently reject malformed data or preserve caller ownership:

- Delegation payloads, update proposals, and votes lack complete structural or signature validation.
- Malformed block-body proofs can produce a zero hash instead of an error.
- Proxy validation mutates a caller-owned validation input.

Acceptance criteria:

- Validate payload structure and signatures before use.
- Return errors for malformed body proofs.
- Treat validation inputs as immutable.
- Add valid, malformed, signature-mismatch, and input-reuse tests.

AniketR10:
image

https://github.com/user-attachments/assets/1dcf2697-6eea-473e-a4e4-1feb554a26b7

Fixes: #2100

This PR fixes three bugs in the Byron-era block code:

First, blocks carry delegation certificates, update proposals, and votes that were never actually inspected, the code checked that their bytes hadn't been tampered with, but never opened them up to confirm they were well-formed or that their signatures were genuine, so a certificate with a broken key or a vote signed by the wrong person slipped straight through. This PR adds that checking (off by default, behind a flag, since the update-proposal format hasn't been confirmed against a real mainnet block yet). 

Second, when a block header was too malformed to read a hash out of, the code quietly substituted all zeros and then reported a "hash mismatch"  sending anyone debugging it after a problem that didn't exist, so it now says the header is malformed instead. 

Third, a validation function was writing data back into the caller's own struct, meaning if you reused that struct for the next block, it arrived carrying the previous block's signature; it now writes to a copy.

---
## Summary by cubic
Fixes three Byron validation bugs: payloads now get structural and signature checks (opt-in via `EnableByronPayloadValidation`), malformed body proofs are reported as malformed instead of as hash mismatches, and proxy signature validation no longer mutates the caller's input.

**Bug Fixes**
- Delegation certificates, update proposals, and votes are now validated for well-formedness and genuine signatures; previously only their bytes were compared against the header proof.
- Body proofs that don't parse as a hash now return a "malformed" error instead of silently substituting a zero hash and reporting a mismatch; nil blocks and headers report the same error.
- `validateBlockSignature` extracts the signature into a copy of `ValidateHeaderInput`, preventing a reused struct from carrying the previous block's signature into the next validation.
- Payload decoding preserves each certificate's CBOR wire encoding (`[]cbor.RawMessage`) since signatures cover it; `ApplyPayloadCbor` replaces the deprecated `ApplyPayload` at the consensus layer.

The new payload validation is off by default because the update-proposal/vote signing format hasn't been confirmed against a real mainnet block; delegation certificate verification uses the same format already trusted in header validation.

<sup>Written for commit f9fc9f6573e308fb439a821331553b5cf5f77988. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2111?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## New Features
- Added optional validation for Byron delegation certificates, update proposals, and update votes, including signature verification.
- Added checked body-hash validation to identify malformed proof formats distinctly.

## Bug Fixes
- Prevented validation from modifying caller-provided data.
- Prevented reused validation inputs from retaining stale signatures or hashes.

## Tests
- Expanded coverage for valid, malformed, and signature-invalid Byron payloads and body proofs.

### DIFF
--- a/consensus/byron/pbft_delegation.go
+++ b/consensus/byron/pbft_delegation.go
@@ -179,12 +179,46 @@ func (s PBFTDelegationState) Tick(
 // payload and performs the delegation tick for that block. The update is
 // atomic: malformed or invalid certificates leave the input state unchanged.
 //
+// It takes the payload as the decoder produced it, which means each
+// certificate's epoch is re-encoded before its signature is checked. CBOR
+// admits non-shortest integer encodings that this decoder accepts, so a
+// certificate whose epoch arrived in one of those forms will not verify
+// here -- the issuer signed the bytes on the wire, and a decoded uint64
+// cannot reproduce them. See byron.DelegationCertificate.EpochCbor.
+//
+// Deprecated: use ApplyPayloadCbor, which takes the preserved encodings and
+// does not have that limitation. This method is kept so existing callers
+// holding a ByronMainBlockBody.DlgPayload keep compiling while they
+// migrate; it behaves exactly as it did before ApplyPayloadCbor existed.
+func (s PBFTDelegationState) ApplyPayload(
+	currentEpoch uint64,
+	currentSlot uint64,
+	payload []any,
+) (PBFTDelegationState, error) {
+	raw := make([]cbor.RawMessage, 0, len(payload))
+	for i, certificate := range payload {
+		encoded, err := cbor.Encode(certificate)
+		if err != nil {
+			return PBFTDelegationState{}, fmt.Errorf(
+				"encode Byron PBFT delegation certificate %d: %w", i, err,
+			)
+		}
+		raw = append(raw, encoded)
+	}
+	return s.ApplyPayloadCbor(currentEpoch, currentSlot, raw)
+}
+
+// ApplyPayloadCbor schedules the certificates in a Byron main-block
+// delegation payload and performs the delegation tick for that block. The
+// update is atomic: malformed or invalid certificates leave the input state
+// unchanged.
+//
 // The payload is taken as each certificate's preserved CBOR rather than as
 // decoded values, because a certificate's signature covers its epoch
 // field's wire encoding -- see byron.DelegationCertificate.EpochCbor. Call
 // it with byron.ByronMainBlockBody.DlgPayloadCbor decoded into
 // []cbor.RawMessage.
-func (s PBFTDelegationState) ApplyPayload(
+func (s PBFTDelegationState) ApplyPayloadCbor(
 	currentEpoch uint64,
 	currentSlot uint64,
 	payload []cbor.RawMessage,
--- a/consensus/byron/pbft_delegation.go
+++ b/consensus/byron/pbft_delegation.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"math"
 
+	"github.com/blinklabs-io/gouroboros/cbor"
+	"github.com/blinklabs-io/gouroboros/ledger/byron"
 	"github.com/blinklabs-io/gouroboros/ledger/common"
 )
 
@@ -176,10 +178,16 @@ func (s PBFTDelegationState) Tick(
 // ApplyPayload schedules the certificates in a Byron main-block delegation
 // payload and performs the delegation tick for that block. The update is
 // atomic: malformed or invalid certificates leave the input state unchanged.
+//
+// The payload is taken as each certificate's preserved CBOR rather than as
+// decoded values, because a certificate's signature covers its epoch
+// field's wire encoding -- see byron.DelegationCertificate.EpochCbor. Call
+// it with byron.ByronMainBlockBody.DlgPayloadCbor decoded into
+// []cbor.RawMessage.
 func (s PBFTDelegationState) ApplyPayload(
 	currentEpoch uint64,
 	currentSlot uint64,
-	payload []any,
+	payload []cbor.RawMessage,
 ) (PBFTDelegationState, error) {
 	next := s.clone()
 	for i, rawCertificate := range payload {
@@ -241,20 +249,13 @@ func (s PBFTDelegationState) clone() PBFTDelegationState {
 func (s *PBFTDelegationState) scheduleCertificate(
 	currentEpoch uint64,
 	currentSlot uint64,
-	rawCertificate any,
+	rawCertificate cbor.RawMessage,
 ) error {
-	certificate, ok := rawCertificate.([]any)
-	if !ok || len(certificate) != 4 {
-		return fmt.Errorf(
-			"invalid certificate shape: got %T with %d elements",
-			rawCertificate,
-			len(certificate),
-		)
-	}
-	delegationEpoch, err := extractUint64(certificate[0])
+	certificate, err := byron.ParseDelegationCertificate(rawCertificate)
 	if err != nil {
-		return fmt.Errorf("decode delegation epoch: %w", err)
+		return err
 	}
+	delegationEpoch := certificate.Epoch
 	if delegationEpoch < currentEpoch ||
 		delegationEpoch-currentEpoch > 1 {
 		return fmt.Errorf(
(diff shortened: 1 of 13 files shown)


AniketR10:
@wolf31o2 please review

ledger/byron/payload.go
```diff
@@ -0,0 +1,399 @@
...
+// pair.
+func (v *UpdateVote) Verify(protocolMagic uint32) error {
+	if v == nil {
+		return fmt.Errorf("%w: update vote is nil", ErrInvalidPayload)
+	}
+	const (
+		cborArrayLen2 byte = 0x82
+		cborTrue      byte = 0xf5
+	)
+	proposalIdCbor, err := cbor.Encode(v.ProposalId)
```
wolf31o2: Preserve the signed fields' original CBOR here. `ParseUpdateVote` discards the proposal-ID annotation, so this re-encodes an accepted non-shortest byte string (`59 00 20 ...`) as `58 20 ...`; a signature over the original wire bytes then fails with `ErrInvalidSignature`. [cardano-ledger's vote verifier](https://github.com/IntersectMBO/cardano-ledger/blob/abcf0a87b6ad060fde9c50b9e23ee176b4d56f1d/eras/byron/ledger/impl/src/Cardano/Chain/Update/Vote.hs#L215-L227) uses the annotated proposal-ID bytes, and [its certificate verifier](https://github.com/IntersectMBO/cardano-ledger/blob/abcf0a87b6ad060fde9c50b9e23ee176b4d56f1d/eras/byron/ledger/impl/src/Cardano/Chain/Delegation/Certificate.hs#L163-L177) likewise uses the annotated epoch while `sign.go` re-encodes it. Thread the raw field encodings through both parsers and add non-shortest regression vectors.

ledger/byron/payload.go
```diff
@@ -0,0 +1,399 @@
...
+}
+
+// Validate structurally validates an update proposal and verifies its
+// issuer signature, reproducing cardano-ledger-byron's
+// Cardano.Chain.Update.Proposal signing format:
+//
+//	signed = 0x04 || CBOR(protocolMagic) || signedBody()
+//
+// See signedBody for how the signed body is recovered.
+func (p *ByronUpdateProposal) Validate(protocolMagic uint32) error {
```
wolf31o2, line 329: `Validate` authenticates the first five raw fields but never validates what `Data` and `Attributes` decoded to. A proposal with `uint64` values in both fields can be correctly signed and returns `nil` here, while [the reference decoder](https://github.com/IntersectMBO/cardano-ledger/blob/abcf0a87b6ad060fde9c50b9e23ee176b4d56f1d/eras/byron/ledger/impl/src/Cardano/Chain/Update/Proposal.hs#L157-L173) requires typed metadata and calls `dropEmptyAttributes`. Enforce those shapes before accepting the proposal so `EnableByronPayloadValidation` cannot accept payloads that cardano-ledger rejects.

ledger/byron/byron.go
```diff
@@ -140,7 +140,20 @@ func (h *ByronMainBlockHeader) Era() common.Era {
...
+// gives it no way to report a malformed proof, so it stands in a zero hash
+// instead. A caller that then compares that zero hash reports a hash
+// mismatch, naming a value the block never carried, when the real fault is
+// that the proof could not be parsed at all.
+func (h *ByronMainBlockHeader) BlockBodyHashChecked() (
+	common.Blake2b256,
+	error,
+) {
 	// Byron BodyProof is an array: [tx_proof, ssc_proof, dlg_proof, upd_proof]
 	// tx_proof is: [txCount, txBodyMerkleRoot, txWitnessMerkleRoot]
```
wolf31o2, line 158: `BlockBodyHashChecked` promises malformed proofs return `ErrMalformedBodyProof`, but the `>=` checks accept oversized outer/transaction arrays and the raw-byte fallback below accepts a main-block proof that is not an array. Focused probes with a 5-element outer/4-element transaction proof and with a bare 32-byte proof both returned `nil`. Require the exact 4/3-element shapes and remove the main-block raw-byte fallback.

wolf31o2 (changes requested):

AniketR10:
please review,  made the changes

consensus/byron/pbft_delegation.go
```diff
@@ -176,10 +178,16 @@ func (s PBFTDelegationState) Tick(
...
+// The payload is taken as each certificate's preserved CBOR rather than as
+// decoded values, because a certificate's signature covers its epoch
+// field's wire encoding -- see byron.DelegationCertificate.EpochCbor. Call
+// it with byron.ByronMainBlockBody.DlgPayloadCbor decoded into
+// []cbor.RawMessage.
 func (s PBFTDelegationState) ApplyPayload(
 	currentEpoch uint64,
 	currentSlot uint64,
-	payload []any,
+	payload []cbor.RawMessage,
```
wolf31o2, line 224: **P1 — Preserve the existing Dingo call path**

This exported parameter change breaks the current Dingo caller: Dingo main passes `mainBlock.Body.DlgPayload` as `[]any`, and compiling its `ledger` package against this head fails because it cannot be used as `[]cbor.RawMessage`. Add a raw-CBOR method and migrate Dingo before removing the existing API, or land both sides together.

ledger/byron/payload.go
```diff
@@ -0,0 +1,746 @@
...
+			ErrInvalidPayload, tag, len(tag), systemTagMaxLength,
+		)
+	}
+	return nil
+}
+
+// validateInstallerHash reproduces cardano-ledger-byron's InstallerHash
+// decoder: a one-element array wrapping a blake2b-256 hash.
+func validateInstallerHash(tag string, raw cbor.RawMessage) error {
+	fields, err := payloadFields(
```
wolf31o2: **P1 — Match the reference `InstallerHash` wire shape**

Pinned cardano-ledger enforces a four-element array, discards fields 0, 2, and 3, and decodes the `Hash Raw` from field 1. This one-element check rejects reference-valid non-empty metadata, while the added fixture encodes a shape the reference rejects. A signed four-element proposal fails here with `not a 1-element array, got 4 elements`; update both the validator and fixture.

wolf31o2 (changes requested):
Two blockers remain on this replacement head: the proposal validator rejects the reference `InstallerHash` encoding, and the exported payload type change breaks Dingo's current caller.

AniketR10:
resolved both please review

ledger/byron/payload.go
wolf31o2: Addressed in `f9fc9f65`: the parsers retain signature-covered field encodings, and validation constructs signed bytes from preserved CBOR.

ledger/byron/payload.go
wolf31o2: Addressed in `f9fc9f65`: `Validate` now calls `validateProposalMetadata` and `validateProposalAttributes`, with regression cases for invalid integer fields.

ledger/byron/byron.go
wolf31o2: Addressed in `f9fc9f65`: main-block proofs now require exact outer and transaction-array lengths, with no raw-hash fallback and malformed-shape regression coverage.

consensus/byron/pbft_delegation.go
wolf31o2: Addressed in `f9fc9f65`: `ApplyPayload([]any)` remains source-compatible for Dingo, while `ApplyPayloadCbor` provides the preserved-wire path.

ledger/byron/payload.go
wolf31o2: Addressed in `f9fc9f65`: `validateInstallerHash` requires the reference four-element array and validates element 1 as the 32-byte hash.

wolf31o2:
@coderabbitai review

wolf31o2 (approved):

wolf31o2:
Merged. Thank you for the contribution.

PR #2119, opened 2026-08-29
fix(leiosfetch): bound abandoned request slot waits

### LINKED ISSUE #2118: fix(leiosfetch): bound abandoned request slots
## Summary

The Leios fetch client can leave a request slot occupied after a request
context expires before the peer responds. A later request then waits for the
slot indefinitely if the late response never arrives.

This can stall a caller that is trying to fetch endorser-block transactions,
even though the process and peer connection remain healthy.

## Expected behavior

A request that cannot receive a response within its caller-provided context
should return an error and must not permanently prevent later requests from
making progress.

Late responses must not be delivered to a subsequent request.

## Acceptance criteria

- A timed-out request cannot leave later requests blocked indefinitely.
- A late response is discarded or otherwise correlated with the original
  request and cannot be delivered to a newer request.
- Add a regression test covering timeout, missing response, and a subsequent
  request.
- Preserve protocol shutdown behavior and shared-bearer safety.

wolf31o2:
## Summary

Bound abandoned Leios-fetch request-slot waits so a canceled or timed-out request cannot block subsequent requests indefinitely. Late responses remain correlated to their original request, and the slot is released for failover.

Fixes #2118

## Validation

- `go test ./protocol/leiosfetch`
- `go test -race ./protocol/leiosfetch`
- `git diff --check`

---
## Summary by cubic
Binds abandoned leios-fetch request slot waits so a canceled or timed-out request can't block later requests indefinitely. The slot now waits up to one second for the late response, then returns `ErrRequestSlotAbandoned` so callers fail over to another connection. Late responses are still correlated to their original request and safely discarded. Fixes #2118.

- Acquirers already waiting when a request is abandoned are woken so they apply the bounded grace period instead of parking on the old drain channel.

<sup>Written for commit 4022542712f4b81ce50a5c25b2b0b3af3652655a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2119?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented leios-fetch requests from waiting indefinitely when a previous request times out without receiving a response.
  * Subsequent requests now fail with a clear error after a short grace period, allowing the connection to recover safely once the late response is handled.

### DIFF
--- a/protocol/leiosfetch/client.go
+++ b/protocol/leiosfetch/client.go
@@ -51,11 +51,12 @@ type Client struct {
 //     drained, so a late response to an abandoned request is never
 //     mis-delivered to a subsequent request.
 type requestSlot struct {
-	mu        sync.Mutex
-	waiter    chan protocol.Message
-	busy      bool
-	abandoned bool
-	drainedCh chan struct{}
+	mu              sync.Mutex
+	waiter          chan protocol.Message
+	busy            bool
+	abandoned       bool
+	drainedCh       chan struct{}
+	beforeDrainWait func() // test hook for an acquirer reaching the drain wait
 }
 
 const abandonedRequestWait = time.Second
@@ -92,7 +93,11 @@ func (s *requestSlot) acquire(
 			continue
 		}
 		drained := s.drainedCh
+		beforeDrainWait := s.beforeDrainWait
 		s.mu.Unlock()
+		if beforeDrainWait != nil {
+			beforeDrainWait()
+		}
 		select {
 		case <-drained:
 		case <-ctx.Done():
@@ -127,6 +132,13 @@ func (s *requestSlot) abandon(w chan protocol.Message) {
 	if s.waiter == w {
 		s.waiter = nil
 		s.abandoned = true
+		// Acquirers that started before this request was abandoned are
+		// waiting on the current channel through the non-abandoned path.
+		// Wake them so they recheck abandoned and use its bounded grace
+		// period. Keep the slot busy on a fresh channel until the late
+		// response drains it, preserving response correlation.
+		close(s.drainedCh)
+		s.drainedCh = make(chan struct{})
 	}
 	s.mu.Unlock()
 }
--- a/protocol/leiosfetch/client.go
+++ b/protocol/leiosfetch/client.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/blinklabs-io/gouroboros/protocol"
 	pcommon "github.com/blinklabs-io/gouroboros/protocol/common"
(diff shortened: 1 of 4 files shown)


arepala-uml (changes requested):
**Request changes: an already-waiting request can still block indefinitely.**

In `requestSlot.acquire`, a second request that arrives while the first request is active observes `busy == true` and `abandoned == false`, so it waits on `drainedCh`:

    drained := s.drainedCh
    s.mu.Unlock()
    select {
    case <-drained:
    case <-ctx.Done():
        return nil, ctx.Err()
    case <-done:
        return nil, protocol.ErrProtocolShuttingDown
    }

If the first request is subsequently abandoned, `requestSlot.abandon` sets `abandoned = true`, but it does not close or otherwise signal `drainedCh`:

    if s.waiter == w {
        s.waiter = nil
        s.abandoned = true
    }

Therefore, the second request never wakes up to re-check `s.abandoned` and never enters the new one-second timeout path. If it uses `context.Background()` and the peer never responds, it remains blocked indefinitely.

The existing test starts the second request only after the first request has already been abandoned, so it does not cover this ordering.

I reproduced the issue with this regression test:

    func TestBlockRequestAlreadyWaitingWhenFirstIsAbandoned(t *testing.T) {
        type requestResult struct {
            resp protocol.Message
            err  error
        }

        conversation := append(
            conversationHandshake,
            ouroboros_mock.ConversationEntryInput{
                ProtocolId:  leiosfetch.ProtocolId,
                MessageType: leiosfetch.MessageTypeBlockRequest,
            },
        )
        runTest(
            t,
            conversation,
            func(t *testing.T, oConn *ouroboros.Connection) {
                client := oConn.LeiosFetch().Client
                ctx1, cancel1 := context.WithTimeout(
                    context.Background(),
                    500*time.Millisecond,
                )
                defer cancel1()
                firstResult := make(chan requestResult, 1)
                go func() {
                    resp, err := client.BlockRequest(
                        ctx1,
                        pcommon.NewPoint(12345, []byte{0x01, 0x02}),
                    )
                    firstResult <- requestResult{resp: resp, err: err}
                }()

                // Start the second request while the first request still owns
                // the slot. It enters acquire while abandoned is still false.
                time.Sleep(100 * time.Millisecond)
                ctx2, cancel2 := context.WithCancel(context.Background())
                secondResult := make(chan requestResult, 1)
                go func() {
                    resp, err := client.BlockRequest(
                        ctx2,
                        pcommon.NewPoint(23456, []byte{0x03, 0x04}),
                    )
                    secondResult <- requestResult{resp: resp, err: err}
                }()

                first := <-firstResult
                require.ErrorIs(t, first.err, context.DeadlineExceeded)
                assert.Nil(t, first.resp)

                select {
                case second := <-secondResult:
                    require.ErrorIs(
                        t,
                        second.err,
                        leiosfetch.ErrRequestSlotAbandoned,
                    )
                    assert.Nil(t, second.resp)
                case <-time.After(1500 * time.Millisecond):
                    // Unblock the goroutine so the test can cleanly tear down
                    // after demonstrating that the bound was not applied.
                    cancel2()
                    <-secondResult
                    t.Error(
                        "second request remained blocked after the abandonment bound",
                    )
                }
                cancel2()
            },
        )
    }

Running:

    go test ./protocol/leiosfetch \
      -run TestBlockRequestAlreadyWaitingWhenFirstIsAbandoned \
      -count=1 -v

produces:

    === RUN   TestBlockRequestAlreadyWaitingWhenFirstIsAbandoned
        leiosfetch_test.go:325: second request remained blocked after the abandonment bound
    --- FAIL: TestBlockRequestAlreadyWaitingWhenFirstIsAbandoned (2.01s)
    FAIL

The transition to `abandoned` should wake callers that are already waiting for the slot, allowing them to re-check the state and enter the bounded abandonment wait. A regression test covering this ordering should also be included.

wolf31o2:
Addressed the waiting-request ordering case in current head `40225427`: abandoning the active request now wakes acquirers already blocked on `drainedCh`, which re-check the abandoned state and enter the bounded wait. The regression covers a second request started before abandonment. Please re-review the current head.

wolf31o2:
The current head `40225427` includes the waiting-request wake-up fix and its ordering regression coverage. The changes-requested review is from the prior head; please re-review the current head.

arepala-uml (approved):

PR #2122, opened 2026-08-29
fix(consensus): bound configuration and leadership searches

### LINKED ISSUE #2027: consensus: bound Byron configuration, VRF thresholds, and leadership search
Consensus helper bounds and failure checks must be explicit:

- `consensus/byron/byron.go` accepts K=0 and performs unchecked epoch/multiplier arithmetic.
- `consensus/threshold.go` reads raw TPraos VRF bytes without requiring the expected length.
- `consensus/leader.go` runs one VRF proof per slot without a maximum range or cancellation.

Acceptance criteria:

- Reject zero/overflowed configuration and short VRF output.
- Bound leadership work and return promptly on cancellation.
- Cover maximum, malformed, and cancelled inputs.

wolf31o2:
## Summary
- reject zero and overflowing Byron security parameters
- reject malformed VRF output lengths
- add a cancellable, bounded leadership-search API while preserving the legacy wrapper

## Validation
- go test ./consensus/... -count=1
- git diff --check

Fixes #2027

---
## Summary by cubic
Fixes #2027 by hardening consensus validation and preventing unbounded leadership searches.

- Rejects Byron security parameter K when zero or large enough to overflow slots per epoch.
- VRF outputs must be exactly 64 bytes; other lengths now return an error instead of being treated as non-leader.
- Adds `FindNextSlotLeadershipContext`, which supports context cancellation, caps each search at 1,000,000 slots, and stops at the requested upper bound.
- `FindNextSlotLeadership` keeps its signature but now delegates with a background context and the same range cap.

<sup>Written for commit d0fc406929615fa6d2798d8449451d13443649e7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2122?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/consensus/byron/byron.go
+++ b/consensus/byron/byron.go
@@ -161,8 +161,11 @@ func (c *ByronConfig) SlotLeader(slot uint64) (int, []byte) {
 //   - Genesis delegate key hashes (sorted for OBFT slot leader assignment)
 func NewByronConfigFromGenesis(genesis *ledgerbyron.ByronGenesis) (ByronConfig, error) {
 	// Validate security parameter K
-	if genesis.ProtocolConsts.K < 0 {
-		return ByronConfig{}, fmt.Errorf("invalid security parameter K: %d (must be non-negative)", genesis.ProtocolConsts.K)
+	if genesis.ProtocolConsts.K <= 0 {
+		return ByronConfig{}, fmt.Errorf("invalid security parameter K: %d (must be positive)", genesis.ProtocolConsts.K)
+	}
+	if uint64(genesis.ProtocolConsts.K) > math.MaxUint64/10 {
+		return ByronConfig{}, fmt.Errorf("invalid security parameter K: %d overflows slots per epoch", genesis.ProtocolConsts.K)
 	}
 
 	// Validate protocol magic fits in uint32
--- a/consensus/byron/validate_test.go
+++ b/consensus/byron/validate_test.go
@@ -19,6 +19,9 @@ import (
 	"crypto/rand"
 	"encoding/hex"
 	"errors"
+	"fmt"
+	"math"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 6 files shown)


arepala-uml (changes requested):
**Request changes: the invalid-security-parameter test is not portable to 32-bit platforms.**

The PR currently tests:

    for _, k := range []int{0, int(^uint(0) >> 1)} {
        genesis.ProtocolConsts.K = k
        _, err = NewByronConfigFromGenesis(&genesis)
        require.Error(t, err)
    }

On a 64-bit platform, the maximum `int` is large enough that `10*K` overflows `uint64`, so the second case correctly returns an error.

On a 32-bit platform, however:

    K = math.MaxInt32
      = 2,147,483,647

    10 * K
      = 21,474,836,470

    math.MaxUint64
      = 18,446,744,073,709,551,615

Therefore, `10*math.MaxInt32` fits safely in `uint64`. The production code correctly accepts it, but the existing test will incorrectly require an error.

I added this test beside the PR’s test to demonstrate the valid 32-bit boundary using the actual genesis parser and configuration constructor:

    func TestNewByronConfigFromGenesisAcceptsMaxInt32SecurityParameter(
        t *testing.T,
    ) {
        genesis, err := byron.NewByronGenesisFromReader(
            strings.NewReader(testByronGenesisJSON),
        )
        require.NoError(t, err)

        // On a 32-bit platform this is the largest value representable by the
        // genesis K field. Multiplying it by 10 is still safely within uint64,
        // so it must not be rejected by the slots-per-epoch overflow guard.
        genesis.ProtocolConsts.K = math.MaxInt32
        config, err := NewByronConfigFromGenesis(&genesis)
        require.NoError(t, err)
        require.Equal(t, uint64(10)*math.MaxInt32, config.SlotsPerEpoch)
    }

Running:

    go test ./consensus/byron \
      -run 'TestNewByronConfigFromGenesis(AcceptsMaxInt32|RejectsInvalid)SecurityParameter' \
      -count=1 -v

on the current 64-bit host produces:

    === RUN   TestNewByronConfigFromGenesisRejectsInvalidSecurityParameter
    --- PASS: TestNewByronConfigFromGenesisRejectsInvalidSecurityParameter (0.00s)
    === RUN   TestNewByronConfigFromGenesisAcceptsMaxInt32SecurityParameter
    --- PASS: TestNewByronConfigFromGenesisAcceptsMaxInt32SecurityParameter (0.00s)
    PASS

Both pass on 64-bit because the existing test uses `math.MaxInt64`, while the new test independently confirms that `math.MaxInt32` is valid.

On a 32-bit host, the existing expression:

    int(^uint(0) >> 1)

evaluates to `math.MaxInt32`. The existing test would then require that value to fail, while the regression test correctly requires it to succeed. This demonstrates the architecture-dependent failure.

Suggested correction:

    func TestNewByronConfigFromGenesisRejectsZeroSecurityParameter(
        t *testing.T,
    ) {
        // Test K == 0 on every architecture.
    }

    func TestNewByronConfigFromGenesisRejectsSecurityParameterOverflow(
        t *testing.T,
    ) {
        if strconv.IntSize < 64 {
            t.Skip(
                "an overflow-producing K cannot be represented by int",
            )
        }

        genesis.ProtocolConsts.K = math.MaxInt
        _, err := NewByronConfigFromGenesis(&genesis)
        require.Error(t, err)
    }

The new proof test is currently present as an uncommitted change in `consensus/byron/validate_test.go`. `git diff --check` passes.

wolf31o2:
Addressed in d0fc406. The zero and overflow cases are now architecture-specific, and a parser-backed MaxInt32 control verifies that `10*K` remains accepted on 32-bit targets.

arepala-uml (approved):

PR #2123, opened 2026-08-29
fix(blockfetch): validate streamed block ranges

### LINKED ISSUE #2002: blockfetch: validate streamed blocks against requested ranges
`protocol/blockfetch/blockfetch.go` accepts every streamed `MsgBlock` while in `StateStreaming` and never correlates the block point with the requested range. A peer can send a block outside the request and the client treats it as valid.

Acceptance criteria:

- Reject or terminate on a block outside the requested point range.
- Preserve valid pipelined ranges and add before/inside/after-range vectors.

wolf31o2:
## Summary

- correlate every streamed block with the requested inclusive point range
- reject an out-of-range block before pipeline or callback delivery
- add before, inside, after, boundary, and same-slot hash coverage

Fixes #2002

## Validation

- `go test -race ./protocol/blockfetch`
- `git diff --check`

---
## Summary by cubic
Validates every streamed block against the requested inclusive point range before pipeline or callback delivery. Out-of-range, duplicate, or non-contiguous blocks, and batches that end before the requested range end, now fail the request and tear down the protocol instead of being delivered. Fixes #2002.

<sup>Written for commit 1977ee6ec5f5f4e6f10e4f05b05abb526d5ba911. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2123?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/protocol/blockfetch/client.go
+++ b/protocol/blockfetch/client.go
@@ -15,6 +15,7 @@
 package blockfetch
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -70,6 +71,8 @@ const (
 // owns the next MsgStartBatch, MsgNoBlocks, MsgBlock, and MsgBatchDone.
 type rangeRequest struct {
 	id             uint64
+	start          pcommon.Point
+	end            pcommon.Point
 	protocol       *protocol.Protocol
 	delivery       requestDelivery
 	pipelined      bool
@@ -720,6 +723,8 @@ func (c *Client) sendRequestRange(
 		c.nextRequestId++
 		req = &rangeRequest{
 			id:            c.nextRequestId,
+			start:         start,
+			end:           end,
 			protocol:      proto,
 			delivery:      delivery,
 			pipelined:     pipelined,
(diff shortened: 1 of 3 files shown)


arepala-uml (changes requested):
## Review: Changes requested

The new range validation checks each received block independently, but `handleBatchDone()` does not verify that the requested inclusive range was actually delivered.

A peer can respond to a non-empty request with:

1. `MsgStartBatch`
2. No blocks
3. `MsgBatchDone`

The client then completes the request successfully and calls `RangeDoneFunc` with a `nil` error. Partial batches, duplicate blocks, and out-of-order interior blocks are similarly not detected.

The following regression test proves the incomplete-range case. Add it to `protocol/blockfetch/client_queue_internal_test.go`:

    func TestBatchDoneRejectsIncompleteRequestedRange(t *testing.T) {
        var completionErr error
        c := newQueueTestClient(&Config{
            RequestPipelining: true,
            RangeDoneFunc: func(_ CallbackContext, err error) error {
                completionErr = err
                return nil
            },
        })

        req := c.appendTestRequest(1024)
        req.start = pcommon.NewPoint(100, []byte("start"))
        req.end = pcommon.NewPoint(200, []byte("end"))

        require.NoError(t, c.handleStartBatch())

        err := c.handleBatchDone()
        require.Error(
            t,
            err,
            "BatchDone must reject a batch that never delivered the requested range",
        )
        require.Error(
            t,
            completionErr,
            "RangeDoneFunc must not report an incomplete range as successful",
        )
    }

Run with:

    go test ./protocol/blockfetch \
        -run '^TestBatchDoneRejectsIncompleteRequestedRange$' \
        -count=1

Current result:

    --- FAIL: TestBatchDoneRejectsIncompleteRequestedRange (0.00s)
        Error: An error is expected but got nil.
        Messages: BatchDone must reject a batch that never delivered the requested range
    FAIL

This confirms that a non-empty inclusive range can be completed successfully without delivering any blocks.

The request should track the last successfully delivered point. Before accepting `BatchDone`, the client should verify that the requested end point was delivered. It should also reject duplicate and non-increasing points within the streamed batch.

**Recommendation: Do not approve until incomplete and incorrectly ordered range responses are rejected and covered by handler-level tests.**

wolf31o2:
Commit 1977ee6 requires the first and last delivered blocks to match the requested inclusive endpoints and verifies every interior block through its previous-hash link. Empty, truncated, duplicate, skipped, and backward batches now fail the request instead of reporting successful completion. Handler-level coverage exercises each rejection and a complete linked range.

arepala-uml (approved):

PR #2129, opened 2026-08-29
fix(ledger): require certificate authorization witnesses

### LINKED ISSUE #2127: Require witnesses for Conway certificate and voter authorization
### Scope
Ensure Conway certificate and voter state transitions require the complete, correct key witness set before validation succeeds.

### Acceptance criteria
- Require key witnesses for every key-credential certificate, including pool operators and every pool owner.
- Require key-hash committee, DRep, and SPO voter witnesses across all supported voter forms.
- Traverse credentials through one shared path so key and script credentials cannot drift.
- Keep script credentials on the script-validation path; do not accept a key witness as a substitute.
- Add independent positive and negative tests for each certificate family, pool operator/owner authorization, and every voter type.
- Add a completeness table covering every common certificate type.
- Run focused normal and race tests plus the repository's required checks.

### Notes
This is a consensus/ledger correctness change. Preserve compatibility with supported eras and document any intentionally excluded certificate forms.

wolf31o2:
## Summary

Require authorization witnesses for certificate and governance-voter state changes, taken from `getVKeyWitnessTxCert` and `getScriptWitnessTxCert` in cardano-ledger's Shelley and Conway `TxCert.hs`.

## Changes

- Collect required key credentials from certificates, pool authorization, and governance voters through one shared traversal, `forEachCertificateCredential`, which names all 19 certificate forms and fails closed on an unhandled one.
- Share that traversal with script-witness validation, so a credential that authorizes nothing creates no script purpose and no redeemer requirement.
- Index reward redeemers on cardano-ledger's `Withdrawals` key order, the derived `Ord` on `AccountAddress` = `(Network, Credential)` with `ScriptHashObj` before `KeyHashObj`, replacing reward-address byte order. Address headers are `0xE_` for a key hash and `0xF_` for a script hash, which inverts it.
- Evaluate native scripts reachable through reference inputs and consumed outputs from Allegra onward via the shared `common.FirstInvalidNativeScriptIn`. Allegra and Mary have no `ScriptRef`, so the behavioral change is Babbage onward.
- Add MIR genesis-delegate quorum validation as exported API, left unregistered in every era's rule list pending a consumer for `common.GenesisDelegationState` (blinklabs-io/dingo#3748). Nothing enforces it yet.
- Add positive and negative coverage for certificate, voter, withdrawal-index, and MIR authorization, resolved out of each era's production `UtxoValidationRules`.

## Validation

- `go build ./...`, `go vet ./ledger/...`, `go test ./ledger/...`, `go test -race ./ledger/...`, `golangci-lint run ./ledger/...` (0 issues), `gofmt -s -l ledger/` (clean).
- Amaru conformance: 315/315 vectors on `36f10e2`. The corpus has no MIR certificate, no tag-7 registration whose credential abstains, and no transaction mixing withdrawal credential types, so it does not discriminate those three decisions; `ledger/common/rules_test.go` is their only coverage.

Follow-up: `blinklabs-io/gouroboros#2167` tracks duplicate input resolution between `UtxoValidateScriptWitnesses` and `UtxoValidateNativeScripts`.

Closes #2127


ledger/common/rules.go
```diff
@@ -160,6 +161,168 @@ func (e MissingRequiredVKeyWitnessForSignerError) Error() string {
...
+		if c == nil {
+			return typedNil("pool retirement certificate")
+		}
+		return nil
+	case *GenesisKeyDelegationCertificate:
+		if c == nil {
+			return typedNil("genesis key delegation certificate")
+		}
+		return nil
+	case *MoveInstantaneousRewardsCertificate:
```
arepala-uml, line 433: ## Blocking issue: MIR certificates are accepted without genesis-delegate authorization

The new certificate-authorization logic explicitly treats `MoveInstantaneousRewardsCertificate` as requiring no witness:

- `ledger/common/rules.go:314`
- `ledger/common/rules_test.go:517`

The accompanying comment claims that Shelley through Babbage enforce the required genesis-delegate quorum in a separate UTXOW predicate. However, no such validation appears to exist in the repository.

As a result, the Shelley witness-validation entry point accepts an MIR certificate with no genesis-delegate witnesses.

### Regression test

I added the following test to `ledger/common/rules_test.go`:

```go
// MIR certificates are authorized by a quorum of the currently delegated
// genesis keys. The Shelley UTXOW witness rule must therefore reject an MIR
// certificate when the transaction carries no vkey witnesses at all.
func TestMIRCertificateRequiresGenesisDelegateQuorum(t *testing.T) {
	cert := &common.MoveInstantaneousRewardsCertificate{
		CertType: uint(common.CertificateTypeMoveInstantaneousRewards),
	}
	tx := mockledger.NewTransactionBuilder().WithCertificates(cert)
	ls := mockledger.NewLedgerStateBuilder().Build()

	require.Error(
		t,
		shelley.UtxoValidateRequiredVKeyWitnesses(tx, 0, ls, nil),
		"an MIR certificate without genesis-delegate witnesses must be rejected",
	)
}
```

### Reproduction

Run:

```bash
go test ./ledger/common \
  -run '^TestMIRCertificateRequiresGenesisDelegateQuorum$' \
  -count=1
```

Current result:

```text
--- FAIL: TestMIRCertificateRequiresGenesisDelegateQuorum (0.00s)
    rules_test.go:539:
        Error:       An error is expected but got nil.
        Test:        TestMIRCertificateRequiresGenesisDelegateQuorum
        Messages:    an MIR certificate without genesis-delegate witnesses must be rejected
FAIL
```

The failure demonstrates that `shelley.UtxoValidateRequiredVKeyWitnesses` currently accepts an unauthorized MIR certificate.

### Requested change

Implement the Shelley-through-Babbage MIR genesis-delegate quorum witness validation that the new comment says is handled separately, and retain this test as regression coverage.

The tests should ultimately verify:

1. No genesis-delegate witnesses are rejected.
2. An insufficient number of delegate witnesses is rejected.
3. A valid quorum of current genesis-delegate witnesses is accepted.
4. Unrelated or obsolete genesis-delegate witnesses do not count toward the quorum.

wolf31o2: Confirmed and fixed in 14609bc. There was no equivalent of the upstream `validateMIRInsufficientGenesisSigs` UTXOW predicate, so an MIR certificate passed witness validation unauthorized.

`common.ValidateMIRGenesisQuorum` counts distinct signatures from currently delegated genesis keys and requires the update quorum. It is exposed as `UtxoValidateMIRGenesisQuorum` and appended to the Shelley, Allegra, Mary, Alonzo, and Babbage rule lists. Appending rather than inserting keeps existing rule indices stable, which dingo pins by index and name. Conway and Dijkstra are unchanged because they expunge MIR.

The genesis delegation mapping and update quorum are not on `common.LedgerState`, so they arrive through `GenesisDelegationState`, an optional capability following the existing `DRepDelegationState` pattern. A state that does not implement it fails closed with `GenesisDelegationStateUnavailableError`, and only when the transaction carries an MIR certificate; other transactions are unaffected.

`TestValidateMIRGenesisQuorum` covers the four requested cases, plus the unavailable-state and repeated-signature paths. Seven of its subtests fail with the predicate stubbed to `return nil`.

Follow-up for dingo: it must implement `GenesisDelegationState` before consuming a release containing this, or Shelley-through-Babbage replay will reject historical MIR transactions. The inputs are already stored there (`shelleyGenesis.GenDelegs`, `UpdateQuorum`, `database/models/genesis_delegation.go`).

wolf31o2:
Merged current main and implemented the MIR genesis-delegate quorum.

- `common.ValidateMIRGenesisQuorum` requires the update quorum of distinct signatures from currently delegated genesis keys, exposed as `UtxoValidateMIRGenesisQuorum` and appended to the Shelley, Allegra, Mary, Alonzo, and Babbage rule lists. Appending keeps existing rule indices stable.
- `common.GenesisDelegationState` is a new optional ledger-state capability, following `DRepDelegationState`. A state that does not implement it fails closed, and only for transactions carrying an MIR certificate.
- `TestValidateMIRGenesisQuorum` covers absent, insufficient, sufficient, obsolete, and repeated genesis signatures. Seven subtests fail with the predicate stubbed out.

Merge resolution kept this branch's map-based witness collection and certificate/voter authorization, and adopted main's reward-account credential validation and withdrawal-address checks.

Dingo must implement `GenesisDelegationState` before consuming a release with this change: blinklabs-io/dingo#3748.

`go build`, `go vet`, `gofmt`, `go test ./ledger/...`, and `go test -race` on the affected era packages pass locally.

ledger/shelley/rules.go
```diff
@@ -51,6 +51,7 @@ var UtxoValidationRules = []common.UtxoValidationRuleFunc{
 	UtxoValidateMaxTxSizeUtxo,
 	UtxoValidateDelegation,
 	UtxoValidateWithdrawals,
+	UtxoValidateMIRGenesisQuorum,
```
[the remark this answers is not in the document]
wolf31o2: Real, and deliberate, but it needs a human call rather than another flip.

The rule fails closed only for a transaction that carries an MIR certificate, and MIR exists only in Shelley through Babbage, so ordinary operation is unaffected. What is affected is replaying that history on a ledger state that does not implement `GenesisDelegationState`.

This follows the pattern already in the repo: `UtxoValidateWithdrawals` returns `DRepDelegationStateUnavailableError` when a state does not implement `DRepDelegationState`. The alternative is to skip the quorum when the capability is absent, which is exactly the hole the review that requested this rule asked to close, so silently failing open would undo it.

Dingo is tracked in blinklabs-io/dingo#3748 and already stores the inputs. If the preference is to hold this rule until consumers implement the capability, that is a reasonable call, but it is a release-sequencing decision rather than a code fix, so leaving it for @arepala-uml to weigh against the finding that prompted the rule.

ledger/common/rules.go
```diff
@@ -207,15 +370,65 @@ func ValidateRequiredVKeyWitnesses(tx Transaction) error {
...
+		}); err != nil {
+			return err
+		}
+		switch c := cert.(type) {
+		case *PoolRegistrationCertificate:
+			if c == nil {
+				continue
+			}
+			required[Blake2b224(c.Operator)] = struct{}{}
+			for _, owner := range c.PoolOwners {
```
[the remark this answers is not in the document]
wolf31o2: Not correct as stated, so no change.

Pool owner witnesses are required in every era from Shelley onward, not just pre-Conway. In cardano-ledger the requirement comes from the `owners` term of `witsVKeyNeeded`, which folds `ppOwners` over the registration certificates. Conway builds its own needed-witness set on `getShelleyWitsVKeyNeededNoGov`, which still contains that term, and adds only the governance voter authors on top. There is no era override that drops pool owners.

That is also what the certificate-authorization comment on `forEachCertificateCredential` records for tag 3: the operator is collected separately and every owner is collected by the independent owners term.

All 315 Amaru conformance vectors pass on this head.

ledger/common/errors.go
```diff
@@ -269,6 +269,23 @@ func (e ExtraneousRedeemerError) Error() string {
 	)
 }
 
+// MissingRedeemerForScriptError indicates that an available Plutus script is
+// required by a concrete purpose but the witness set has no redeemer at that
+// purpose's exact tag and canonical index.
+type MissingRedeemerForScriptError struct {
```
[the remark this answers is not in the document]
wolf31o2: Confirmed and fixed in 3ff6a51 by deleting the Conway copy.

It was already unreferenced before this branch, and nothing in the module ever returned it, so `errors.As` against it could never match. The two shapes differ, `Tag`/`Index` versus a `RedeemerKey`, so a type alias was not available; the common type is the one every call site uses.

ledger/common/rules.go
```diff
@@ -253,260 +518,434 @@ func ValidateUnsupportedPlutusExecution(tx Transaction, era string) error {
...
+		return err
+	}
+	for required := range requirements.required {
+		if _, ok := requirements.available[required]; !ok {
+			return MissingScriptWitnessesError{ScriptHash: required}
 		}
-		if _, ok := optionalScriptHashes[provided]; !ok {
+	}
+	for provided := range requirements.explicit {
+		if _, ok := requirements.required[provided]; !ok {
```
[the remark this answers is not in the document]
wolf31o2: Not correct as stated, so no change.

Upstream treats a legacy type-0 registration as needing no script. Conway's `getScriptWitnessTxCert` returns `Nothing` for `ConwayRegCert`, so the credential contributes nothing to `scriptsNeeded`, and a script supplied for it is genuinely extraneous rather than optional. Rejecting it matches `ExtraneousScriptWitnessesUTXOW`.

This is a deliberate tightening in this branch relative to main, which tracked such scripts as optional, and the reasoning is recorded inline at the extraneous check.

ledger/common/rules.go
```diff
@@ -160,6 +161,168 @@ func (e MissingRequiredVKeyWitnessForSignerError) Error() string {
...
+		}
+		if c.StakeCredential == nil {
+			return typedNil("stake delegation credential")
+		}
+		return visitCredential(*c.StakeCredential, true)
+	case *RegistrationCertificate:
+		if c == nil {
+			return typedNil("registration certificate")
+		}
+		return visitCredential(c.StakeCredential, true)
```
chrisguiney, line 362: **Blocking.** I think this should be `false`, and I'd point at your own reasoning on the extraneous-script thread as the reason.

CDDL type 0 (`stake_registration`) and type 7 (`reg_cert`) both decode to the same cardano-ledger constructor:

```haskell
ConwayRegCert !StakeCredential !(StrictMaybe Coin)
```

Type 0 is `SNothing`, type 7 is `SJust coin`. Both `getVKeyWitnessConwayTxCert` and `getScriptWitnessConwayTxCert` match `ConwayRegCert{}` with a wildcard on both fields and return `Nothing`.

That's the same fact you cite to justify rejecting a type-0 script as extraneous:

> Conway's `getScriptWitnessTxCert` returns `Nothing` for `ConwayRegCert`, so the credential contributes nothing to `scriptsNeeded`

One constructor can't both contribute nothing to `scriptsNeeded` and require a script witness. Whichever way it resolves, types 0 and 7 need to agree — and right now the doc comment above groups 7 with "the certificate credential named below authorizes it" while 0 is carved out.

If this is wrong the blast radius is large: gouroboros rejects transactions cardano-node accepts, which is a chain split for dingo. There's a second-order effect too — `collectTransactionScriptRequirements` gives a script-credential type-7 a `RedeemerTagCert` requirement, so a Plutus-script registration now needs a redeemer the node would treat as extraneous.

The vectors can't settle this. I scanned all 315: there are 106 type-7 certs across 91 vectors, and in all 103 key-credential cases the credential signs, so the corpus passes either way. `TestValidateRequiredVKeyWitnessesExplicitRegistration` and `TestValidateScriptWitnessesExplicitRegistration` assert the new behavior rather than derive it.

I'm reasoning from the constructor structure rather than the source in front of me, so I'd want this confirmed directly against `Cardano.Ledger.Conway.TxCert` before merge — but given nothing in the test corpus discriminates, it does need a direct check rather than a green run.

ledger/common/rules.go
```diff
@@ -253,260 +518,434 @@ func ValidateUnsupportedPlutusExecution(tx Transaction, era string) error {
...
 
-	// Collect script hashes required by withdrawals
+	withdrawals := make([]*Address, 0, len(tx.Withdrawals()))
 	for addr := range tx.Withdrawals() {
+		if _, err := addr.RewardAccountCredential(); err != nil {
+			return ret, err
+		}
+		withdrawals = append(withdrawals, addr)
+	}
+	sort.Slice(withdrawals, func(i, j int) bool {
```
chrisguiney, line 933: **Blocking.** Sorting withdrawals by raw reward-address bytes gives a different order than cardano-ledger, so `RedeemerTagReward` indices can diverge from the node's.

Measured on this branch, same hash under both credential types:

```
key    addr: e1aaaa...
script addr: f1aaaa...
bytes.Compare(key, script) = -1   -> key sorts FIRST
```

cardano-ledger indexes `Withdrawals` by `RewardAccount` Ord, which is `(Network, Credential)`, and `Credential`'s derived Ord is `ScriptHashObj | KeyHashObj` — so **script sorts first**. The address header puts keyhash at `0xE_` and scripthash at `0xF_`, which inverts it.

Any transaction carrying both a key-credential and a script-credential withdrawal therefore gets a different index for the script withdrawal than the node assigns, and the redeemer check below fails it with `MissingRedeemerForScriptError`.

What makes me confident this is an oversight rather than a deliberate choice: `voterPurposeOrder` a few lines down encodes the very same rule correctly — CC-script 0, CC-key 1, DRep-script 2, DRep-key 3, SPO 4, matching `CommitteeVoter < DRepVoter < StakePoolVoter` with Script < Key inside each. Withdrawals just went through address bytes instead.

Sorting on `(network, credType, credential)` — script before key — would line them up. The corpus has 9 script-withdrawal transactions but 0 mixed, so nothing here catches it.

ledger/shelley/rules.go
chrisguiney: You asked for a human call on Cubic's fail-closed thread rather than another flip, so here's mine: **I'd sequence this behind dingo#3748** rather than land it now.

Two things push me that way beyond the wiring concern Cubic raised.

First, coverage. There are **zero** MIR certificates across all 315 Amaru vectors, and the conformance harness runs the Conway rule list, which doesn't include this rule. So "all 315 vectors pass" carries no information about `ValidateMIRGenesisQuorum` — it is entirely unexercised by the corpus. The unit tests in `rules_test.go` are the only coverage, and they exercise the quorum arithmetic rather than a real Shelley-era transaction.

Second, the failure mode is silent and total. Any consumer replaying Shelley-through-Babbage history hits `GenesisDelegationStateUnavailableError` on the first MIR certificate, and mainnet history contains them. That turns a `gouroboros` version bump into a hard stop for a syncing node, with the fix living in a different repository.

I agree with your reasoning that failing open would undo the hole this rule was added to close, and the `DRepDelegationState` precedent is a fair analogy. My concern is purely ordering: land `GenesisDelegationState` and `ValidateMIRGenesisQuorum` as available API, let dingo#3748 implement it, then add the rule-list entries. That gets the same end state without a window where a released gouroboros can't replay mainnet.

ledger/allegra/rules.go
```diff
@@ -249,9 +250,46 @@ func UtxoValidateNativeScripts(
 	ls common.LedgerState,
 	pp common.ProtocolParameters,
 ) error {
-	if scriptHash, failed := common.FirstInvalidNativeScript(tx, slot); failed {
-		return NativeScriptFailedError{ScriptHash: scriptHash}
+	witnesses := tx.Witnesses()
+	nativeScripts, err := common.NativeScriptsForValidation(tx, ls)
```
chrisguiney, line 320: This rewrite isn't mentioned in the PR description, and it does two things worth separating.

The good part: because Mary delegates here, and Alonzo to Mary, and Babbage to Alonzo, this is what gives **Babbage** evaluation of native scripts supplied via reference inputs. That's a real correctness fix — it just arrives in the Allegra function, which is a surprising place to find it, and reviewers won't know to look for it from the description.

The duplication part: the block from here through the evaluation loop is a line-for-line copy of `common.FirstInvalidNativeScript` (`ledger/common/script.go:753`) — same key-hash collection including bootstrap witnesses, same `^uint64(0)` validity-end default, same evaluation loop. That same block now also exists in `conway/rules.go:2987` and `dijkstra/rules.go:1616`, so three copies of it, while **Shelley still calls the helper** and so never picks up any of this.

Giving `FirstInvalidNativeScript` a script-list parameter (or adding a sibling that takes one) would let all four eras share it and would close the Shelley gap in the same move.

Worth noting for Allegra and Mary specifically: neither era has `ScriptRef`, so `NativeScriptsForValidation` can't find anything the witness set didn't already carry there. The behavior change is entirely for Babbage.

ledger/conway/rules.go
```diff
@@ -2985,26 +2985,27 @@ func UtxoValidateNativeScripts(
 	pp common.ProtocolParameters,
 ) error {
 	witnesses := tx.Witnesses()
-	if witnesses == nil {
-		return nil
+	nativeScripts, err := common.NativeScriptsForValidation(tx, ls)
```
chrisguiney: Performance: this call sits *before* the `len(nativeScripts) == 0` early return, so every transaction now pays for it — including ordinary payment transactions with no scripts at all, which previously returned immediately once `witnesses.NativeScripts()` came back empty.

`NativeScriptsForValidation` calls `collectTransactionScriptRequirements`, which resolves every input and every reference input through `ls.UtxoById` and then runs four sorts. And since `UtxoValidateScriptWitnesses` is also in this rule list and makes the same call, **every transaction now resolves its full input set twice** where it used to do so once.

For a node validating every transaction in every block that's a meaningful regression on the common path. Two options that would recover most of it: hoist the cheap `witnesses.NativeScripts()` / `ScriptRef`-presence check ahead of the resolution so script-free transactions bail early, or compute the requirements once per transaction and share them between the two rules.

Same applies to `dijkstra/rules.go:1616` and `allegra/rules.go:254`.

ledger/common/rules.go
```diff
@@ -253,260 +518,434 @@ func ValidateUnsupportedPlutusExecution(tx Transaction, era string) error {
...
+// redeemer pointer for every Plutus purpose. Native purposes reject a redeemer
+// at that pointer and are evaluated separately, including reference scripts.
+func ValidateScriptWitnesses(tx Transaction, ls LedgerState) error {
+	if err := ValidateWithdrawalAddresses(tx.Withdrawals()); err != nil {
+		return err
+	}
+	if !tx.IsValid() {
+		return nil
+	}
+	requirements, err := collectTransactionScriptRequirements(tx, ls)
```
chrisguiney, line 1067: Undocumented behavior change: `main` had an explicit early return here —

```go
if ls == nil {
    return nil
}
```

— which this drops. With a nil ledger state, `collectTransactionScriptRequirements` skips input resolution but still collects script requirements from certificates, withdrawals, voters, mint policies, and proposal procedures. Those requirements are then enforced against an `available` set that can only ever contain witness-set scripts, so anything that would have been satisfied by a reference script now fails as `MissingScriptWitnessesError`.

That may well be the intended tightening — a nil state arguably shouldn't silently disable script validation. But it's a semantic change for any consumer passing nil (the tests in this PR pass `nil` to the vkey rule, so the pattern is in use), and it isn't in the PR description. Worth either restoring the guard or calling it out explicitly as intentional.

chrisguiney (changes requested):
## Summary

The shared-traversal design here is the right answer to #2127. `forEachCertificateCredential` naming all 19 certificate forms with typed-nil handling and a fail-closed default is exactly the "one shared path" and "completeness table" the issue asked for, and it reads well. Your rebuttal to Cubic on pool owners is also correct — the `owners` term of `witsVKeyNeeded` survives into Conway via `getShelleyWitsVKeyNeededNoGov`, so requiring owner witnesses in the common path is right.

I'm requesting changes on two consensus-critical items, plus some findings about what the validation evidence does and doesn't cover.

## What I checked

`go build`, `go vet`, `go test ./...`, `go test -race ./ledger/...`, and `golangci-lint run ./ledger/...` (0 issues) all pass on this branch. I also reproduced your conformance claim by pointing ouroboros-mock at this branch with a `replace` directive: **315/315 vectors pass.**

Then I instrumented the corpus to see what those vectors actually exercise:

```
cert type counts: {0:457, 1:6, 2:18, 3:110, 4:5, 7:106, 8:10, 9:106,
                   10:2, 11:89, 12:9, 13:3, 14:154, 15:10, 16:110, 17:8, 18:3}
type-7 reg_cert, key credential DID sign:            103
type-7 reg_cert, credential did NOT sign:              0
MIR certificates (type 6):                             0
genesis key delegation (type 5):                       0
txs with mixed key+script withdrawals:                 0
```

So the corpus passes identically whether or not the three most consensus-sensitive new decisions are correct: it never contains a type-7 registration whose credential abstains, never contains an MIR certificate, and never mixes withdrawal credential types. The 315/315 result is real, but it isn't evidence for those three. Details inline.

## Blocking

1. **Type-7 `RegistrationCertificate` requires witnesses** — this contradicts the cardano-ledger fact you cite on the extraneous-script thread. `rules.go:243`.
2. **Withdrawal redeemer indices use address-byte order** where cardano-ledger uses `Credential` order (script before key). `rules.go:785`.

## Non-blocking

Native-script duplication across three eras with Shelley left on the old helper, an extra full UTxO resolution on every transaction including script-free ones, the undocumented `ls == nil` removal, and my human call on the MIR merge ordering you asked for. All inline.

One trivial nit with no inline anchor: `gofmt -s` rewrites `&common.Address{}` to `{}` at `ledger/common/rules_test.go:1252`. CI's golangci-lint doesn't run `-s` so it's green, but `make format` would touch it.

## Scope note

#2127 is about certificate and voter witnesses. The MIR quorum, the `ValidateScriptWitnesses` redeemer-pointer rewrite, and the Allegra native-script rewrite all go beyond it. The MIR work came from review feedback, which is fine — but the redeemer-pointer and Allegra rewrites aren't mentioned in the PR description at all, and both change consensus behavior. Worth listing them so reviewers know to look.

ledger/common/rules.go
wolf31o2: Confirmed, fixed, and the corpus does discriminate after all, just not the way either of us expected.

Type 7 now visits its credential with `requiresWitness = false`, matching type 0. The doc comment groups them as `0 and 7`.

Making that change turned two vectors red, both `extraneous script witness`: `DELEG/Register_stake_credential/When_already_already_registered` and `DELEG/Unregister_stake_credentials/When_registered`. They supply the script for a script-credential registration and expect success. So the constructor argument is right that registration creates no script purpose, but rejecting a supplied script as extraneous is wrong, which means my rebuttal to Cubic on that thread was also wrong.

The resolution that satisfies both: a certificate that authorizes nothing records its script hash as optional. Not required, not a redeemer purpose, and not extraneous when present. `transactionScriptRequirements` grows an `optional` set, and the extraneous check skips it. That also removes the `RedeemerTagCert` requirement a script-credential type 7 was getting.

315/315 vectors pass. The three tests that asserted the old behaviour now assert this.

ledger/common/rules.go
wolf31o2: Confirmed and fixed. Withdrawals are now sorted on `(network, credential-type, hash)` with script ordered before key, rather than on `addr.Bytes()`.

Your read of the cause matches what the code shows: `voterPurposeOrder` had the rule right and the withdrawal path went through address bytes instead. The comment at the sort now records the `RewardAccount` Ord derivation and why the `0xE_`/`0xF_` headers invert it.

ledger/shelley/rules.go
wolf31o2: Taking your sequencing. `UtxoValidateMIRGenesisQuorum` is out of all five era rule lists; `common.GenesisDelegationState` and the rule itself ship as available API with a comment saying it is unregistered and pointing at blinklabs-io/dingo#3748.

Your coverage point is the one that settles it: the conformance harness runs the Conway rule list and the corpus has no MIR certificates, so 315/315 was never evidence about this rule. That should have been in my reply rather than cited as support.

ledger/allegra/rules.go
wolf31o2: Fixed for the duplication, and I want to disagree slightly on the Shelley half.

`FirstInvalidNativeScript` now delegates to `FirstInvalidNativeScriptIn`, which takes the script list. Allegra and Conway call it with the resolved list from `NativeScriptsForValidation`, so the copied key-hash collection, validity-interval defaulting, and evaluation loop are gone from both.

Dijkstra keeps its own loop: it calls `EvaluateWithGuards` with `nativeScriptGuardCredentials`, which the shared helper has no notion of. Folding that in would mean pushing guards into `common`, which seemed worse than one specialised loop.

Shelley stays on `FirstInvalidNativeScript`. A Shelley output cannot carry a `ScriptRef`, so routing it through `NativeScriptsForValidation` would resolve the input set to find nothing the witness set did not already have. That is the same reasoning you gave for Allegra and Mary, and it makes the Shelley call the correct one rather than a gap.

ledger/conway/rules.go
wolf31o2: Confirmed and fixed. `common.TransactionMayCarryNativeScript` runs first in Allegra, Conway, and Dijkstra: it reads the witness set and the produced outputs only, no ledger state, and returns false for a transaction with no native script witness, no native reference script on any output it produces, and no inputs to resolve. A script-free transaction no longer resolves its input set here at all.

Sharing one requirement computation between this rule and `UtxoValidateScriptWitnesses` would recover the rest, but that needs per-transaction caching across rules and seemed like a larger change than this review should carry.

ledger/common/rules.go
wolf31o2: Restored rather than kept. `ValidateScriptWitnesses` returns early again on a nil ledger state, with a comment recording why: without state a reference script cannot be resolved, so every requirement it would have satisfied is reported missing.

You are right that the tightening is arguably correct, but it was a side effect rather than a decision, and this PR is already carrying more consensus change than #2127 asked for.

wolf31o2:
Both blocking items fixed in 1de43f7, and one of them showed my earlier rebuttal to Cubic was wrong.

Type 7 no longer requires a witness. Making that change turned two vectors red on `extraneous script witness`, both supplying a script for a script-credential registration and expecting success. So registration creates no script purpose, as the constructor argument says, but a supplied script is accepted rather than extraneous. Registration script hashes are now tracked as optional, which satisfies both facts. 315/315 pass.

Withdrawal redeemer indices now sort on `(network, credential-type, hash)` with script first.

Taking your sequencing call on MIR: the rule and `GenesisDelegationState` ship as API, out of every era rule list, behind blinklabs-io/dingo#3748.

Non-blocking: native-script evaluation is shared through `FirstInvalidNativeScriptIn` for Allegra and Conway; Dijkstra keeps its own loop because it evaluates with guard credentials, and Shelley stays on the witness-only helper because a Shelley output cannot carry a `ScriptRef`. A cheap `TransactionMayCarryNativeScript` check now runs before input resolution, so script-free transactions no longer resolve their input set there. The `ls == nil` early return is restored. `gofmt -s` applied.

The description now lists the redeemer-pointer and native-reference-script work as consensus changes beyond #2127, and records which of the three decisions the corpus does and does not discriminate on.

ledger/shelley/rules.go
```diff
@@ -419,6 +419,23 @@ func UtxoValidateRequiredVKeyWitnesses(
 	return common.ValidateRequiredVKeyWitnesses(tx)
 }
 
+// UtxoValidateMIRGenesisQuorum ensures a move instantaneous rewards
+// certificate is authorized by a quorum of the current genesis delegates.
+//
+// Not yet registered in any era's UtxoValidationRules. The rule fails closed
```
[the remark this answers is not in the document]
wolf31o2: This is the deliberate outcome of a human review decision, so no change.

@chrisguiney reviewed exactly this trade-off and called for sequencing: land `GenesisDelegationState` and `ValidateMIRGenesisQuorum` as API, let a consumer implement the capability, then register the rule. The reason is that the rule fails closed for a state that cannot answer the query, so registering it now would stop a syncing node on the first MIR certificate in Shelley-through-Babbage history, with the fix living in another repository.

Tracked in blinklabs-io/dingo#3748. The rule comment in `ledger/shelley/rules.go` records that it is unregistered and why.

ledger/conway/rules.go
```diff
@@ -2984,46 +2984,24 @@ func UtxoValidateNativeScripts(
...
-		return nil
-	}
-
-	nativeScripts := witnesses.NativeScripts()
-	if len(nativeScripts) == 0 {
+	// Resolving the full input set is expensive, and UtxoValidateScriptWitnesses
+	// in this same rule list resolves it too. Skip it entirely for a
+	// transaction that carries no native script and produces no reference
+	// script, which is the ordinary payment case.
+	if !common.TransactionMayCarryNativeScript(tx) {
```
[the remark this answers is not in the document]
wolf31o2: Correct, and removed rather than redesigned.

The predicate ended in `len(tx.ReferenceInputs()) > 0 || len(tx.Inputs()) > 0`, which is true for any real transaction, so it never skipped anything and the comment above it was wrong.

Redesigning it is not available either: a consumed input can carry a reference script that satisfies a native purpose, and deciding that requires resolving the input. So there is no sound pre-resolution answer, and the honest fix is to drop the guard. The double resolution @chrisguiney raised remains, and the real fix for it is sharing one requirement computation between this rule and `UtxoValidateScriptWitnesses`.

The deduplication half of that review stands: Allegra and Conway share `FirstInvalidNativeScriptIn`.

ledger/common/rules.go
```diff
@@ -253,260 +522,479 @@ func ValidateUnsupportedPlutusExecution(tx Transaction, era string) error {
...
+	// Reward redeemer indices follow cardano-ledger's Withdrawals key order,
+	// which is RewardAccount Ord: (Network, Credential). Credential's derived
+	// Ord puts ScriptHashObj before KeyHashObj, so a script credential sorts
+	// ahead of a key credential with the same hash. Address bytes invert that,
+	// because the reward header is 0xE_ for a key hash and 0xF_ for a script
+	// hash, so sorting on them would give a script withdrawal a different
+	// index than the node assigns. voterPurposeOrder below encodes the same
+	// script-before-key rule.
+	type withdrawalKey struct {
+		address    *Address
```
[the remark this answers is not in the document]
wolf31o2: Fixed. `address` is gone from `withdrawalKey` and from the literal; the sort and the redeemer-index loop only ever read `network` and `credential`.

ledger/common/rules.go
wolf31o2: Reversed: the earlier reply on this thread was wrong, and so was the commit that acted on it. Tag 7 does require the credential witness.

`getVKeyWitnessConwayTxCert` in `eras/conway/impl/src/Cardano/Ledger/Conway/TxCert.hs`:

```haskell
ConwayRegCert _ SNothing -> Nothing
ConwayRegCert cred (SJust _) -> credKeyHashWitness cred
```

`getScriptWitnessConwayTxCert` in the same file splits identically: `SNothing -> Nothing`, `SJust _ -> credScriptHash cred`. The deposit field is not matched with a wildcard; it is what separates the two forms. `ShelleyRegCert _ -> Nothing` in `eras/shelley/impl/src/Cardano/Ledger/Shelley/TxCert.hs` is the same `SNothing` case, carrying the comment "Registration certificates do not require a witness".

So tag 0 authorizes nothing and tag 7 is authenticated, which is what 3ff6a517 had. 36f10e2 restores `requiresWitness = true` for tag 7 and removes the `optional` set added in 1de43f7.

The two vectors the earlier reply named as red under this rule are green. With tag 7 requiring the witness and no `optional` exemption, `DELEG/Register_stake_credential/When_already_already_registered` and `DELEG/Unregister_stake_credentials/When_registered` both PASS. That claim was wrong, and it was also inconsistent with the 315/315 measured on 3ff6a517, which already had `true`.

Extraneous handling therefore stays strict, matching `validateMissingScripts` in `eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxow.hs` and `babbageMissingScripts` in `eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxow.hs`, which both fail on `sProvided` minus the needed set.

New coverage for the shapes the corpus does not contain: `TestValidateRequiredVKeyWitnessesExplicitRegistration` asserts a tag-7 registration whose credential abstains is rejected (both with no witness set and with an unrelated signature), and `TestValidateScriptWitnessesExplicitRegistration` asserts a tag-0 registration that supplies a script is `ExtraneousScriptWitnessesError`. `TestCertificateAuthorizationCompleteness` moves tag 7 to `requiresWitness: true` and asserts extraneous for every non-authorizing form. All three fail on 905af51 with the fix reverted, and the failing runs compiled.

ledger/common/rules.go
wolf31o2: Verified against the source and now covered by a test.

The order is the derived `Ord` on `AccountAddress` in `libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs`:

```haskell
data AccountAddress = AccountAddress
  { aaNetworkId :: !Network
  , aaId :: !AccountId
  }
  deriving (Show, Eq, Generic, Ord, ...)
```

`AccountId` is a newtype over `Credential Staking`, and `Credential` in `libs/cardano-ledger-core/src/Cardano/Ledger/Credential.hs` derives `Ord` with `ScriptHashObj` declared before `KeyHashObj`. `Network` in `libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs` is `Testnet | Mainnet`, so the numeric network comparison matches. The sort comment now names all three files.

`TestWithdrawalRedeemerIndexFollowsCredentialOrder` builds the case the corpus has none of: one key-credential and one script-credential reward account over the same 28 bytes. It asserts `bytes.Compare(keyBytes, scriptBytes) < 0` first, so the premise that address bytes invert the order is checked rather than assumed, then requires the reward redeemer at index 0. Reverting the credential order makes it fail with `MissingRedeemerForScriptError` expected but nil, and the failing run compiled.

ledger/shelley/rules.go
wolf31o2: The sequencing holds, and it now has a test rather than only a comment.

`TestMIRGenesisQuorumIsNotYetRegistered` walks `UtxoValidationRules` for Shelley, Allegra, Mary, Alonzo, and Babbage and asserts no entry is any era's `UtxoValidateMIRGenesisQuorum`. Each list is checked against all five function pointers, because the four pre-Conway wrappers all delegate to the Shelley rule and any of them would wire it in. Registering the rule in Shelley makes it fail; the entries and the test are meant to change together when blinklabs-io/dingo#3748 lands.

ledger/allegra/rules.go
wolf31o2: No change since the previous reply. The shared `FirstInvalidNativeScriptIn` still carries the key-hash collection, validity-interval defaulting, and evaluation loop for Allegra, Mary, Alonzo, Babbage, and Conway, and Shelley reaches the same evaluator through `FirstInvalidNativeScript`.

ledger/conway/rules.go
wolf31o2: Correcting the previous reply on this thread: the guard it described no longer exists. `common.TransactionMayCarryNativeScript` was removed in 905af51 because its final condition was true for any transaction with inputs, so it never skipped the resolution its comment claimed.

The regression stands as described here. A sound pre-resolution skip is not available: a reference script satisfying a native purpose can come from a consumed input as well as a reference input, and a spending purpose depends on the resolved output's address, so both options need the resolution they would be trying to avoid.

That leaves computing the requirements once per transaction and sharing them between the two rules. `common.UtxoValidationRuleFunc` is `(tx, slot, ls, pp)` with no per-transaction context, so that needs a validation context threaded through the rule lists or a transaction-keyed cache. Filed as blinklabs-io/gouroboros#2167 rather than carried here, since it changes a signature used by every era and is not a correctness fix.

ledger/common/rules.go
wolf31o2: No change since the previous reply. `ValidateScriptWitnesses` still returns early on a nil ledger state, with the comment recording that a reference script cannot be resolved without state and every requirement it would have satisfied would be reported missing.

wolf31o2:
Correction: the current conformance corpus passes 315/315. The earlier 311/315 result came from comparing against a newer ouroboros-mock corpus and was not a failure caused by this PR. MIR quorum validation is now registered in Shelley through Babbage in d20716e.

ledger/shelley/rules.go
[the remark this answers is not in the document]
wolf31o2: Now enforced by `TestMIRGenesisQuorumIsNotYetRegistered`, which asserts no entry in the Shelley, Allegra, Mary, Alonzo, or Babbage `UtxoValidationRules` is any era's `UtxoValidateMIRGenesisQuorum`. Registering it makes the test fail, so the rule-list entries and the test change together when blinklabs-io/dingo#3748 lands.

ledger/common/rules.go
wolf31o2: Confirmed directly against the file, as asked. `getVKeyWitnessConwayTxCert` in `eras/conway/impl/src/Cardano/Ledger/Conway/TxCert.hs` on `master`:

```haskell
      ConwayRegCert _ SNothing -> Nothing
      ConwayRegCert cred (SJust _) -> credKeyHashWitness cred
```

`getScriptWitnessConwayTxCert` splits identically. The deposit field is matched, not wildcarded, so tag 0 authorizes nothing and tag 7 requires the credential. Head keeps `true` for tag 7 and `false` for tag 0.

`TestCertificateAuthorizationRulesAreRegistered` now resolves `UtxoValidateRequiredVKeyWitnesses` and `UtxoValidateScriptWitnesses` out of every era production `UtxoValidationRules` by function pointer and runs both verdicts through the resolved value. Flipping tag 7 fails the Conway and Dijkstra subtests; removing the rule from the Conway list fails on resolution.

ledger/common/rules.go
wolf31o2: Order verified against the derived instances. `AccountAddress { aaNetworkId :: !Network, aaId :: !AccountId }` derives `Ord` in `libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs`, `AccountId` is a newtype over `Credential Staking`, `Credential = ScriptHashObj | KeyHashObj` derives `Ord` in `Credential.hs`, and `Network = Testnet | Mainnet` in `BaseTypes.hs` matches the numeric network id. Head sorts on `(network, credential-type, hash)` with script first.

ledger/shelley/rules.go
wolf31o2: Still deferred at head. `d20716e` registered the rule in the five pre-Conway lists; `4879dd9` reverted that, and `TestMIRGenesisQuorumIsNotYetRegistered` pins the deferral again. Registering it in Shelley fails the shelley subtest.

ledger/common/rules.go
wolf31o2: Partially met at head, deliberately. `common.ValidateMIRGenesisQuorum` and `UtxoValidateMIRGenesisQuorum` exist and `TestValidateMIRGenesisQuorum` covers the four cases you listed plus the unavailable-state and repeated-signature paths.

The rule is not in any era rule list. It fails closed for a ledger state that does not implement `common.GenesisDelegationState`, and no consumer implements it yet, so registering it now would stop a syncing node on the first MIR certificate in Shelley-through-Babbage history. Sequencing is tracked in blinklabs-io/dingo#3748; `TestMIRGenesisQuorumIsNotYetRegistered` pins the deferral so the rule-list entries and the test change together.

So the shape in your regression test is still accepted by `shelley.UtxoValidateRequiredVKeyWitnesses` at head.

ledger/common/rules_test.go
```diff
@@ -512,3 +1451,185 @@ func TestValidateExtraneousRedeemers_Common(t *testing.T) {
 		require.NoError(t, common.ValidateExtraneousRedeemers(tx))
 	})
 }
+
+// UtxoValidateMIRGenesisQuorum is registered in every pre-Conway era. It
+// fails closed when a ledger state cannot provide the genesis-delegation
+// capability, preventing unauthorized MIR certificates from being accepted.
+func TestMIRGenesisQuorumIsRegistered(t *testing.T) {
```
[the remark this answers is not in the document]
wolf31o2: Already reverted. `d20716e` registered the rule and `4879dd9` undid that. At head the rule is in no era rule list and the test is `TestMIRGenesisQuorumIsNotYetRegistered`, asserting the negative.

wolf31o2:
@chrisguiney both blocking items are settled at `45a8ecc`.

**1. Tag 7 — no change, your reading was of a wildcard that is not there.** `getVKeyWitnessConwayTxCert` in `eras/conway/impl/src/Cardano/Ledger/Conway/TxCert.hs` on `master` matches the deposit field:

```haskell
      ConwayRegCert _ SNothing -> Nothing
      ConwayRegCert cred (SJust _) -> credKeyHashWitness cred
```

`getScriptWitnessConwayTxCert` splits identically. Tag 0 authorizes nothing, tag 7 requires the credential. Head keeps `true` for 7 and `false` for 0. `ledger/common/rules.go:253`.

**2. Withdrawal order — fixed.** `(network, credential-type, hash)` with script first, per the derived `Ord` on `AccountAddress` = `(Network, AccountId)`, `AccountId` over `Credential = ScriptHashObj | KeyHashObj`, and `Network = Testnet | Mainnet`. `ledger/common/rules.go:827`.

Per-certificate-type verdict, from `getVKeyWitnessTxCert` and `getScriptWitnessTxCert` (Conway shadows Shelley for tags 0-4 and 7-18; Shelley is authoritative for 5 and 6):

| tag | required witness |
| --- | --- |
| 0 | none |
| 1, 2, 8-13 | the stake credential |
| 3 | pool operator and every owner, vkey only |
| 4 | pool key, vkey only |
| 5 | genesis root key, vkey only |
| 6 | none at field level; the quorum rule stays unregistered |
| 7 | the stake credential |
| 14, 15 | the committee cold credential |
| 16-18 | the DRep credential |

New since your review: `TestCertificateAuthorizationRulesAreRegistered` resolves `UtxoValidateRequiredVKeyWitnesses` and `UtxoValidateScriptWitnesses` out of all seven era production `UtxoValidationRules` by function pointer and runs the tag-0 and tag-7 verdicts through the resolved value, so neither can pass against a function no era registers.

Your other four: `ls == nil` restored, `FirstInvalidNativeScriptIn` shared, MIR still unregistered (`d20716e` registered it, `4879dd9` reverted), input resolution filed as #2167. `gofmt -s -l ledger/` is clean. The description now lists the redeemer-index and native-script changes.

`origin/main` merged, CI green on `45a8ecc`. Re-requesting your review.

ledger/dijkstra/rules.go
```diff
@@ -86,7 +86,6 @@ var UtxoValidationRules = []common.UtxoValidationRuleFunc{
 	UtxoValidateNativeScripts,
 	conway.UtxoValidateDelegation,
 	conway.UtxoValidateWithdrawals,
-	conway.UtxoValidateCertificateDeposits,
 	conway.UtxoValidateCommitteeCertificates,
```
chrisguiney: **Blocking.** This merge drops `conway.UtxoValidateCertificateDeposits` from Dijkstra's rule list, which silently reverts part of #2130 for that era.

It's present in `origin/main` and in the merge base `b46b24a`, and `b432d18` (#2130) *is* an ancestor of head — so this came out of a merge resolution rather than a deliberate edit. `45a8ecc`'s message notes the Conway list keeps every entry from both sides; the Dijkstra list quietly lost one.

Proven with a throwaway probe that resolves the rule by name out of `UtxoValidationRules`:

```go
func TestDijkstraRegistersCertificateDeposits(t *testing.T) {
	dijkstraValidationRule(t, "ledger/conway.UtxoValidateCertificateDeposits")
}
```

```
at 45a8ecc: --- FAIL: validation rule ledger/conway.UtxoValidateCertificateDeposits is not registered
at b46b24a: ok  github.com/blinklabs-io/gouroboros/ledger/dijkstra
```

Nothing in the tree pins this entry, which is why CI stayed green. Dijkstra has no substitute either — `UtxoValidateProposalDeposit` is governance-proposal only, not certificate deposits.

Restoring the line fixes it. It'd also be worth extending `TestDijkstraGovernanceValidationRules`, or adding a sibling, so the Dijkstra list is pinned the way the governance rules already are — this is the second consensus rule to go missing in a merge without a test noticing.

ledger/allegra/rules.go
```diff
@@ -249,7 +249,15 @@ func UtxoValidateNativeScripts(
 	ls common.LedgerState,
 	pp common.ProtocolParameters,
 ) error {
-	if scriptHash, failed := common.FirstInvalidNativeScript(tx, slot); failed {
+	nativeScripts, err := common.NativeScriptsForValidation(tx, ls)
```
chrisguiney, line 320: This rewrite doesn't change behavior in any era it reaches, and it adds a full UTxO resolution pass to three of them.

The description says it delivers reference-script evaluation "from Allegra onward ... the behavioral change is Babbage onward." Babbage doesn't route through here — it has its own `UtxoValidateNativeScripts` at `ledger/babbage/rules.go:1095` built on `script.NativeScriptsToEvaluate`, and Conway delegates to Babbage. So this chain reaches Allegra, Mary and Alonzo only, and `MaryTransactionOutput.ScriptRef()` / `AlonzoTransactionOutput.ScriptRef()` both `return nil`. With no script refs, `available` can only ever hold witness-set scripts, so `NativeScriptsForValidation` returns exactly what `FirstInvalidNativeScript` already returned.

What it does add is a `collectTransactionScriptRequirements` call, which resolves every input and every reference input through `ls.UtxoById` plus four sorts — on a path that previously did zero ledger reads when the witness set had no native scripts.

I checked that it's inert rather than just arguing it: reverting this to `common.FirstInvalidNativeScript(tx, slot)` passes `go test ./ledger/...` and all 431 conformance subtests.

Two ways out, either fine by me: drop the Allegra change and let #2149's Babbage path stand on its own, or keep it and reword the description, since as written it claims a Babbage-onward win this code doesn't deliver.

ledger/dijkstra/rules.go
```diff
@@ -1665,37 +1664,47 @@ func scriptRefIsNativeHash(
...
 
-// UtxoValidateNativeScripts evaluates the native scripts this transaction has
-// to satisfy, with Dijkstra's guard credentials in scope for RequireGuard.
-//
-// The scripts to evaluate come from the resolved transaction view, as in
-// Babbage: a native script some script purpose requires counts whether the
-// witness set, a reference input, or the spent input's own reference-script
-// field supplies it. Dijkstra keeps its own body rather than delegating to
-// Babbage because only it evaluates guards.
 func UtxoValidateNativeScripts(
```
chrisguiney, line 3027: Same shape as the Allegra comment, plus a documentation loss.

Reverting this body to its previous form — `script.NewTxScriptViewSkippingUnresolved` + `NewNativeScriptEnv` + `NativeScriptsToEvaluate` — passes `go test ./ledger/...` and all 431 conformance subtests, so the switch is behaviorally inert here too.

What it changed instead:

- The eight-line doc comment explaining *why* Dijkstra carries its own body (guards) is gone. That was the most useful comment in the function, and it's the question the next reader will have.
- Three lines of `script.NewNativeScriptEnv(tx, slot)` became eighteen hand-rolled lines deriving `keyHashes`, `validityStart` and `validityEnd` — a line-for-line copy of the block inside `FirstInvalidNativeScriptIn`. `NewNativeScriptEnv` still exists and Babbage still uses it.

The wider cost is that the tree now has two implementations of "which native scripts must be evaluated," split by era: `common.NativeScriptsForValidation` for Allegra/Mary/Alonzo/Dijkstra, `script.NativeScriptsToEvaluate` for Babbage/Conway. That's the drift this PR set out to remove, and it's the kind that shows up later as an era-specific consensus divergence.

You made the case earlier that guards can't fold into the shared helper, and I agree — but keeping `NewNativeScriptEnv` for the environment gives you that separation at three lines instead of eighteen, which is roughly where this function already was.

ledger/common/errors.go
```diff
@@ -269,21 +269,22 @@ func (e ExtraneousRedeemerError) Error() string {
...
-// from script presence.
+// MissingRedeemerForScriptError indicates that an available Plutus script is
+// required by a concrete purpose but the witness set has no redeemer at that
+// purpose's exact tag and canonical index.
 type MissingRedeemerForScriptError struct {
-	ScriptHash ScriptHash
-	Tag        RedeemerTag
-	Index      uint32
+	ScriptHash  ScriptHash
+	RedeemerKey RedeemerKey
```
chrisguiney, line 319: `RedeemerKey` now carries the same two values as the `Tag` and `Index` fields below it, and `Error()` reads only `RedeemerKey`.

That makes a construction site setting just `Tag`/`Index` compile cleanly and print `tag=0, index=0`. `ledger/common/script/redeemers.go:107` now has to populate all three to stay correct, which is the tell.

This landed in the same PR that deleted `conway.MissingRedeemerForScriptError` for being a second representation of one error, so I'd rather not reintroduce the split inside the type. Either works:

- keep `Tag`/`Index` (already exported API on `main`) and drop `RedeemerKey`, or
- keep `RedeemerKey` alone and take the API break deliberately.

Whichever you pick, having `Error()` silently ignore two exported fields is the part I'd like gone.

ledger/common/rules.go
```diff
@@ -253,260 +537,480 @@ func ValidateUnsupportedPlutusExecution(tx Transaction, era string) error {
...
+					utxo.Output.ScriptRef(),
+				); err != nil {
+					return ret, err
+				}
+			}
 		}
-		if utxo.Output == nil {
+	}
+
+	inputs := append([]TransactionInput(nil), tx.Inputs()...)
```
chrisguiney, line 848: Two smaller overlaps in this function, both easy:

**Input sort.** `script.SortInputs` (`ledger/common/script/context.go:573`) already sorts by `(Id().Bytes(), Index())` on a copy — same comparator, same semantics. Calling it keeps the canonical spend order defined once.

**Missing-redeemer check.** The spend branch of the redeemer loop at the bottom of `ValidateScriptWitnesses` reproduces `script.ValidateRequiredRedeemers`: same sorted-input indexing, same Plutus test, same `MissingRedeemerForScriptError`. Both rules are registered in Conway (`ledger/conway/rules.go:70` and `:71`), so a missing spend redeemer is now caught twice, by whichever runs first.

The cert/reward/mint/vote/propose coverage the new loop adds is genuinely new and I'm glad it's here — it's specifically the spend half that's already covered. Folding the shared piece into `script.ValidateRequiredRedeemers`, or leaving spend to it, would leave one owner per purpose.

ledger/common/rules.go
```diff
@@ -229,14 +461,66 @@ func ValidateRequiredVKeyWitnesses(tx Transaction) error {
 	for _, vw := range w.Vkey() {
 		vkeyHashes[Blake2b224Hash(vw.Vkey)] = struct{}{}
 	}
-	for _, req := range required {
+	for req := range required {
```
chrisguiney, line 564: Minor: moving `required` from a slice to a map makes the *identity* of the reported signer nondeterministic across runs — `MissingRequiredVKeyWitnessForSignerError` will name a different hash each time for a transaction missing several.

The valid/invalid verdict is unaffected, so this isn't a consensus problem, and the dedup is a real improvement over the old slice. But it can produce confusing bug reports and flaky assertions if a consumer ever matches on the signer. Collecting into a map and then iterating a sorted key slice would keep both properties.

ledger/common/rules.go
```diff
@@ -207,15 +389,65 @@ func ValidateRequiredVKeyWitnesses(tx Transaction) error {
...
+				required[credential.Credential] = struct{}{}
+			}
+		}); err != nil {
+			return err
+		}
+		switch c := cert.(type) {
+		case *PoolRegistrationCertificate:
+			if c == nil {
+				continue
+			}
```
chrisguiney, line 523: Nit: this guard and the two `if c != nil` checks below it are unreachable. `forEachCertificateCredential` runs three lines above and already returns `MalformedAuthorizationError{Subject: "nil pool registration certificate"}` (and the retirement/genesis equivalents) for exactly these typed nils, so the `err != nil` return fires first.

Not harmful, but it reads as if a nil is expected here, which undercuts the fail-closed contract the traversal establishes.

ledger/common/rules.go
```diff
@@ -253,260 +537,480 @@ func ValidateUnsupportedPlutusExecution(tx Transaction, era string) error {
...
+	if ls != nil {
+		for _, input := range tx.Inputs() {
+			utxo, err := ls.UtxoById(input)
+			if err != nil {
+				// BadInputsUtxo reports unresolved consumed inputs before script
+				// evaluation. Preserve that error precedence here.
+				continue
+			}
+			resolvedInputs[input.String()] = utxo
+			if utxo.Output != nil && utxo.Output.ScriptRef() != nil {
```
chrisguiney, line 823: Nit: `utxo.Output.ScriptRef()` is called twice here — once in the condition, once as the argument — and the same pattern repeats at line 737. It's cheap on the Babbage output type, but a local binding reads better and avoids assuming the accessor stays free.

chrisguiney (dismissed):
Re-reviewed at head `45a8ecc` in a clean worktree.

First, the two things I blocked on last time both hold up, and one of them I had backwards:

- **Tag 0 vs tag 7** — I checked `getVKeyWitnessConwayTxCert` and `getScriptWitnessConwayTxCert` on cardano-ledger `master` directly. Both split `ConwayRegCert _ SNothing -> Nothing` / `ConwayRegCert cred (SJust _) -> cred...Witness cred`; the deposit field is matched, not wildcarded. Head's `false` for tag 0 and `true` for tag 7 is correct and my earlier reading was wrong. Thanks for pushing back and re-deriving it.
- **Withdrawal ordering, pool owners, MIR deferral** — all confirmed, and `TestMIRGenesisQuorumIsNotYetRegistered` pins the sequencing nicely.

The certificate and voter authorization work — the part #2127 actually asked for — reads well and I have no correctness objection to it.

## Evidence

| Check | Result |
|---|---|
| `go build ./...`, `go vet ./ledger/...` | pass |
| `go test ./ledger/...` | pass |
| `go test -race ./ledger/...` | pass, no races |
| `go test ./internal/test/conformance/...` | pass, 431 subtests |
| `golangci-lint run ./ledger/...` | 0 issues |
| `gofmt -s -l ledger/` | clean |

No concurrency defects found: every sorted slice is a private copy (`MultiAsset.Policies()` allocates fresh, `inputs` is an explicit copy) and the new code is pure.

Worth flagging: **neither bot has reviewed this head.** CodeRabbit's last pass ended at `57cca07`, ten commits back, and its green check text masks an exhausted hourly quota. Cubic's last run was 15:19 UTC against `d20716e`, which `4879dd9` already reverted — its four P1s are stale, not outstanding.

## Why I'm not approving yet

One blocking item, in the inline comments: a merge resolution dropped `conway.UtxoValidateCertificateDeposits` from Dijkstra's rule list, silently reverting part of #2130. Nothing pins it, so nothing caught it.

The other four are "I'd rather fix these here than carry them" rather than hard blocks — three of them are places where the change costs more than it buys, and given that de-duplication is this PR's own stated goal, they feel in-scope. Happy to be talked out of any of them.

Skipped checks: no Cardano devnet, no Antithesis run, no live-node replay. Coverage is the in-repo conformance corpus and unit tests only.

ledger/dijkstra/rules.go
```diff
@@ -86,7 +86,6 @@ var UtxoValidationRules = []common.UtxoValidationRuleFunc{
 	UtxoValidateNativeScripts,
 	conway.UtxoValidateDelegation,
 	conway.UtxoValidateWithdrawals,
```
arepala-uml: **Blocking.** `conway.UtxoValidateCertificateDeposits` was dropped from Dijkstra's `UtxoValidationRules` with no replacement. Conway still registers it (`ledger/conway/rules.go:89`); Dijkstra's list goes straight from `UtxoValidateWithdrawals` to `UtxoValidateCommitteeCertificates`. Confirmed against the true merge-base (`b46b24a`, not the stale ref): the rule is present in both eras' lists at the base, and this branch's own diff on `ledger/dijkstra/rules.go` removes it from Dijkstra's while adding the new voting-restriction entries around it.

No rule elsewhere in Dijkstra's list checks certificate deposits or refunds, so a Dijkstra transaction with a wrong stake/DRep registration deposit or an incorrect deregistration refund is accepted.

### Regression test

```go
// PR #2129 dropped conway.UtxoValidateCertificateDeposits from Dijkstra's
// UtxoValidationRules while leaving it registered in Conway's. This proves a
// production Dijkstra validation pass accepts a stake-registration
// certificate with an incorrect deposit amount that Conway's identical rule
// list correctly rejects for the same transaction.
func TestDijkstraRuleListAcceptsIncorrectCertificateDeposit(t *testing.T) {
	pp := &conway.ConwayProtocolParameters{
		KeyDeposit:   2_000_000,
		DRepDeposit:  500_000_000,
		MaxTxSize:    16_384,
		MaxValueSize: 5_000,
		ProtocolVersion: common.ProtocolParametersProtocolVersion{
			Major: common.ProtocolVersionVanRossem,
		},
	}
	// ... build one stake-registration certificate (tag 7) claiming a
	// deposit one lovelace short of pp.KeyDeposit, signed by the credential's
	// key, spending one UTxO ...

	// Control: Conway's production rule list rejects it.
	conwayTx := buildTx(t)
	var target conway.CertificateDepositIncorrectError
	conwayErr := runAllRules(conway.UtxoValidationRules, conwayTx, ls, pp)
	require.ErrorAs(t, conwayErr, &target) // passes

	// Same certificate, same parameters, as a real *dijkstra.DijkstraTransaction,
	// run through Dijkstra's production rule list instead.
	dijkstraTx := buildDijkstraTx(t)
	dijkstraErr := runAllRules(dijkstra.UtxoValidationRules, dijkstraTx, ls, pp)
	require.NoError(t, dijkstraErr) // also passes: no rejection
}
```

### Reproduction

```
$ go test ./ledger/common/ -run TestDijkstraRuleListAcceptsIncorrectCertificateDeposit -v -count=1
=== RUN   TestDijkstraRuleListAcceptsIncorrectCertificateDeposit
--- PASS: TestDijkstraRuleListAcceptsIncorrectCertificateDeposit (0.00s)
PASS
```

The test passes at `45a8ecc` — i.e. it *confirms* the gap: Conway rejects the transaction with `CertificateDepositIncorrectError`, and the byte-for-byte equivalent Dijkstra transaction sails through Dijkstra's entire production `UtxoValidationRules` with no error at all.

### Requested change

Register `conway.UtxoValidateCertificateDeposits` in Dijkstra's `UtxoValidationRules`, in the same position relative to `UtxoValidateWithdrawals` that Conway uses, and add a test asserting its presence the way this PR already does for `UtxoValidateMIRGenesisQuorum`'s absence.

ledger/common/rules.go
```diff
@@ -253,260 +537,480 @@ func ValidateUnsupportedPlutusExecution(tx Transaction, era string) error {
...
+				RedeemerTagCert,
+				index,
+			)
+		}); err != nil {
+			return ret, err
 		}
 	}
 
-	// Collect script hashes required by withdrawals
+	// Reward redeemer indices follow cardano-ledger's Withdrawals key order,
```
arepala-uml, line 899: **Blocking.** This fix (script credential before key credential, per `AccountAddress`'s derived `Ord`) is applied only to the witness-completeness check here. It is not mirrored in the two places that build the real `ScriptContext`/`ScriptPurpose` for actual Plutus execution:

- `ledger/common/script/context.go:697` (`withdrawalsInfo`) still sorts withdrawals by raw address bytes (`bytes.Compare(aBytes, bBytes)`) — key header `0xE_` before script header `0xF_`, the exact order this comment says is wrong.
- `ledger/common/script/purpose.go:384` and `:483` (both `scriptPurposeBuilder` and the exported `BuildScriptPurpose`) resolve `RedeemerTagReward` by indexing positionally into that same byte-sorted list.

Neither file is touched anywhere else in this PR (only `ledger/common/script/redeemers.go` gets 4 added lines). So for any transaction with both a key-credential and a script-credential withdrawal, the redeemer index this completeness check now requires for the script withdrawal is, at execution time, resolved to the *other* withdrawal's credential — a completeness-check/execution mismatch on a consensus-critical path, not just a rejected-vs-accepted disagreement.

### Regression test

```go
// PR #2129 fixed the reward-redeemer index used by the completeness check in
// rules.go to follow cardano-ledger's (Network, Credential) order — script
// credential before key credential — but did not update the two withdrawal
// orderings used by actual Plutus script execution in
// ledger/common/script/context.go (withdrawalsInfo) and
// ledger/common/script/purpose.go (scriptPurposeBuilder / BuildScriptPurpose),
// which still sort by raw address bytes (key header 0xE before script header
// 0xF). This proves the redeemer index the completeness check requires for a
// script withdrawal is resolved, at execution time, to the wrong credential.
func TestWithdrawalRedeemerIndexMismatchBetweenValidationAndExecution(t *testing.T) {
	scriptPlutus := common.PlutusV3Script([]byte{0x41, 0x02})
	scriptHash := scriptPlutus.Hash()
	scriptAddr, _ := common.NewAddressFromParts(
		common.AddressTypeNoneScript, common.AddressNetworkTestnet, nil, scriptHash.Bytes())

	keyHash := common.Blake2b224Hash([]byte("unrelated-key-credential"))
	keyAddr, _ := common.NewAddressFromParts(
		common.AddressTypeNoneKey, common.AddressNetworkTestnet, nil, keyHash.Bytes())

	tx := mockledger.NewTransactionBuilder().
		WithWithdrawals(map[*common.Address]uint64{&scriptAddr: 1, &keyAddr: 2}).
		WithWitnesses(mockledger.NewMockTransactionWitnessSet().
			WithPlutusV3Scripts(scriptPlutus).
			WithRedeemers(conway.ConwayRedeemers{Redeemers: map[common.RedeemerKey]common.RedeemerValue{}}))
	ledgerState := mockledger.NewLedgerStateBuilder().Build()

	// Step 1: ask the completeness check which index it requires for the
	// script withdrawal's redeemer.
	var missing common.MissingRedeemerForScriptError
	require.ErrorAs(t, conway.UtxoValidateScriptWitnesses(tx, 0, ledgerState, nil), &missing)
	requiredIndex := missing.RedeemerKey.Index // 0, since script sorts first

	// Step 2: resolve that exact redeemer key with the real execution-path builder.
	purpose, err := script.BuildScriptPurpose(
		common.RedeemerKey{Tag: common.RedeemerTagReward, Index: requiredIndex},
		nil, nil, common.MultiAsset[common.MultiAssetTypeMint]{}, nil,
		tx.Withdrawals(), nil, nil, nil,
	)
	require.NoError(t, err)
	rewarding := purpose.(script.ScriptPurposeRewarding)

	// It resolves to the unrelated KEY credential, not the script credential
	// the completeness check required a redeemer for.
	require.Equal(t, keyHash, rewarding.StakeCredential.Credential)
	require.NotEqual(t, common.Blake2b224(scriptHash), rewarding.StakeCredential.Credential)
}
```

### Reproduction

```
$ go test ./ledger/common/ -run TestWithdrawalRedeemerIndexMismatchBetweenValidationAndExecution -v -count=1
=== RUN   TestWithdrawalRedeemerIndexMismatchBetweenValidationAndExecution
--- PASS: TestWithdrawalRedeemerIndexMismatchBetweenValidationAndExecution (0.00s)
PASS
```

The test passes at `45a8ecc`: index 0, which `UtxoValidateScriptWitnesses` requires a redeemer for to authorize the script withdrawal, is resolved by `script.BuildScriptPurpose` to the unrelated key credential's hash instead.

### Requested change

Apply the same `(Network, Credential)` ordering (script before key) to `withdrawalsInfo` in `ledger/common/script/context.go` and to the withdrawal-sort in both purpose builders in `ledger/common/script/purpose.go`, so the redeemer index the completeness check requires matches the index actual execution resolves. A shared ordering helper used by all three sites would prevent this from splitting again.

arepala-uml (changes requested):
Independent re-verification of the current head (`45a8ecc`) against the actual PR diff (not just the review thread). The extensive back-and-forth with @chrisguiney and cubic is all correctly resolved as of this head — tag-7 witness requirement, pool-owner witnesses, the duplicate error type, and the MIR quorum sequencing all check out on direct re-verification.

Two things weren't caught by that thread because they're outside `ledger/common/rules.go`. Both come with a regression test that fails to demonstrate the fix and passes to demonstrate the bug (i.e. it currently passes, which is the problem).

ledger/dijkstra/rules.go
arepala-uml: Confirmed fixed at `eae6048` (via the `4be2d549` merge of `origin/main`, which carries #2130's `UtxoValidateCertificateDeposits` registration through the new descriptor-based `UtxoValidationRules` in Dijkstra). @chrisguiney independently caught the same regression in a parallel review — good corroboration.

Re-ran the regression test from my last comment against the current head: same incorrect-deposit transaction, same rule lists.

```
$ go test ./ledger/common/ -run TestDijkstraRuleListAcceptsIncorrectCertificateDeposit -v -count=1
=== RUN   TestDijkstraRuleListAcceptsIncorrectCertificateDeposit
    regression_pr2129_test.go:204: dijkstra validation result: incorrect deposit for certificate type 7: supplied 1999999, expected 2000000
--- PASS: TestDijkstraRuleListAcceptsIncorrectCertificateDeposit (0.01s)
PASS
```

Dijkstra now rejects it with `CertificateDepositIncorrectError`, matching Conway. Resolved.

arepala-uml (changes requested):
Re-verified at the latest head (only a routine `merge origin/main` since my last pass — no content changes to the relevant files). Status is unchanged from my last review.

**Still `CHANGES_REQUESTED`.**

**Resolved:** Dijkstra dropping `conway.UtxoValidateCertificateDeposits` from its rule list — fixed via the earlier `4be2d549` merge, confirmed by regression test and independently caught by @chrisguiney in parallel.

**Still blocking:** the withdrawal redeemer-index mismatch between the witness-completeness check (`ledger/common/rules.go`) and actual script execution (`ledger/common/script/context.go`'s `withdrawalsInfo`, `purpose.go`'s `BuildScriptPurpose`) — both still sort withdrawals by raw address bytes while the completeness check now requires the `(network, credential-type, hash)` order. Re-ran the regression test from my earlier comment against `9437648`; it still fails:

```
completeness check required index 0 for script cred d7588e26...; execution resolved
it to aa9b5626... (the unrelated key credential's hash)
--- FAIL: TestWithdrawalRedeemerIndexMismatchBetweenValidationAndExecution
```

For the record: #2187 was filed yesterday for this exact defect (`withdrawalsInfo`/`BuildScriptPurpose` ordering by address bytes instead of `Credential`), independently, with no link back to this PR. Since this PR's own new completeness check is what created the divergence against that pre-existing code, the fix belongs here (or #2187 needs to land and be merged into this branch) before this is safe to merge — filing the issue doesn't change the current code.

No other activity to report — no new bot passes or reviewer comments since yesterday.

arepala-uml (approved):
Re-verified at head `4274a911`. All blocking items are resolved.

**Confirmed fixed:**
- Certificate/voter witness requirements (tag 0 vs 7, pool owners, MIR sequencing) — verified against cardano-ledger source.
- Dijkstra dropping `conway.UtxoValidateCertificateDeposits` — restored via merge.
- Withdrawal redeemer-index mismatch between the completeness check (`ledger/common/rules.go`) and actual script execution (`ledger/common/script/context.go`, `purpose.go`) — `5512c205`/`4274a911` share one `sortWithdrawalAddresses` helper across both files, and its comparator now matches `rules.go`'s. Verified with a regression test: the redeemer index required for a script-credential withdrawal now resolves to that same script credential during execution, not the unrelated key credential it resolved to before.

Issue #2034, opened 2026-08-24
cbor: enforce set tags and strict network decode limits


wolf31o2:
CBOR strictness is defined but not enforced at the protocol boundaries that need it:

- `cbor/tags.go` `SetType.UnmarshalCBOR` accepts an untagged array and `CheckForDuplicates` skips duplicate checks whenever the tag is absent. Conway transaction-input sets carry the same gap.
- `cbor/decode.go` documents `DecodeStrict` as the bounded decoder for untrusted network messages, but no production or test caller uses it; network paths continue through the decoder configured for up to 10,000,000 map or array elements.

Acceptance criteria:

- Make duplicate checks unconditional and enforce the required set tag according to protocol version in every affected set type.
- Route untrusted mini-protocol message decoding through the bounded strict mode, or remove the unused API and enforce one documented bounded mode.
- Add tagged, untagged, duplicate-member, and oversized-collection vectors at protocol boundaries.

PR #2134, opened 2026-08-29
fix(ledger): reject duplicate collection members

### LINKED ISSUE #2133: ledger: reject duplicate transaction map keys and certificates
Duplicate CBOR members in ledger transaction maps can survive decoding when pointer-backed keys hide logical equality. Certificate semantics differ by era: Shelley through Babbage use ordinary lists that permit duplicate entries, while Conway and Dijkstra use ordered sets that reject duplicate logical members.

Acceptance criteria:

- reject duplicate logical keys in withdrawals, governance maps, reward maps, and other affected ledger maps regardless of pointer identity
- preserve duplicate certificate entries in Shelley through Babbage transaction lists
- reject duplicate certificate members in Conway and Dijkstra ordered sets for tagged, untagged, and equivalent encodings
- preserve valid single-entry controls and prove the regression fails against the previous baseline

Related: #2067, #2034

wolf31o2:
Summary:
- reject duplicate logical keys in pointer-backed ledger maps
- keep Shelley through Babbage certificate arrays as ordinary lists, including duplicates
- reject duplicate certificate members in Conway and Dijkstra ordered sets, including equivalent CBOR encodings
- validate duplicate governance, committee, and MIR keys
- give treasury withdrawals the same reward-account validation the transaction body path uses

Certificate identity is built from decoded fields rather than a canonical re-encoding, so two encodings of one certificate still compare equal without paying reflection on the decode path. Types that aggregate over slices or maps, and any certificate type added later, fall back to the re-encoding. An ordinary two-certificate transaction drops from 28909 ns and 94 allocs to 684 ns and 4 allocs; benchmark in `ledger/common/duplicates_bench_test.go`.

`ValidateWithdrawalsMap` is removed. `ValidateWithdrawalAddresses` already rejected nil and duplicate reward accounts in the same decoders, so the eight transaction bodies were scanning withdrawals twice.

`DuplicateCertificateError` carries the offending index and certificate type, and `DuplicateLogicalMapKeyError` carries the repeated key.

Validation:
- full repository and affected example race suites
- build, vet, NilAway, root and example lint, formatting, and tidy checks
- fail-before era matrix against the previous PR head

Closes #2133


arepala-uml (dismissed):

ledger/common/duplicates.go
```diff
@@ -0,0 +1,171 @@
...
+	binary.BigEndian.PutUint32(
+		key[len(actionId.TransactionId):],
+		actionId.GovActionIdx,
+	)
+	return string(key), nil
+}
+
+// ValidateWithdrawalsMap rejects nil and duplicate logical address keys while
+// preserving the existing pointer-keyed public representation.
+func ValidateWithdrawalsMap(withdrawals map[*Address]uint64) error {
```
chrisguiney: This duplicates `ValidateWithdrawalAddresses` (`ledger/common/address.go:789`), which already builds the same `addr.Bytes()` -> string key -> `seen` map and already rejects duplicates.

I ported this PR's own `TestTransactionBodiesRejectDuplicateLogicalWithdrawalKeys` onto `origin/main` (35a63ba) and **it passes without this PR**. The nil case is covered on `main` too — decoding a body with `{5: {null: 1}}` returns `invalid withdrawal address: nil withdrawal address` rather than panicking, so the nil guard here isn't adding protection either.

Cost of the second scan, Conway body decode:

| withdrawals | main | this PR |
|---|---|---|
| 8 | 3829 ns, 97 allocs | 4416 ns, 113 allocs |
| 64 | 28705 ns, 666 allocs | 32116 ns, 797 allocs |

The one call site that genuinely fails on `main` is `TreasuryWithdrawalGovAction`.

Suggestion: drop the `ValidateWithdrawalsMap` calls from the eight transaction-body decoders, use `ValidateWithdrawalAddresses` in `TreasuryWithdrawalGovAction` (see my note there), and remove this function. That also lets `ledger/shelley/withdrawals_decode_test.go` keep its original `duplicate withdrawal reward account` assertion.

Minor, if it does stay: the explicit nil loop here and the nil check inside `validateLogicalMapKeys` scan the map twice for the same condition.

ledger/shelley/shelley.go
```diff
@@ -359,6 +359,9 @@ func (b *ShelleyTransactionBody) UnmarshalCBOR(cborData []byte) error {
 	if _, err := cbor.Decode(cborData, &tmp); err != nil {
 		return err
 	}
+	if err := common.ValidateWithdrawalsMap(tmp.TxWithdrawals); err != nil {
```
chrisguiney: Same as my note on `ledger/common/duplicates.go` — `ValidateWithdrawalAddresses` two lines below already rejects these duplicates, so this pass is doing the work twice. Same applies in allegra, mary, alonzo, babbage, conway, dijkstra, and dijkstra_sub.

ledger/common/gov.go
```diff
@@ -739,6 +753,9 @@ func (a *TreasuryWithdrawalGovAction) UnmarshalCBOR(cborData []byte) error {
 			return errors.New("treasury withdrawal contains a nil address")
 		}
 	}
+	if err := ValidateWithdrawalsMap(tmp.Withdrawals); err != nil {
```
chrisguiney: This is the one call site that genuinely adds coverage — the treasury-withdrawal test does fail on `origin/main`. Nice catch.

Could this use `ValidateWithdrawalAddresses` instead? Treasury withdrawal targets are `reward_account`s in the Conway CDDL, so they should get the same reward-account validation the transaction-body path gets, and that helper already covers the duplicate and nil cases in a single pass. That would also let `ValidateWithdrawalsMap` be removed entirely.

Also minor: the `for address := range tmp.Withdrawals` nil loop just above, `ValidateWithdrawalsMap`'s own nil loop, and `validateLogicalMapKeys`'s nil check make three scans for the same condition.

ledger/common/duplicates.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	if !hasRepeatedType {
+		return nil
+	}
+	seen := make(map[string]struct{}, len(certificates))
+	for _, certificate := range certificates {
+		if typeCounts[certificate.Certificate.Type()] < 2 {
+			continue
+		}
+		encoded, err := cbor.EncodeGeneric(certificate.Certificate)
```
chrisguiney: `EncodeGeneric` is reflection plus `copier` per certificate, and this lands in the transaction decode path. The `typeCounts` gate above only helps when every certificate has a distinct type — real transactions routinely repeat one (two stake delegations, batch registrations, two DRep registrations), so the expensive path is the common one rather than the exception.

Measured on a realistic Conway body (2 inputs, 2 outputs, fee, ttl), `origin/main` vs this branch:

| certificates | main | this PR |
|---|---|---|
| 0 | 2668 ns, 45 allocs | 2700 ns, 45 allocs |
| 1 | 3551 ns, 62 allocs | 3407 ns, 62 allocs |
| **2, same type** | **4336 ns, 77 allocs** | **9022 ns, 171 allocs** |
| 16, same type | 15616 ns, 257 allocs | 51601 ns, 1012 allocs |

The 0- and 1-certificate rows confirm the gate works; the 2-certificate row is an ordinary transaction taking 2.1x longer to decode.

I checked this isn't lock contention — the `EncodeGeneric` type cache is atomic-pointer copy-on-write and the parallel benchmark scales — so it's raw reflection cost.

Would you be open to building the identity from the decoded fields, the way `addressLogicalKey`, `credentialLogicalKey`, and `voterLogicalKey` already do a few lines up? It's more code per certificate type but roughly free. Alternatively, a benchmark in the PR showing the impact is acceptable for dingo's sync path would settle it for me — I just don't want to land an unmeasured 2x on a consensus-critical decoder.

ledger/common/duplicates.go
```diff
@@ -0,0 +1,171 @@
...
+	field string,
+) error {
+	return validateLogicalMapKeys(values, field, credentialLogicalKey)
+}
+
+// ValidateCertificateSet rejects duplicate logical transaction certificates.
+// EncodeGeneric deliberately ignores preserved source CBOR so equivalent
+// definite, indefinite, and non-shortest encodings compare by decoded value.
+func ValidateCertificateSet(certificates []CertificateWrapper) error {
+	var typeCounts [CertificateTypeUpdateDrep + 1]uint
```
chrisguiney: This bound is a forward-compatibility trap. If a future certificate type 19 is added to `certs.go` — plausible, given Dijkstra/Leios is actively evolving and `PoolRegistrationCertificate` already carries `LeiosKey` — then every transaction carrying it is rejected at decode with `invalid certificate type: 19`, from a file nobody would think to update alongside the new certificate.

Could the bound be derived from the constant set, or the counts held in a small `map[uint]int`? The allocation difference is negligible next to the `EncodeGeneric` calls below.

Also worth noting: the `certificateType > uint(CertificateTypeUpdateDrep)` check is unreachable for decoded input, since `CertificateWrapper.UnmarshalCBOR` already rejects unknown types in its `default` branch. Fine as defence for programmatically-built wrappers, just flagging that it isn't load-bearing on the decode path.

ledger/common/duplicates.go
```diff
@@ -0,0 +1,171 @@
...
+	Field string
+}
+
+func (e DuplicateLogicalMapKeyError) Error() string {
+	return "duplicate logical key in " + e.Field
+}
+
+// DuplicateCertificateError indicates that a transaction certificate set
+// contains the same logical certificate more than once.
+type DuplicateCertificateError struct{}
```
chrisguiney: Non-blocking: `DuplicateLogicalMapKeyError` carries `Field`, but this one carries nothing. When a block gets rejected in the wild, an index or certificate type here would make triage a lot easier than `duplicate certificate in transaction certificate set` alone.

ledger/common/duplicates.go
```diff
@@ -0,0 +1,171 @@
...
+
+func addressLogicalKey(address *Address) (string, error) {
+	addressBytes, err := address.Bytes()
+	if err != nil {
+		return "", err
+	}
+	return string(addressBytes), nil
+}
+
+func credentialLogicalKey(credential *Credential) (string, error) {
```
chrisguiney, line 96: Non-blocking, cosmetic: this spends 8 bytes on `CredType`, which `Credential.UnmarshalCBOR` already constrains to `<= CredentialTypeScriptHash` (1). `voterLogicalKey` just below uses a single byte for the same job — worth matching for consistency.

ledger/common/gov.go
```diff
@@ -48,12 +48,26 @@ func (vps *VotingProcedures) UnmarshalCBOR(cborData []byte) error {
...
 			return errors.New("voting procedures contain an empty action map")
 		}
+		if err := validateLogicalMapKeys(
+			procedures,
+			"voting procedure action map",
+			govActionIdLogicalKey,
+		); err != nil {
+			return err
+		}
 		for actionId := range procedures {
```
chrisguiney: This loop is now unreachable. `validateLogicalMapKeys` runs just above and its own nil check fires first (including on the `len(values) < 2` path), returning `voting procedure action map contains a nil key`. So `voting procedures contain a nil action id` can no longer be produced, and the error message for a nil action id has silently changed.

Suggest either deleting this loop, or moving it above the `validateLogicalMapKeys` call to preserve the existing message. The existing test at `ledger/common/gov_test.go:146` only asserts `require.Error`, so it doesn't catch the change either way.

ledger/conway/conway.go
```diff
@@ -634,12 +634,18 @@ func (b *ConwayTransactionBody) UnmarshalCBOR(cborData []byte) error {
...
 		return err
 	}
-	if err := common.ValidateWithdrawalAddresses(tmp.TxWithdrawals); err != nil {
+	if err := common.ValidateWithdrawalsMap(tmp.TxWithdrawals); err != nil {
 		return err
 	}
 	if err := validateConwayCertificateTypes(tmp.TxCertificates); err != nil {
 		return err
 	}
+	if err := common.ValidateCertificateSet(tmp.TxCertificates); err != nil {
```
chrisguiney, line 643: This is the correct behaviour — cardano-ledger's `decodeOSet` enforces uniqueness regardless of whether tag 258 is present — but it does leave three set-shaped fields on the same transaction body following different rules. Measured on this branch:

| field | untagged duplicates | tagged duplicates |
|---|---|---|
| `4` certificates (this PR) | rejected | rejected |
| `0` inputs (`cbor.SetType`) | accepted | rejected |
| `20` proposal_procedures | accepted | accepted |

All three are `set` / `nonempty_set` in the Conway CDDL. `SetType.CheckForDuplicates` skipping untagged arrays is already tracked in #2034, but duplicate `proposal_procedures` being accepted even when tagged looks untracked — worth an issue if you agree.

No change requested here; just flagging so the inconsistency is deliberate rather than incidental.

ledger/conway/conway.go
```diff
@@ -634,12 +634,18 @@ func (b *ConwayTransactionBody) UnmarshalCBOR(cborData []byte) error {
...
+		return err
+	}
+	if err := common.ValidateWithdrawalAddresses(tmp.TxWithdrawals); err != nil {
```
chrisguiney: Small thing: moving `ValidateWithdrawalAddresses` below the certificate checks changes error precedence for a body that has both a non-reward withdrawal address and an invalid certificate — it now reports the certificate error. I couldn't find a reason for the move in the description; was it deliberate?

If the `ValidateWithdrawalsMap` call above is dropped per my other comment, this resolves itself. Same pattern in `ledger/dijkstra/dijkstra.go` for both body types.

chrisguiney (changes requested):
Thanks for this — the certificate work is solid and the era split is right. I confirmed against cardano-ledger that `decodeOSet` uses `decodeSetLikeEnforceNoDuplicates`, so rejecting (rather than deduping) in Conway/Dijkstra matches the node. I also probed past the tests in this PR: duplicates differing by a non-shortest encoding *nested inside a `Credential`*, and by an indefinite-length certificate array, are both caught. The governance, MIR, and voting-procedure holes were real — those tests do fail on `origin/main`.

Two things I'd like to resolve before approving, plus a process note.

**1. The withdrawals half looks redundant with `main`.** I ported `TestTransactionBodiesRejectDuplicateLogicalWithdrawalKeys` onto `origin/main` (35a63ba) and it passes without this PR — `ValidateWithdrawalAddresses` already does the same dedup and already handles a nil key safely. Details inline on `ledger/common/duplicates.go`.

**2. `ValidateCertificateSet` roughly doubles decode cost for ordinary Conway transactions.** Measured numbers inline; the `typeCounts` gate doesn't help when a transaction repeats a certificate type, which is common.

**Bot review status:** neither bot has reviewed the current head (11eede2a). Cubic reported "AI review not needed for merge commit"; CodeRabbit reviewed only bc59c657, the branch's first commit, and reports `0 remain` on its hourly quota. That leaves 7c3005bf and cc315eda unreviewed by bots — which is the code in finding 1. Worth a re-run before human sign-off.

Validation I ran on 11eede2a: `go build`, `go vet`, `gofmt`, `golangci-lint` (0 issues), and `go test -race ./...` — all clean. No data races, goroutine leaks, or lock contention introduced; the `EncodeGeneric` type cache is copy-on-write and the parallel benchmark scales fine.

Everything below finding 2 is non-blocking.

ledger/common/duplicates.go
wolf31o2: Confirmed. `ValidateWithdrawalsMap` is removed, along with `addressLogicalKey`, which had no other caller.

ledger/shelley/shelley.go
wolf31o2: Confirmed. The `ValidateWithdrawalsMap` call is gone from all eight transaction body decoders: shelley, allegra, mary, alonzo, babbage, conway, dijkstra, and dijkstra_sub. `ValidateWithdrawalAddresses` is the single pass in each.

ledger/common/gov.go
wolf31o2: Done. `TreasuryWithdrawalGovAction.UnmarshalCBOR` now calls `ValidateWithdrawalAddresses`, which also removed the separate nil loop above it. Treasury withdrawal targets get the same reward-account validation as the transaction body path, so `TestNewTreasuryWithdrawalGovAction` needed a reward address rather than the enterprise address it used.

`ledger/shelley/withdrawals_decode_test.go` is back to asserting `duplicate withdrawal reward account`, and the treasury and transaction-body cases in `duplicate_collections_integration_test.go` assert the same.

ledger/common/duplicates.go
wolf31o2: Fixed. `certificateLogicalKey` builds the identity from decoded fields, the way `credentialLogicalKey` and `voterLogicalKey` do. Seventeen certificate types are enumerated; the aggregate types keep the canonical re-encoding.

`BenchmarkValidateCertificateSet` is in `ledger/common/duplicates_bench_test.go`. Distinct stake delegations of one type, which is the case the gate cannot skip:

| certificates | before | after |
|---|---|---|
| 2, same type | 28909 ns, 94 allocs | 684 ns, 4 allocs |
| 16, same type | 336897 ns, 755 allocs | 12426 ns, 38 allocs |

The non-shortest-encoding and indefinite-length cases still fail closed, since the key is built after decode.

ledger/common/duplicates.go
wolf31o2: Fixed. `typeCounts` is a `map[uint]int`, so the fixed bound and the `certificateType > uint(CertificateTypeUpdateDrep)` check are both gone.

The forward-compatibility trap would have moved into `certificateLogicalKey` if its `default` branch returned an error, so `default` falls back to the canonical re-encoding instead. A certificate type added to `certs.go` keeps decoding without touching `duplicates.go`, just on the slower path.

ledger/common/duplicates.go
wolf31o2: Done. `DuplicateCertificateError` carries `Index` and `CertificateType`, and the message renders both.

ledger/common/duplicates.go
wolf31o2: Left at eight bytes, and `appendLogicalCredential` uses the same width. `Voter.Type` is a `uint8`, so a single byte cannot alias, but `Credential.CredType` is a `uint` with no type-level bound. `Credential.UnmarshalCBOR` constrains decoded input, but these keys are also built for programmatically constructed credentials, where a single byte would alias `CredType` 1 and 257.

ledger/common/gov.go
wolf31o2: Fixed by moving the nil action id loop above `validateLogicalMapKeys`, so `voting procedures contain a nil action id` is reachable again and the message is unchanged.

ledger/conway/conway.go
wolf31o2: Confirmed. `TxProposalProcedures` is a plain `[]ConwayProposalProcedure` (`ledger/conway/conway.go:626`), not a `cbor.SetType`, so nothing checks it for duplicates whether or not the array is tagged. Same shape on both Dijkstra bodies. Filed as #2161, scoped to `proposal_procedures` since #2034 already covers untagged `SetType`.

The certificate behaviour here is deliberate: `decodeOSet` enforces uniqueness regardless of the tag.

ledger/conway/conway.go
wolf31o2: The move was incidental. With `ValidateWithdrawalsMap` removed, `ValidateWithdrawalAddresses` is back above the certificate checks in Conway and in both Dijkstra bodies, so a body with both a non-reward withdrawal address and an invalid certificate reports the withdrawal error again.

wolf31o2:
Both blocking findings addressed in 5511354; details inline.

- Withdrawals: `ValidateWithdrawalsMap` removed. `ValidateWithdrawalAddresses` was already doing the same work in the same decoders, so the eight bodies scanned withdrawals twice. Treasury withdrawals now use it too, which is the one call site that added coverage. Original error precedence and the original shelley decode assertion are restored.
- Certificates: identity is built from decoded fields rather than `EncodeGeneric`. Ordinary two-certificate case drops from 28909 ns / 94 allocs to 684 ns / 4 allocs; benchmark added in `ledger/common/duplicates_bench_test.go`.
- Type bound is now a map, and unknown certificate types fall back to canonical re-encoding rather than failing, so a future certificate type does not need this file updated.
- `DuplicateCertificateError` carries index and type. The unreachable nil action id loop in voting procedures runs before `validateLogicalMapKeys` again.
- `proposal_procedures` having no duplicate check filed as #2161.

`go build`, `go vet`, `gofmt`, `go test ./...`, and `go test -race ./ledger/...` pass locally.

ledger/common/duplicates.go
```diff
@@ -0,0 +1,273 @@
...
+
+// DuplicateLogicalMapKeyError indicates that two CBOR map keys decoded to the
+// same ledger value. This cannot be left to Go's map duplicate handling when
+// the decoded map is keyed by pointers, because each wire key gets a distinct
+// pointer identity.
+type DuplicateLogicalMapKeyError struct {
+	Field string
+}
+
+func (e DuplicateLogicalMapKeyError) Error() string {
```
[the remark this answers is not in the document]
wolf31o2: Fixed. `DuplicateLogicalMapKeyError` carries `Identity`, the repeated logical key, and renders it as hex after the field name. The key is the same opaque identity `validateLogicalMapKeys` compares, so it is the address bytes for withdrawals and the tagged credential or voter bytes elsewhere. The message falls back to the old form when `Identity` is empty, so a zero-value error still reads correctly.

chrisguiney (approved):

PR #2140, opened 2026-08-30
fix(ledger): gate certificate and governance rules

### LINKED ISSUE #2139: ledger: partition phase-2-invalid validation rules by era
## Problem

Phase-2-invalid transactions must apply only the UTXOW/UTXO checks required before collateral processing. Certificate and governance predicates belong to the phase-2-valid path and must not reject an otherwise valid collateral transition.

## Acceptance criteria

- Partition the shared rule composition explicitly for every supported era.
- Skip certificate and governance rules when `TxIsValid == false`.
- Preserve all well-formedness, witness, script-integrity, collateral, and era-specific invalid-transaction rules that still apply, including Dijkstra-only requirements.
- Keep valid-transaction rule ordering and rule identity stable for consumers and tests.
- Cover invalid transactions containing certificates, unknown voters, unknown governance action IDs, and Dijkstra-specific data.
- Pass complete normal, race, conformance, lint, vet, and NilAway gates.

## Compatibility

Compose newly added certificate and governance rules through the same phase partition so future rules cannot bypass it.

wolf31o2:
Fixes #2139

Skip certificate and governance predicates for phase-2-invalid transactions while preserving required UTXOW, collateral, redeemer, and structural checks.

## Changes

- Group UTXO validation rules into always-run and phase-2-valid-only sets via `ComposeUtxoValidationRules` in `ledger/common`, preserving composed rule order and indexes across Alonzo, Babbage, Conway, and Dijkstra.
- Gate `UtxoValidateRefScriptSizePerTx` in Dijkstra, matching `validateAllRefScriptSize` inside the phase-2-valid branch of upstream `Dijkstra/Rules/Ledger.hs`.
- Pin the composed layout in tests: every ungated rule by name at its index, every gated position by its phase-2 behaviour. Two gated rules swapped with each other are not distinguishable, because the composition wraps each in the same closure.
- Cover `UtxoValidateDelegation` gating for Alonzo and Babbage, which has no internal `IsValid` guard and so is the rule the partition actually changes there.

The inner `IsValid` guards in `UtxoValidateWithdrawals` are kept for callers that invoke the rule directly, and now say so.

## Validation

- `go test ./...` and `go test -race ./ledger/...`
- `go build`, `go vet`, `gofmt`

---

## Summary by cubic

Gates certificate, governance, and Dijkstra reference script size validation rules for phase-2-invalid transactions while preserving UTXOW, collateral, redeemer, and structural checks. Previously these ledger-state rules ran unconditionally and could reject invalid transactions that should only pay collateral.

- Groups UTXO validation rules into always-run and phase-2-valid-only sets via `ComposeUtxoValidationRules` in `ledger/common`, preserving rule order and indexes across Alonzo, Babbage, Conway, and Dijkstra.
- Pins the composed rule layout in tests so reordering or mis-grouping fails, with regression coverage that phase-2-invalid transactions still fail UTXOW, collateral, and redeemer checks.
- Verifies certificate and governance rules are skipped for phase-2-invalid transactions but run for valid ones, including Alonzo/Babbage delegation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction validation across supported protocol versions, including more consistent handling of malformed scripts, withdrawals, credentials, governance data, and certificate deposits.
  * Phase-2 checks now correctly skip transactions marked invalid while required foundational checks continue to run.
  * Improved validation of reference scripts and Plutus witnesses using the active protocol rules.
  * Corrected deposit refunds, historical deposit handling, governance guardrails, committee certificates, and voter validation.
* **Tests**
  * Added comprehensive coverage for validation ordering, phase-2 behavior, script checks, governance rules, and deposit accounting.


arepala-uml (dismissed):

ledger/dijkstra/rules_test.go
```diff
@@ -40,16 +40,26 @@ func dijkstraValidationRule(
 	want string,
 ) (common.UtxoValidationRuleFunc, int) {
 	t.Helper()
-	for idx, rule := range UtxoValidationRules {
-		if strings.HasSuffix(dijkstraValidationRuleName(rule), want) {
-			return rule, idx
+	groups := [][]common.UtxoValidationRuleFunc{
```
chrisguiney: This helper now walks `dijkstraGovernanceRulesBeforeUtxow` and `dijkstraLedgerRulesAfterUtxow` — the same two literals the test is meant to be checking — so `TestDijkstraGovernanceValidationRules` can only re-confirm their declaration order. It no longer looks at the composed list at all.

I checked this rather than assuming: with the two `Phase2ValidUtxoValidationRules(...)` groups swapped inside `ComposeUtxoValidationRules`, so the composed list genuinely is mis-ordered, `go test -run 'TestDijkstraGovernanceValidationRules$'` still passes. On `main` the helper scanned `UtxoValidationRules`, and the same mutation would have failed there. A mis-ordered composition is the specific new failure mode this PR introduces, so it would be good to keep it covered.

The only thing that caught my mutation was the incidental `UtxoValidationRules[1]` assertion further down, which is not really what that test is for.

Would it work to assert against the composed list directly — check the always-run rules by name at their expected indices, and assert the gated positions no-op for a `TxIsValid: false` transaction and delegate for a valid one? Happy to look at alternatives if you would rather keep the group slices as the source of truth.

ledger/dijkstra/rules.go
```diff
@@ -91,10 +59,60 @@ var UtxoValidationRules = []common.UtxoValidationRuleFunc{
...
+		UtxoValidateNativeScripts,
+	),
+	common.Phase2ValidUtxoValidationRules(
+		dijkstraLedgerRulesAfterUtxow...,
+	),
+	common.AlwaysUtxoValidationRules(
+		UtxoValidateMalformedReferenceScripts,
+	),
+	common.Phase2ValidUtxoValidationRules(
+		UtxoValidateRefScriptSizePerTx,
```
chrisguiney: This gating is correct — I confirmed `validateAllRefScriptSize` sits inside the `isPhase2ValidTxL == Phase2Valid` branch in upstream `Dijkstra/Rules/Ledger.hs`. My concern is only that it is not pinned anywhere.

It landed in its own commit (c5d9b34) after CodeRabbit's only pass, and it is the most spec-sensitive call in the PR: the one place where a size-limit check is deliberately skipped for an invalid transaction. Nothing would fail today if it were moved back into the always group, or if a later rule addition reordered it.

Could we add a small case — a transaction whose reference scripts exceed `MaxRefScriptSizePerTx`, asserting `RefScriptSizePerTxTooLargeError` with `TxIsValid: true` and `NoError` with `TxIsValid: false`?

ledger/alonzo/rules.go
```diff
@@ -25,38 +25,42 @@ import (
...
+		UtxoValidateWrongNetworkWithdrawal,
+		UtxoValidateMaxTxSizeUtxo,
+		UtxoValidateExUnitsTooBigUtxo,
+		UtxoValidateNativeScripts,
+		UtxoValidateExtraneousRedeemers,
+		UtxoValidatePlutusScripts,
+	),
+	common.Phase2ValidUtxoValidationRules(
+		UtxoValidateDelegation,
+		UtxoValidateWithdrawals,
```
chrisguiney: `UtxoValidateWithdrawals` already opens with `if !tx.IsValid() { return nil }` (`ledger/alonzo/rules.go:801`, and likewise `babbage:1096` and `conway:3260`), so for this one rule the gating now lives in two places, while its neighbour `UtxoValidateDelegation` relies solely on the wrapper.

Harmless today, but #2139's compatibility note asks that the partition be the single mechanism so future rules cannot bypass it, and two sources of truth leave it ambiguous which one a new rule should follow. Could we drop the inner guards now that the composition covers them — or, if they are deliberately kept for callers that invoke the rule directly, add a short comment saying so?

(Related, and not something I am asking you to change: `UtxoValidatePlutusScripts` sits in `AlwaysUtxoValidationRules` but returns early on `!tx.IsValid()` internally at `conway:2665` and `dijkstra:387`. The behavior is right; it just makes the "Always" grouping read as slightly inaccurate.)

ledger/dijkstra/rules_test.go
```diff
@@ -40,16 +40,26 @@ func dijkstraValidationRule(
...
+			}
+			idx++
 		}
 	}
 	t.Fatalf("validation rule %s is not registered", want)
 	return nil, -1
 }
 
 func TestDijkstraGovernanceValidationRules(t *testing.T) {
+	require.Len(t, UtxoValidationRules, 56)
```
chrisguiney: Nit: this is now the only remaining guard on the composed list, and a bare `56` is hard to act on when it eventually fails — the next person cannot tell whether a rule was dropped or the count legitimately moved. It also only catches count changes, not reordering or a rule landing in the wrong phase group.

Could it move to a dedicated test with a named constant, or be replaced by the positional assertion suggested on the helper above? Minor either way.

ledger/dijkstra/rules_test.go
```diff
@@ -91,11 +101,20 @@ func TestDijkstraGovernanceValidationRejectsTypedNilParameterChange(
...
-		}},
-	}}
+	tx := &DijkstraTransaction{
+		Body: DijkstraTransactionBody{
+			TxProposalProcedures: []DijkstraProposalProcedure{{
+				PPGovAction: DijkstraGovAction{Action: action},
+			}},
+		},
+	}
+	require.NoError(t, UtxoValidationRules[1](
```
chrisguiney: Nit: this pins `UtxoValidateProposalProcedures` to absolute index 1. CodeRabbit flagged the same pattern in `rules_phase2_invalid_test.go` and you fixed it there in c5d9b34 by resolving rules by name, so it would be nice to be consistent here.

It is also asserting phase-2 gating inside a test named for typed-nil handling. As things stand this line is what actually catches a mis-ordered composition, which is easy to lose in a future edit — moving it into an explicit gating test would make that intent visible.

ledger/common/rules.go
```diff
@@ -38,6 +38,68 @@ type UtxoValidationRuleFunc func(
...
+			continue
+		}
+		for _, rule := range group.rules {
+			ret = append(ret, func(
+				tx Transaction,
+				slot uint64,
+				ledgerState LedgerState,
+				protocolParams ProtocolParameters,
+			) error {
+				if tx != nil && !tx.IsValid() {
```
chrisguiney: Nit: a nil `Transaction` falls through this guard and runs the wrapped rule, which will panic in most of them, so the check does not buy much. It also will not help against a typed nil, since `ConwayTransaction.IsValid` has a value receiver and `(*ConwayTransaction)(nil)` panics on the call.

`TestComposeUtxoValidationRules` does not exercise either path. Either dropping the `tx != nil` half or adding a case that documents the intended nil behavior would make the intent clearer.

ledger/babbage/rules.go
```diff
@@ -28,43 +28,49 @@ import (
...
+		UtxoValidateMaxTxSizeUtxo,
+		UtxoValidateExUnitsTooBigUtxo,
+		UtxoValidateTooManyCollateralInputs,
+		UtxoValidateNativeScripts,
+		UtxoValidateExtraneousRedeemers,
+		UtxoValidatePlutusScripts,
+	),
+	common.Phase2ValidUtxoValidationRules(
+		UtxoValidateDelegation,
+		UtxoValidateWithdrawals,
```
chrisguiney: Nit, on coverage: the era-level phase-2-invalid regression test only covers Conway, so Alonzo and Babbage gating rests entirely on `TestComposeUtxoValidationRules` in `ledger/common`.

An Alonzo or Babbage case for `UtxoValidateDelegation` would be cheap and would genuinely exercise the wrapper — unlike `UtxoValidateWithdrawals`, `UtxoValidateDelegation` has no internal `IsValid` guard, so it is the rule whose behavior actually changes here.

chrisguiney (changes requested):
Thanks for this — the core partition is right, and I checked it against upstream rather than just reading it. Three requested changes below, all in the tests; I did not find a defect in the production behavior.

## Verified

- **Matches upstream.** `dijkstraLedgerTransition` in `Dijkstra/Rules/Ledger.hs` places `validateTreasuryValue`, `validateAllRefScriptSize`, `ENTITIES` and `GOV` inside the `isPhase2ValidTxL == Phase2Valid` branch and calls `UTXOW` unconditionally afterwards. The Always/Phase2Valid split here lines up with that, including the `UtxoValidateRefScriptSizePerTx` move in c5d9b34 — I went looking for a bypassed size/DoS check there, and upstream gates it too.
- **Order and index positions are preserved.** I dumped `runtime.FuncForPC` names for all four eras on `origin/main` and on this head. Every difference is a name change at an identical index; counts unchanged (alonzo 30, babbage 35, conway 55, dijkstra 56).
- **Downstream dingo is unaffected.** Its pinned skip indices (alonzo 27, babbage 31, conway 24/43, dijkstra 38) all land on Always-group rules, so the init-time name check in `validateUtxoValidationSkipIndex` still resolves. dingo also already gates state application on `tx.IsValid()` (`ledger/delta.go:200,288`), so no path applies certs/gov for an invalid transaction.
- **No concurrency concerns.** Composition runs at package init, closures capture per-iteration `rule` under go 1.25, no shared mutable state.

## Checks run locally

`go build ./...`, `go test ./ledger/...`, `go test -race ./ledger/...`, `go vet ./ledger/...`, and `golangci-lint run ./ledger/...` (0 issues) all pass. Not run locally: nilaway (CI green), conformance and devnet.

## Requested changes

Inline at `ledger/dijkstra/rules_test.go:43`, `ledger/dijkstra/rules.go:112`, and `ledger/alonzo/rules.go:61`. The rest are marked as nits.

One thing worth a moment's thought but not a change request: `TxIsValid` zero-values to `false`, so a transaction built in Go without setting it now skips every gated rule rather than just Withdrawals and PlutusScripts. Decoded transactions are unaffected since the flag comes from CBOR, and the one test this touched is already patched — just something for anyone writing fixtures later.

## Process note

CodeRabbit's check is green, but its comment says auto incremental reviews are disabled — its last actual pass was `abf2017`, before both `c5d9b34` and the `62055bc` merge, so the reference-script-size change has not been bot-reviewed. Probably worth a `@coderabbitai review`. Cubic did cover the current head.

ledger/dijkstra/rules_test.go
wolf31o2: Confirmed, and the merge with main surfaced it as a hard failure: main added `TestDijkstraWellFormednessPrecedesPlutusExecution`, which resolves `UtxoValidateMalformedReferenceScripts` through this helper, and the narrowed helper cannot see it.

`dijkstraComposedLayout` now declares the expected composed list, 56 entries of name and gating, and `dijkstraValidationRule` resolves through it into `UtxoValidationRules`. `TestDijkstraComposedRuleLayout` asserts every ungated rule by name at its index and every gated position by no-op for a `TxIsValid: false` transaction with nil parameters, which would panic if the wrapper delegated.

The layout is declared in the test rather than derived from the group literals, so it is a second statement of the ordering rather than a re-read of the first. Verified against the mutation: swapping the two `Phase2ValidUtxoValidationRules` groups fails `TestDijkstraComposedRuleLayout`.

Several subtests needed `TxIsValid: true` once the helper began returning the composed rule, which is the zero-value point noted in the review.

ledger/dijkstra/rules.go
wolf31o2: Added `TestDijkstraRefScriptSizePerTxIsPhase2Gated`: reference scripts of 60 and 60 against `MaxRefScriptSizePerTx: 100`, asserting `RefScriptSizePerTxTooLargeError` with `TxIsValid: true` and `NoError` with `TxIsValid: false`, resolved through the composed list rather than called directly.

Moving the rule into an Always group fails both that test and `TestDijkstraComposedRuleLayout`.

ledger/alonzo/rules.go
wolf31o2: Kept the inner guards with a comment rather than removing them, taking the second option.

Removing all three leaves exactly one failure, `TestUtxoValidateWithdrawals_DRepDelegationProtocolGate/phase-2_invalid_skips_DRep_delegation_check`, which asserts that behaviour for a direct call. Since the rule is invoked directly in many tests and this is a consensus path where the guard can only skip work, removing it risks a valid block being rejected by a direct caller for no gain. The comment now names the composition as the mechanism a new rule should use and points at that test.

`UtxoValidatePlutusScripts` is left alone as noted.

ledger/dijkstra/rules_test.go
wolf31o2: Replaced. `require.Len(t, UtxoValidationRules, 56)` is gone; the length is now asserted against `len(dijkstraComposedLayout)`, and the layout also covers reordering and wrong-phase placement, which the count could not.

ledger/dijkstra/rules_test.go
wolf31o2: Fixed. That call site resolves `ledger/dijkstra.UtxoValidateProposalProcedures` through `dijkstraValidationRule` instead of indexing.

The mis-ordered-composition coverage that line was incidentally providing now lives in `TestDijkstraComposedRuleLayout`.

ledger/common/rules.go
wolf31o2: Dropped the `tx != nil` half. It did not help a typed nil, and a nil transaction reaches the same panic through `tx.IsValid()` that it previously reached inside the wrapped rule, so the partition now has one reading.

ledger/babbage/rules.go
wolf31o2: Added `ledger/babbage/rules_phase2_invalid_test.go` covering `UtxoValidateDelegation` for both Alonzo and Babbage: a stake delegation to an unregistered pool errors with `TxIsValid: true` and returns nil with `TxIsValid: false`. The rule is located by matching the delegation error across the composed list rather than by index.

Moving `UtxoValidateDelegation` into the Alonzo Always group fails it.

wolf31o2:
Merged current main and addressed all three requested changes plus the nits; details inline.

Merge resolution: main moved `UtxoValidateMalformedReferenceScripts` ahead of `UtxoValidatePlutusScripts` in Babbage, Conway, and Dijkstra. That position is adopted inside the composed groups, so the composed list matches main rule for rule. Verified by dumping `runtime.FuncForPC` names for all four eras on this head and on main: counts identical (alonzo 30, babbage 35, conway 55, dijkstra 56) and every difference is a wrapper name at an identical index, with no ungated rule moved.

One note for the earlier downstream check: main's move shifts `UtxoValidatePlutusScripts` to babbage 32, conway 44, dijkstra 44. Dingo pins 31, 43, and 38. That shift is main's, not this branch's, and dingo's init-time name check will catch it on its next bump.

Test changes: the composed layout is now declared in the test and asserted against `UtxoValidationRules`, so swapping the two Phase2Valid groups fails. Ref-script-size gating is pinned directly, Alonzo and Babbage delegation gating is covered, and rules resolve by name rather than absolute index.

`go build`, `go vet`, `gofmt`, `go test ./...`, and `go test -race ./ledger/...` pass locally.

ledger/alonzo/rules.go
```diff
@@ -794,6 +798,10 @@ func UtxoValidateWithdrawals(
 	ls common.LedgerState,
 	pp common.ProtocolParameters,
 ) error {
+	// The composed rule list gates this rule on phase-2 validity. This inner
+	// guard is kept for callers that invoke the rule directly, which
+	// TestUtxoValidateWithdrawals_DRepDelegationProtocolGate covers. A new
+	// rule should rely on the Phase2ValidUtxoValidationRules group instead.
```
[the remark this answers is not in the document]
wolf31o2: Correct, the citation was mine and it overstated the coverage. The Alonzo and Babbage comments no longer name a test; they state only why the inner guard is kept and that a new rule should use the Phase2Valid group.

ledger/babbage/rules.go
```diff
@@ -1086,6 +1090,10 @@ func UtxoValidateWithdrawals(
 	ls common.LedgerState,
 	pp common.ProtocolParameters,
 ) error {
+	// The composed rule list gates this rule on phase-2 validity. This inner
+	// guard is kept for callers that invoke the rule directly, which
+	// TestUtxoValidateWithdrawals_DRepDelegationProtocolGate covers. A new
+	// rule should rely on the Phase2ValidUtxoValidationRules group instead.
```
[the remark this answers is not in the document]
wolf31o2: Same, fixed alongside the Alonzo comment.

ledger/dijkstra/rules_test.go
```diff
@@ -35,20 +35,161 @@ func dijkstraValidationRuleName(rule common.UtxoValidationRuleFunc) string {
...
 
+// TestDijkstraComposedRuleLayout pins the composed list: every ungated rule by
+// name at its index, and every gated position by its phase-2 behavior. A rule
+// moved between groups, dropped, or reordered fails here.
+func TestDijkstraComposedRuleLayout(t *testing.T) {
+	require.Len(t, UtxoValidationRules, len(dijkstraComposedLayout))
+	invalidTx := &DijkstraTransaction{TxIsValid: false}
+	for idx, entry := range dijkstraComposedLayout {
+		rule := UtxoValidationRules[idx]
+		if entry.gated {
```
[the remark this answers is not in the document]
wolf31o2: Correct, and the doc comment now says so rather than implying otherwise.

The composition wraps every gated rule in an identical closure, so a gated position can only be identified by the behavior all gated rules share. What the layout does catch: a rule dropped, a rule moved between the Always and Phase2Valid groups in either direction, and any ungated rule reordered or renamed. What it cannot catch is two gated rules swapped with each other. Both the test and `dijkstraValidationRule` now record that limit, and closing it would need the wrapper to carry the wrapped rule identity.

ledger/dijkstra/rules_test.go
```diff
@@ -194,11 +335,24 @@ func TestDijkstraGovernanceValidationRejectsTypedNilParameterChange(
...
+			TxProposalProcedures: []DijkstraProposalProcedure{{
+				PPGovAction: DijkstraGovAction{Action: action},
+			}},
+		},
+	}
+	proposalRule, _ := dijkstraValidationRule(
+		t,
+		"ledger/dijkstra.UtxoValidateProposalProcedures",
+	)
+	require.NoError(t, proposalRule(
```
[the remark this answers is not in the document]
wolf31o2: Correct, removed. The rule is gated and `tx.TxIsValid` is still false at that point, so the wrapper short-circuited and the assertion could not fail. The `common.VerifyTransaction` call below it, after `TxIsValid` is set, is the one that exercises the rule.

Worth noting it was equally dead before this branch: the previous `UtxoValidationRules[1](...)` call resolved to the same gated wrapper.

ledger/dijkstra/rules_test.go
wolf31o2: Re-ran the group-swap mutation against the current head before touching anything: it passed `TestDijkstraComposedRuleLayout`. The merge with main added `UtxoValidateCertificateDeposits` and `UtxoValidateCommitteeCertificates` to the post-UTXOW group, so both gated groups hold ten rules and the phase-2 no-op pattern is identical either way. The swap still failed four governance tests in the package, but only incidentally, which is the same weakness this thread named.

b4fb956 pins it where the intent is stated. `dijkstraGatedRuleAnchors` identifies index 1 by `ProtocolParameterUpdateEmptyError` from an empty parameter-change proposal and index 46 by the unregistered-pool delegation error, and asserts both positions still return nil for `TxIsValid: false`. That also covers the second half of the suggestion here: a gated position must delegate for a valid transaction.

Proved red, both compiling: swapping the two `Phase2ValidUtxoValidationRules` groups fails `TestDijkstraComposedRuleLayout` at both anchors, and a wrapper that short-circuits for valid transactions too fails the same two.

Composed order re-verified after the merge. Unwrapping the gate and dumping `runtime.FuncForPC` names for all four eras on this head is identical to `origin/main` line for line: alonzo 30, babbage 35, conway 56, dijkstra 57. dingo's pinned indices resolve to the same functions on both sides (alonzo 27 `UtxoValidatePlutusScripts`, conway 24 `UtxoValidateFeeTooSmallUtxo`, babbage 31 and conway 43 `UtxoValidateMalformedReferenceScripts`, dijkstra 38 `UtxoValidateMaxTxSizeUtxo`), and all five positions are in Always groups. The last three no longer match the `UtxoValidatePlutusScripts` name dingo expects; that is main's reordering, present with or without this branch.

ledger/dijkstra/rules_test.go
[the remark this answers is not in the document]
wolf31o2: Two gated rules swapped inside one group are still not distinguishable, and the doc comment says so. The case of the two gated groups exchanged with each other is now closed: b4fb956 anchors one rule in each group by an error only that rule returns, so the swap fails `TestDijkstraComposedRuleLayout`.

wolf31o2:
b4fb956 addresses the coverage gap the merge introduced.

- The merge with main made both Dijkstra gated groups ten rules long, which left `TestDijkstraComposedRuleLayout` unable to distinguish them; swapping the two `Phase2ValidUtxoValidationRules` groups passed it. Anchors now pin index 1 by `ProtocolParameterUpdateEmptyError` and index 46 by the unregistered-pool delegation error, and assert both positions no-op for `TxIsValid: false`. The swap and a never-delegating wrapper both fail it.
- Three Conway assertions that re-ran the call which selected the rule are removed; the helper's "no composed rule reports this error" check is the assertion, and it fails when `UtxoValidateIsValidFlag` is moved into the gated group.

No production code changed in this commit. Composed order re-checked against `origin/main` with the gate unwrapped: identical for all four eras (alonzo 30, babbage 35, conway 56, dijkstra 57).

Local: `go build ./...`, `go vet ./...`, `gofmt -l .`, `go test -race ./...`, `golangci-lint run ./...` (0 issues), `nilaway -include-pkgs=github.com/blinklabs-io/gouroboros ./...` (clean). Bare `nilaway ./...` reports three findings in `ledger/alonzo/genesis.go`, `protocol/blockfetch/client.go`, and `consensus/genesis/genesis_test.go`; all three reproduce on `origin/main` and are outside the CI scope. Conformance and devnet not run.

wolf31o2:
@coderabbitai review

ledger/conway/rules_test.go
```diff
@@ -3563,20 +3563,43 @@ func TestProductionValidationSkipsCommitteeRulesForPhase2Invalid(
...
+					// provider, and no committee error is produced.
+					err := common.VerifyTransaction(
+						tx,
+						0,
+						state,
+						nil,
 						era.rules,
-						conway.UtxoValidateIsValidFlag,
-						operation.rule,
 					)
```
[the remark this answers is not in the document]
wolf31o2: Valid case added, but not through `VerifyTransaction`: as noted, it stops at `UtxoValidateCostModelsPresent`.

f98d45b resolves the composed committee rule out of `UtxoValidationRules` by the lookup error only it reports for a `TxIsValid: true` transaction, asserts the provider is queried at that position, then asserts nil and zero queries at the same position for the invalid transaction. The previous valid case passed `[]{conway.UtxoValidateCommitteeCertificates}`, which tested the function rather than its place in the composed list and passed with the gate reverted.

ledger/dijkstra/rules_test.go
wolf31o2: One layout entry changed in f98d45b: `conway.UtxoValidateGovActionWellFormedness` at index 2 is now `gated: false`, and `TestDijkstraComposedRuleLayout` asserts it by name there. Putting it back in a `Phase2ValidUtxoValidationRules` group fails that assertion.

wolf31o2:
f98d45b, one defect and two test gaps.

**Defect.** `UtxoValidateGovActionWellFormedness` was gated whole, but it bundles two kinds of check. The nil action and the 28-byte policy and constitution script hash lengths stand in for upstream CBOR decoding, which rejects the transaction before the LEDGER rule and never consults phase-2 validity; `PolicyHash` and `Constitution.ScriptHash` are `[]byte` here, so the rule is the only place the length is enforced. Gating it let a phase-2-invalid transaction with a 20-byte policy hash through, where cardano-node fails to decode the block. Those checks are now `govActionStructuralWellFormedness`, the rule sits in the always-run group at its existing index in Conway and Dijkstra, and only its `ConflictingCommitteeUpdate` and guardrails half is gated.

**Per-rule verdict, checked against the era files.** Every other gated rule is on the certificate/governance application path, not UTXOW. Conway `conwayLedgerTransitionTRC` runs `validateTreasuryValue`, `validateRefScriptSize`, `validateWithdrawalsDelegated`, `testIncompleteAndMissingWithdrawals`, CERTS and GOV inside `isPhase2ValidTxL == Phase2Valid` and calls UTXOW unconditionally; Alonzo's `ledgerTransition` (Babbage reuses it) gates `testIncompleteAndMissingWithdrawals` and DELEGS the same way; Dijkstra gates `validateAllRefScriptSize`, ENTITIES and GOV. The ten pre-UTXOW rules map to GOV predicate failures (`MalformedProposal`, `ProposalCantFollow`, `InvalidPrevGovActionId`, `ProposalDepositIncorrect`, `ProposalProcedureNetworkIdMismatch`, `TreasuryWithdrawalsNetworkIdMismatch`, `ProposalReturnAccountDoesNotExist`, `ZeroTreasuryWithdrawals`, `DisallowedProposalDuringBootstrap`), and the post-UTXOW ten to GOV (`VotersDoNotExist`, `GovActionsDoNotExist`, `VotingOnExpiredGovAction`, `DisallowedVotesDuringBootstrap`, `DisallowedVoters`) or CERTS via DELEG/POOL/GOVCERT. `UtxoValidateWrongNetworkWithdrawal` correctly stays unconditional for Conway (`ConwayUtxoPredFailure`).

**Test gaps.** The Conway governance group before UTXOW had no test reaching it through the composition — `UtxoValidateUnknownVoters`, `UtxoValidateCommitteeCertificates` and `UtxoValidateCertificateDeposits` each carry their own phase-2 guard, so the existing subtests pass with the composition gate removed. Added `TestPhase2InvalidSkipsGovernanceProposalRules`. The phase-2-valid committee subtest called the rule in a single-element slice; it now resolves the composed rule out of `UtxoValidationRules` by behavior.

Composed order re-verified: with the gate neutralised, `runtime.FuncForPC` names are identical to `origin/main` line for line in all four eras (alonzo 30, babbage 36, conway 57, dijkstra 58), so dingo's pinned indices are unaffected.

Two pre-existing Dijkstra divergences found while checking, neither introduced here, both getting their own issue: `WrongNetworkWithdrawal` moved from UTXO to ENTITIES in Dijkstra (`Dijkstra/Rules/Utxo.hs` marks the Conway constructor `Impossible`, `Entities.hs:284` runs it inside the Phase2Valid branch), and Dijkstra's unconditional `validateBatchWithdrawals` UTXO check (`Utxo.hs:265`, `WithdrawalsExceedAccountBalance`) has no counterpart here.

ledger/alonzo/rules.go
wolf31o2: Audited the class rather than the three rules named. Three more gated rules carry their own `!tx.IsValid()` guard, all from main rather than this branch: `conway.UtxoValidateUnknownVoters` (rules.go:3846), `conway.UtxoValidateCommitteeCertificates` (:3662) and `conway.UtxoValidateCertificateDeposits` (:3381). Consequence for coverage: the composition gate is unobservable for those three, so the era-level subtests that match them pass with the gate reverted. f98d45b adds `TestPhase2InvalidSkipsGovernanceProposalRules` against a rule that has no inner guard, which is what actually pins the gate for the pre-UTXOW group.

Left the guards and their comments as they are rather than widening the diff; happy to document those three the same way if you would rather have it in this PR.

wolf31o2:
@coderabbitai review

ledger/dijkstra/rules.go
wolf31o2: Fixture updated in 5decc3a: main reworked the per-tx limit to count consumed reference scripts, so the oversized script now arrives through a reference input rather than an output.

One thing that changed with that: main also added an internal `!tx.IsValid()` guard to `UtxoValidateRefScriptSizePerTx` in both Conway and Dijkstra. The composition gate is therefore unobservable for this rule, and this test passes with the gate reverted. `TestDijkstraComposedRuleLayout` still pins the placement (`gated: true` at index 57). Main also added `conway.UtxoValidateRefScriptSizePerTx`; it goes in the trailing Phase2Valid group, matching `validateRefScriptSize` inside the Phase2Valid branch of `conwayLedgerTransitionTRC`.

wolf31o2:
5decc3a merges main, which landed `conway.UtxoValidateRefScriptSizePerTx` and reworked the Dijkstra per-tx limit.

- The new Conway rule goes in the trailing Phase2Valid group, matching `validateRefScriptSize` inside the Phase2Valid branch of `conwayLedgerTransitionTRC`.
- Main also added an internal `!tx.IsValid()` guard to `UtxoValidateRefScriptSizePerTx` in both eras, so the composition gate is unobservable for that rule and `TestDijkstraRefScriptSizePerTxIsPhase2Gated` now pins the behavior rather than the placement. `TestDijkstraComposedRuleLayout` still pins the placement.
- Its fixture reaches the oversized script through a reference input, since the limit now counts consumed rather than published reference scripts.

Composed order re-verified with the gate neutralised: identical to `origin/main` line for line in all four eras (alonzo 30, babbage 36, conway 58, dijkstra 58).

ledger/babbage/rules.go
```diff
@@ -286,12 +290,15 @@ func UtxoValidateInlineDatumsWithPlutusV1(
 	ls common.LedgerState,
 	pp common.ProtocolParameters,
 ) error {
-	// An unresolvable input yields an empty view rather than an error:
-	// UtxoValidateBadInputsUtxo reports it with the right error, and reporting
-	// it here as well would make this rule a second, competing source of
-	// input-resolution failures.
-	view, err := script.NewTxScriptViewSkippingUnresolved(tx, ls)
+	view, err := script.NewTxScriptView(tx, ls)
```
chrisguiney: This reverts main. `origin/main` has `script.NewTxScriptViewSkippingUnresolved(tx, ls)` here (added in 5375087); this head reinstates `NewTxScriptView` plus an inline copy of the helper's `errors.Is` handling.

The behaviour is the same — an empty view makes `NeedsAny` false, so main returns nil too — but it re-duplicates the logic main deliberately extracted, and it is not a change this PR needs.

I traced it rather than guessed: a9e0117 resolved the merge by taking the branch side for `babbage/rules.go`, `conway/rules.go` and `dijkstra/rules.go`. It also dropped `babbage.UtxoValidateRequiredRedeemers` outright, which is why d918133 exists — that one surfaced as a test failure, so it got caught. The rest was comments and this call, so nothing failed.

Could you redo the merge taking main's side for the functions this PR does not touch? A comment-stripped diff confirms this call is the only non-comment body difference outside the intended scope, so it should be a small fix.

ledger/babbage/rules.go
```diff
@@ -1066,12 +1073,8 @@ func UtxoValidateScriptWitnesses(
 }
 
 // UtxoValidateRequiredRedeemers checks that every Plutus script-address
```
chrisguiney, line 1215: Same merge, documentation side: about 30 lines of rationale that exist on `origin/main` are gone at this head. Five sites —

- `babbage/rules.go:1075` and `conway/rules.go:1445` — the CIP-33 note on `UtxoValidateRequiredRedeemers` explaining the gap `script.ValidateRequiredRedeemers` closes, and why Babbage must reject a reference-script-backed input with no redeemer even though it never executes Plutus.
- `babbage/rules.go:1087` — why the scripts to evaluate come from the resolved view rather than the witness set, and that those are the same three sources `UtxoValidateScriptWitnesses` accepts.
- `conway/rules.go:3056` — the "Conway inherits Babbage's rule unchanged" note.
- `dijkstra/rules.go:1690` — why Dijkstra keeps its own body rather than delegating to Babbage (guard credentials for `RequireGuard`).

None of it is behavioural, but it is someone else's design rationale and this PR has no reason to touch it. Restoring main's text for these five should fall out of the same merge fix.

ledger/conway/rules_test.go
```diff
@@ -3725,39 +3725,129 @@ func TestProductionValidationSkipsCommitteeRulesForPhase2Invalid(
...
 					}
-					rules := registeredRules(
-						t,
+					// The composed rule list wraps every phase-2-gated
+					// rule in an identical closure, so a gated rule cannot be
+					// selected out by function pointer. Run the whole list and
+					// assert the property this test is named for: a
+					// phase-2-invalid transaction never reaches the committee
+					// provider, and no committee error is produced.
+					err := common.VerifyTransaction(
```
chrisguiney, line 3746: This makes the subtest vacuous, and I checked before saying so rather than reading it off the shape.

`VerifyTransaction` returns on the first error, and for these fixtures that is `UtxoValidateRedeemerAndScriptWitnesses` — "missing Plutus script witnesses for redeemers", which I confirmed by logging `err`. The committee rules sit around index 50 and are never reached, so `errors.As(...) == false` and `state.queries == 0` hold no matter what the gating does. Adding `WithUtxoById` got the run further but not far enough.

The decisive check: with the composition gate disabled **and** the inner `!tx.IsValid()` guards stripped from both `UtxoValidateCommitteeCertificates` and `UtxoValidateUnknownVoters`, all six `committee_*` subtests still pass in Conway and Dijkstra. Only the new `phase-2 valid transaction runs committee rules` catches it. On main the subtest selected the two rules and did reach them, so this is a step back from where it was.

Would it work to iterate the composed list calling each rule directly and assert `state.queries == 0` across all of them, the way `conwayComposedRulesForError` does in `rules_phase2_invalid_test.go`? That sidesteps the short-circuit and keeps the pointer-selection problem solved.

chrisguiney (dismissed):
Thanks — everything I asked for on 08-31 is addressed, and I re-verified rather than took it on trust. Two new problems came in with the 09-02 merges, both inline. The partition itself is still right.

## Verified on this head

- **Order and identity are byte-identical to main.** I neutralised the wrapper and dumped `runtime.FuncForPC` names for all four eras on both sides: 182 rules (alonzo 30, babbage 36, conway 58, dijkstra 58), `diff` empty. 44 positions are wrapped.
- **dingo is unaffected.** All six pinned indices (`ledger/eras/validation.go:112-117`) resolve to the same function on main and on this head, and none sits at a wrapped position, so `validateUtxoValidationSkipIndex` still resolves.
- **The partition matches upstream, checked as a class.** The gated set is exactly CERTS + GOV + `validateRefScriptSize`. `validateProtocolParameterUpdate` (`conway/rules.go:1204`) is `ppuWellFormed`, i.e. GOV, so gating it is right; only the nil-action and 28-byte hash checks stand in for decoding, which is what f98d45b moves to always-run. That split looks correct to me.
- **Mutation coverage.** Gate disabled → all four era packages fail. Dijkstra gated groups swapped → `TestDijkstraComposedRuleLayout` fails. `UtxoValidateRefScriptSizePerTx` moved to Always → layout fails. Conway `UtxoValidateGovActionWellFormedness` gated → `TestGovActionRepresentabilityIsNotPhase2Gated` fails.
- **No concurrency concerns.** Composition at package init, per-iteration capture under go 1.25, no shared state.

## Checks run locally

`go build ./...`, `go vet ./ledger/...`, `gofmt`, `golangci-lint run ./ledger/...` (0 issues) and `go test -race ./ledger/...` all pass. `go test ./...` had one failure, `pipeline.TestValidateStage_Eta0ProviderError` — a timing flake that passes `-count=3` in isolation, in a package this PR does not touch. Not run: nilaway (CI green), conformance, devnet.

## Process note

Neither bot has reviewed d918133..b791bb3. CodeRabbit's last real pass was 09-02 02:27Z — the 15:41Z request was rate-limited, and its green check reads "incremental reviews are disabled". Cubic's last was 09-01 18:26Z. So neither has seen the merge behind the first finding below, nor the gov-action split. Worth another pass once the window resets.

One observation, not a change request: splitting `UtxoValidateGovActionWellFormedness` into two passes changes which error a multi-proposal transaction reports when an earlier proposal has a conflicting committee update and a later one a malformed hash. Both still reject, so this only affects the message.

ledger/dijkstra/rules_test.go
chrisguiney: Addressed — thanks for taking the harder route with `dijkstraComposedLayout` rather than patching the helper.

Re-ran the mutation myself on this head: swapping the two `Phase2ValidUtxoValidationRules` groups fails `TestDijkstraComposedRuleLayout` at the anchor assertion (rules_test.go:181). The anchors are what makes it stick now that both groups are the same length, so that part was the right call.

ledger/dijkstra/rules.go
chrisguiney: Addressed. Moving `UtxoValidateRefScriptSizePerTx` into an Always group fails `TestDijkstraComposedRuleLayout` on this head — I ran it.

Agreed on your reading of the fixture: main added an inner `!tx.IsValid()` guard to the rule, so `TestDijkstraRefScriptSizePerTxIsPhase2Gated` no longer distinguishes the composition gate. The layout entry is doing the pinning, which is enough for what I was asking — this rule now has the same belt-and-braces shape as `UtxoValidateWithdrawals`.

ledger/dijkstra/rules_test.go
chrisguiney: Addressed — `require.Len(t, UtxoValidationRules, len(dijkstraComposedLayout))`, and the layout covers reordering and wrong-phase placement that the bare count never could.

ledger/dijkstra/rules_test.go
chrisguiney: Addressed. The typed-nil test now runs the whole composed list and the incidental `UtxoValidationRules[1]` assertion is gone, with the mis-ordering coverage moved into `TestDijkstraComposedRuleLayout` where it is visible.

ledger/common/rules.go
chrisguiney: Addressed — the wrapper reads on one mechanism now. Thanks.

ledger/babbage/rules.go
chrisguiney: Addressed. `TestAlonzoBabbagePhase2InvalidSkipsDelegation` fails in both eras when I disable the gate in `ComposeUtxoValidationRules`, so it is exercising the wrapper rather than an inner guard. Locating the rule by matching the delegation error instead of by index is nicer than what I suggested.

ledger/alonzo/rules.go
chrisguiney: Addressed — the comment is at `alonzo/rules.go:801`, `babbage/rules.go:1124` and `conway/rules.go:3350`, and naming the composition as the mechanism a new rule should use is exactly what I was after. Keeping the guards is the right call for a consensus path; thanks for proving it down to the one test that depends on it.

Your audit matches mine. I counted seven gated rules carrying an inner guard on this head: `UtxoValidateWithdrawals` in all three eras, plus `conway.UtxoValidateCertificateDeposits` (rules.go:3423), `conway.UtxoValidateCommitteeCertificates` (:3707), `conway.UtxoValidateUnknownVoters` (:3891) and `UtxoValidateRefScriptSizePerTx` in both Conway (ref_scripts.go:47) and Dijkstra (rules.go:1771). Three of the seven say why.

Consider that a nit and your call — I am not asking for it in this PR. If you would rather leave the diff where it is, a line in the PR description naming the four undocumented ones would do the same job for the next reader.

arepala-uml (approved):

PR #2151, opened 2026-08-30
fix(pipeline): make drain boundary exact

### LINKED ISSUE #2150: fix(pipeline): make WaitForDrain honor in-flight processing
`BlockPipeline.WaitForDrain` decides completion from `PendingCount`, which samples channel depths and apply-stage pending work. Decode or validation workers can temporarily hold an accepted item between channels while every sampled count is zero, allowing `WaitForDrain` to return before ordered processing finishes.

Acceptance criteria:

- Use the pipeline's exact accepted-submission boundary when waiting for drain.
- Preserve context cancellation and stopped/not-started behavior.
- Add deterministic regression coverage that blocks in-flight stage work and proves the wait cannot return early.
- Run the pipeline suite normally and with the race detector.
- Clarify that `PendingCount` is observational and must not be used as a completion barrier.

Related: blinklabs-io/dingo#3532

wolf31o2:
## Summary

- wait on the exact accepted-submission boundary before reporting pipeline drain completion
- preserve caller cancellation, parent cancellation, and explicit shutdown semantics
- document approximate pending counts and add deterministic lifecycle regressions

## Validation

- `make test`
- `make lint`
- focused pipeline and ChainSync tests with the race detector
- `go build ./...`

Fixes #2150

---
## Summary by cubic
Fixes #2150 by making pipeline drain completion wait on the exact accepted-submission boundary. `WaitForDrain` previously polled `PendingCount`, so it could return while an accepted block was still in flight; it now uses the same fence boundary as `Fence`. Submission and fencing now also reject work as soon as shutdown begins, so a stopped pipeline never accepts new work even when the submit gate is ready.

**Lifecycle semantics**
- An explicit `Stop` makes drain succeed even with work in flight; parent context cancellation makes drain return `ErrPipelineStopped`.
- Caller cancellation takes precedence in `Submit`, `Fence`, and `WaitForDrain` and returns `context.Canceled`.
- `Submit` and `Fence` now reject pre-canceled callers and canceled pipelines without entering the submit gate, and refuse work once `Stop` has begun.
- `PendingCount` is now documented as an observational snapshot, not a completion barrier; use `Fence` or `WaitForDrain` instead.

<sup>Written for commit 756bd98f13d9424f9a795e9cec20f21c202e8419. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2151?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pipeline shutdown and cancellation handling to prevent work from being accepted after stopping begins.
  * Improved drain waiting behavior for more reliable completion detection.
  * Prevented sequence gaps during concurrent submission, cancellation, and shutdown scenarios.

* **Tests**
  * Added comprehensive coverage for submission, fencing, draining, cancellation, and concurrent shutdown behavior.

### DIFF
--- a/pipeline/pipeline.go
+++ b/pipeline/pipeline.go
@@ -121,14 +121,30 @@ func NewBlockPipeline(opts ...PipelineOption) *BlockPipeline {
 }
 
 func (p *BlockPipeline) lockSubmit(ctx context.Context) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	if p.ctx.Err() != nil {
+		return ErrPipelineStopped
+	}
+
 	select {
 	case <-ctx.Done():
 		return ctx.Err()
 	case <-p.ctx.Done():
 		return ErrPipelineStopped
 	case <-p.submitGate:
-		return nil
 	}
+
+	if err := ctx.Err(); err != nil {
+		p.unlockSubmit()
+		return err
+	}
+	if p.ctx.Err() != nil {
+		p.unlockSubmit()
+		return ErrPipelineStopped
+	}
+	return nil
 }
 
 func (p *BlockPipeline) unlockSubmit() {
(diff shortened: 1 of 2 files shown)


wolf31o2:
Current head `883397e3` has no unresolved review threads and independent exact-head review is clear. The Go, lint, and NilAway failures reproduce from the undefined Byron signing tag on `main`; #2153 contains the minimal repair and is fully green. Refresh these checks after #2153 lands.

arepala-uml (approved):

PR #2157, opened 2026-08-31
fix(blockfetch): preserve incomplete range error

### LINKED ISSUE #2156: fix(blockfetch): preserve incomplete-range error during shutdown
## Summary

`TestGetBlockRejectsEmptyBatch` intermittently receives `protocol is shutting down` instead of the incomplete-range protocol error after an empty batch closes the connection.

## Scope

- Preserve the incomplete-range failure for the affected request while shutdown propagates.
- Keep connection termination behavior intact.
- Add deterministic regression coverage for both error delivery orders.

## Acceptance criteria

- The requester consistently receives the incomplete-range error.
- The connection error path remains observable and shutdown completes without leaks.
- Focused normal and race repetitions pass on supported Go versions.

wolf31o2:
Closes #2156.

- Preserve incomplete-range errors when shutdown interrupts queued or active BlockFetch requests.
- Keep request completion and in-flight byte accounting balanced across shutdown orderings.
- Cover pending and active request shutdown paths.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved block fetching during protocol shutdowns.
  * Preserves blocks and batch-start results that completed before shutdown was detected.
  * Returns more accurate results when no block is available, including the appropriate “no blocks” condition.
  * Improved handling of concurrent connection errors and request outcomes for more reliable error reporting.


wolf31o2:
Replaced the two helper unit tests with call-site tests. `TestRequestResultOrShutdown` and `TestCompletedBlockOrShutdown` asserted against the new helpers directly, so both passed with the fix reverted and covered nothing.

`GetBlock`'s block wait is now `awaitBlock` and `waitForBatchDone`, making both shutdown branches reachable from tests. `TestWaitForBatchStartPreservesResultAtShutdown` and `TestAwaitBlockPreservesResultAtShutdown` drive an already-closed protocol done channel across repeated iterations of the both-ready select; 6 of 8 rows fail with the fix reverted, the other 2 are shutdown boundary rows.

Merged `main` at ec8104f.

arepala-uml (approved):

PR #2159, opened 2026-09-01
fix(script): validity-interval upper-bound closure is era-dependent (Conway exclusive, pre-Conway closed)

chadaibbhot-git:
## Problem
`TimeRange.ToPlutusData` rendered the validity-interval upper bound with a single closure for all Plutus versions/eras. cardano-ledger's `transValidityInterval` makes the upper-bound closure **era-dependent**: pre-Conway (Alonzo/Babbage) Plutus V1/V2 render a TTL-only (`invalidHereafter`, upper-only) interval as a **closed** upper bound `UpperBound (Finite t) True`, while Conway renders it **exclusive** `UpperBound (Finite t) False` (cardano-ledger#3043; the Conway-only `transVITimeUpperBoundIsOpen` test). A single closure diverges from the node on one side, causing ExUnits / script-evaluation divergence.

## Fix
Gate the closure on the target era. Thread a `strictValidityUpperBound bool` through `NewTxInfoV1FromTransaction` / `NewTxInfoV2FromTransaction` (Plutus V3 is always strict). `TimeRange` carries a `strictUpperBound` field; `ToPlutusData` computes the finite upper-bound closure as `!lowerBoundPresent && !strictUpperBound` — closed only for a pre-Conway upper-only interval, exclusive when strict (Conway/Dijkstra) or when both bounds are present. Callers in `ledger/conway/rules.go` and `ledger/dijkstra/rules.go` pass `true`.

## Tests
- New `context_timerange_test.go`: pre-Conway (closed) vs Conway (exclusive) rendering for upper-only, both-bounds, and lower-only intervals.
- `context_validity_test.go`: V1/V2 Alonzo/Babbage reverted to the inclusive expectation.
- `context_test.go`: V3 Withdraw golden keeps the exclusive bound.

Refs cardano-ledger#3043.


chadaibbhot-git:
@cubic-dev-ai re: the era-dependent upper-bound closure at `ledger/common/script/context.go` (that inline thread is now outdated after the rework push) —

Agreed — the upper-bound closure is era-dependent, good catch. I've reworked the PR to gate it rather than blanket-exclusive: Plutus V1/V2 under pre-Conway (Alonzo/Babbage) keep the **closed** upper bound for a TTL-only interval (`!lowerBoundPresent`), and only Conway/Dijkstra use the **exclusive** bound — matching cardano-ledger#3043 and the Conway-only `transVITimeUpperBoundIsOpen` test. Implementation: a `strictValidityUpperBound` flag threaded through `NewTxInfoV1/V2FromTransaction` (V3 always strict); `TimeRange.ToPlutusData` now computes closure as `!lowerBoundPresent && !strictUpperBound`. Added `context_timerange_test.go` covering both eras. Thanks for the precise pointer.

chadaibbhot-git:
Superseded by #2170 — this PR's branch lost common history with `main` (it had been based on a vendored snapshot rather than a fork of main), which auto-closed it and blocked reopening. #2170 is the same change rebuilt cleanly on current main.

PR #2162, opened 2026-09-01
fix(ledger): validate reference-provided native scripts

### LINKED ISSUE #2149: ledger: validate reference-provided native scripts
## Scope

Validate needed native scripts regardless of whether they are supplied in the
witness set, a reference input, or a spent input's reference-script field.

## Acceptance criteria

- Unsatisfied needed native scripts are rejected for every supported era with
  reference scripts.
- Satisfied reference-provided native scripts and unrelated unused reference
  scripts remain valid.
- Needed-script and provided-script discovery uses the same transaction view
  as the script-witness rules.
- Regression tests cover Babbage, Conway, and Dijkstra through production rule
  sets, including both reference-input and spent-input delivery.
- The tests fail against the pre-fix implementation and pass with the change.
- Normal, race, lint, vet, NilAway, conformance, and integration checks pass.

chrisguiney:
Closes #2149.

## Problem

`UtxoValidateNativeScripts` read only `witnesses.NativeScripts()`. A native script that a script purpose required but that arrived as a *reference* script — on a reference input, or on the spent input's own reference-script field, both permitted by CIP-33 — was never evaluated, so an unsatisfied one passed phase-1 validation.

## Change

Babbage, Conway and Dijkstra take the scripts to evaluate from the resolved `script.TxScriptView`:

- every native script in the witness set, as before;
- plus every **needed** native script that only a reference script supplies.

Those are the same three sources `UtxoValidateScriptWitnesses` accepts a required script from, so a script that rule counts as provided is one this rule evaluates. A reference script no purpose requires stays unevaluated, matching `validateFailedBabbageScripts` in `Cardano.Ledger.Babbage.Rules.Utxow`, so spending a UTxO that happens to carry an unrelated reference script remains valid.

Witness-set order followed by hash order keeps the reported script hash deterministic when more than one script fails.

Alonzo and earlier are untouched: they have no reference scripts.

Two cleanups fall out: Conway delegates to Babbage instead of keeping a copy of the loop, and the skip-unresolved-input idiom `babbage.UtxoValidateInlineDatumsWithPlutusV1` had inline moves into `script.NewTxScriptViewSkippingUnresolved`.

## Tests

`ledger/common/reference_native_script_test.go` drives Babbage, Conway and Dijkstra through their registered `UtxoValidationRules` (`UtxoValidateScriptWitnesses` + `UtxoValidateNativeScripts`, selected out of the production slice) for five cases each: reference-input and spent-input delivery of an unsatisfied script, both of those satisfied, and an unrelated unused reference script.

Against the pre-fix implementation the two unsatisfied cases fail in all three eras (6 subtests); the other three pass before and after:

```
--- FAIL: TestReferenceProvidedNativeScriptsAreValidated/Babbage/reference_input_provides_unsatisfied_script
--- FAIL: TestReferenceProvidedNativeScriptsAreValidated/Babbage/spent_input_provides_unsatisfied_script
--- FAIL: TestReferenceProvidedNativeScriptsAreValidated/Conway/reference_input_provides_unsatisfied_script
--- FAIL: TestReferenceProvidedNativeScriptsAreValidated/Conway/spent_input_provides_unsatisfied_script
--- FAIL: TestReferenceProvidedNativeScriptsAreValidated/Dijkstra/reference_input_provides_unsatisfied_script
--- FAIL: TestReferenceProvidedNativeScriptsAreValidated/Dijkstra/spent_input_provides_unsatisfied_script
```

`ledger/common/script/native_test.go` covers the shared helper directly: order stability across 20 runs, exclusion of available-but-unneeded and non-native needed scripts, deduplication of a script provided twice, the zero view, and `NewNativeScriptEnv`.

## Checks

`go build ./...`, `gofmt -l`, `go vet ./...`, `golangci-lint run ./...` (0 issues), `go test -race ./...`, `go test ./internal/test/conformance/...`, and NilAway (`-include-pkgs=github.com/blinklabs-io/gouroboros`) all pass.

`golines --dry-run` reports one long line in `ledger/babbage/rules.go` around the ExUnits accumulation; it is present on `main` before this change and is left alone.

https://claude.ai/code/session_01KaMGAPzVsZe1gpxMvfXJf1

---
## Summary by cubic
Fixes phase-1 native script validation so an unsatisfied native script supplied as a reference script — on a reference input or the spent input's own reference-script field, both allowed by CIP-33 — no longer passes. Babbage, Conway, and Dijkstra now evaluate every needed native script from the resolved transaction view instead of the witness set alone.

- Closes #2149.
- A reference script no script purpose requires stays unevaluated, so spending a UTxO carrying an unrelated reference script remains valid.
- The reported script hash stays deterministic when several scripts fail: witness-set order, then hash order.
- An unresolvable input yields an empty view; the earlier bad-inputs and script-witness rules reject such transactions, so this rule never competes with them.
- Conway delegates to Babbage, and the skip-unresolved-input idiom from the inline-datums rule moves into `script.NewTxScriptViewSkippingUnresolved`.
- Adds per-era tests for five delivery cases, unresolvable-input rejection, and unit tests for the new helper.
- Alonzo and earlier are untouched; they have no reference scripts.

<sup>Written for commit 585f342bd7104ac225c2aab014aad5b165d66da2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2162?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Native scripts can now be validated when supplied through reference scripts on reference inputs or spent outputs.
  - Validation is supported across Babbage, Conway, and Dijkstra transaction rules.

- **Bug Fixes**
  - Evaluates only native scripts required by the transaction.
  - Reports the specific failing script when validation fails.
  - Rejects unresolved consumed or reference inputs before native-script validation.

### DIFF
--- a/ledger/babbage/rules.go
+++ b/ledger/babbage/rules.go
@@ -285,15 +285,12 @@ func UtxoValidateInlineDatumsWithPlutusV1(
 	ls common.LedgerState,
 	pp common.ProtocolParameters,
 ) error {
-	view, err := script.NewTxScriptView(tx, ls)
+	// An unresolvable input yields an empty view rather than an error:
+	// UtxoValidateBadInputsUtxo reports it with the right error, and reporting
+	// it here as well would make this rule a second, competing source of
+	// input-resolution failures.
+	view, err := script.NewTxScriptViewSkippingUnresolved(tx, ls)
 	if err != nil {
-		if errors.Is(err, common.ErrInputResolution) ||
-			errors.Is(err, common.ErrReferenceInputResolution) {
-			// UtxoValidateBadInputsUtxo reports an unresolvable input with the
-			// right error; reporting it here as well would make this rule a
-			// second, competing source of input-resolution failures.
-			return nil
-		}
 		return err
 	}
 	needsV1 := view.NeedsAny(func(s common.Script) bool {
@@ -1067,14 +1064,41 @@ func UtxoValidateScriptWitnesses(
 	return common.ValidateScriptWitnesses(tx, ls)
 }
 
-// UtxoValidateNativeScripts evaluates native scripts in the transaction.
+// UtxoValidateNativeScripts evaluates the native scripts this transaction has
+// to satisfy.
+//
+// Babbage is the first era in which a script can reach a transaction as a
+// reference script rather than a witness, so unlike the Alonzo rule this
+// replaces, the scripts to evaluate come from the resolved transaction view
+// rather than the witness set alone: a native script some script purpose
+// requires counts whether the witness set, a reference input, or the spent
+// input's own reference-script field supplies it -- the same three sources
+// UtxoValidateScriptWitnesses accepts a required script from, so a script that
+// rule counts as provided is a script this rule evaluates rather than ignores.
 func UtxoValidateNativeScripts(
 	tx common.Transaction,
 	slot uint64,
 	ls common.LedgerState,
 	pp common.ProtocolParameters,
 ) error {
-	return alonzo.UtxoValidateNativeScripts(tx, slot, ls, pp)
+	view, err := script.NewTxScriptViewSkippingUnresolved(tx, ls)
+	if err != nil {
+		return err
+	}
+	env := script.NewNativeScriptEnv(tx, slot)
+	for _, nativeScript := range script.NativeScriptsToEvaluate(tx, view) {
+		if !nativeScript.Evaluate(
+			env.Slot,
+			env.ValidityStart,
+			env.ValidityEnd,
+			env.KeyHashes,
+		) {
+			return allegra.NativeScriptFailedError{
+				ScriptHash: nativeScript.Hash(),
+			}
+		}
+	}
+	return nil
 }
 
 // UtxoValidateWithdrawals validates withdrawals against ledger state.
--- a/ledger/common/reference_native_script_test.go
+++ b/ledger/common/reference_native_script_test.go
@@ -1,0 +1,300 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common_test
+
+import (
+	"bytes"
+	"errors"
+	"reflect"
+	"runtime"
+	"strings"
+	"testing"
+
+	"github.com/blinklabs-io/gouroboros/cbor"
+	"github.com/blinklabs-io/gouroboros/ledger/allegra"
+	"github.com/blinklabs-io/gouroboros/ledger/babbage"
+	"github.com/blinklabs-io/gouroboros/ledger/common"
+	"github.com/blinklabs-io/gouroboros/ledger/conway"
+	"github.com/blinklabs-io/gouroboros/ledger/dijkstra"
+	"github.com/blinklabs-io/gouroboros/ledger/shelley"
+	mockledger "github.com/blinklabs-io/ouroboros-mock/ledger"
+	"github.com/stretchr/testify/require"
+)
+
+// testPubkeyNativeScript builds a native script that is satisfied only when
+// vkey has signed the transaction, so a case toggles satisfied/unsatisfied by
+// adding or omitting the vkey witness rather than by moving the slot.
+func testPubkeyNativeScript(
+	t *testing.T,
+	vkey []byte,
+) common.NativeScript {
+	t.Helper()
+	scriptCbor, err := cbor.Encode(common.NativeScriptPubkey{
+		Type: 0,
+		Hash: common.Blake2b224Hash(vkey).Bytes(),
+	})
+	require.NoError(t, err)
+	var nativeScript common.NativeScript
+	require.NoError(t, nativeScript.UnmarshalCBOR(scriptCbor))
+	return nativeScript
+}
+
+func testScriptPaymentAddress(
+	t *testing.T,
+	scriptHash common.ScriptHash,
+) common.Address {
+	t.Helper()
+	addr, err := common.NewAddressFromParts(
+		common.AddressTypeScriptNone,
+		common.AddressNetworkMainnet,
+		scriptHash.Bytes(),
+		nil,
+	)
+	require.NoError(t, err)
+	return addr
+}
+
+func testKeyPaymentAddress(t *testing.T, vkey []byte) common.Address {
+	t.Helper()
+	addr, err := common.NewAddressFromParts(
+		common.AddressTypeKeyNone,
+		common.AddressNetworkMainnet,
+		common.Blake2b224Hash(vkey).Bytes(),
+		nil,
+	)
+	require.NoError(t, err)
+	return addr
+}
+
+// testOutput builds a Babbage output, optionally carrying scriptRef as its
+// reference script. Babbage is the first era with reference scripts and the
+// output type every later era wraps, so the same output serves all three
+// eras under test.
+func testOutput(
+	addr common.Address,
+	scriptRef common.Script,
+) common.TransactionOutput {
+	out := &babbage.BabbageTransactionOutput{OutputAddress: addr}
+	if scriptRef != nil {
+		out.TxOutScriptRef = &common.ScriptRef{
+			Type:   common.ScriptRefTypeNativeScript,
+			Script: scriptRef,
+		}
+	}
+	return out
+}
+
+// testLedgerState resolves exactly the inputs the case set up. An input the
+// case did not register fails to resolve, so a rule that reaches for one the
+// test did not intend is caught rather than handed a zero value.
+func testLedgerState(
+	outputs map[string]common.TransactionOutput,
+) common.LedgerState {
+	return mockledger.NewLedgerStateBuilder().
+		WithUtxoById(
+			func(input common.TransactionInput) (common.Utxo, error) {
+				output, ok := outputs[input.String()]
+				if !ok {
+					return common.Utxo{}, errors.New("unknown input")
+				}
+				return common.Utxo{Id: input, Output: output}, nil
+			},
+		).
+		Build()
+}
+
+// scriptAuthorizationRules picks the two rules that decide whether a needed
+// script is provided and whether it is satisfied out of an era's production
+// rule set, so a case runs the registered rules rather than a function the
+// era might not have wired in.
+func scriptAuthorizationRules(
+	t *testing.T,
+	rules []common.UtxoValidationRuleFunc,
+) []common.UtxoValidationRuleFunc {
+	t.Helper()
+	out := make([]common.UtxoValidationRuleFunc, 0, 2)
+	for _, rule := range rules {
+		name := runtime.FuncForPC(reflect.ValueOf(rule).Pointer()).Name()
+		if strings.HasSuffix(name, ".UtxoValidateScriptWitnesses") ||
+			strings.HasSuffix(name, ".UtxoValidateNativeScripts") {
+			out = append(out, rule)
+		}
+	}
+	require.Len(t, out, 2, "authorization rules are not registered")
+	return out
+}
+
+// A native script some purpose of the transaction requires must be evaluated
+// wherever the transaction supplies it. The rule used to read only
+// witnesses.NativeScripts(), so a script delivered as a reference script --
+// on a reference input or on the spent input itself, both permitted by
+// CIP-33 -- was never evaluated and an unsatisfied one passed phase-1.
+// cardano-ledger evaluates the needed native scripts out of the scripts the
+// whole transaction provides (Cardano.Ledger.Babbage.Rules.Utxow,
+// validateFailedBabbageScripts).
+func TestReferenceProvidedNativeScriptsAreValidated(t *testing.T) {
+	vkey := bytes.Repeat([]byte{0x61}, 32)
+	nativeScript := testPubkeyNativeScript(t, vkey)
+	scriptHash := nativeScript.Hash()
+	scriptAddr := testScriptPaymentAddress(t, scriptHash)
+
+	// An unrelated script parked on a reference input the transaction names
+	// but no purpose requires. It can never be satisfied, so evaluating it
+	// would reject a transaction the ledger accepts.
+	unusedVkey := bytes.Repeat([]byte{0x62}, 32)
+	unusedScript := testPubkeyNativeScript(t, unusedVkey)
+	keyAddr := testKeyPaymentAddress(t, vkey)
+
+	spentInput := shelley.NewShelleyTransactionInput(
+		"1111111111111111111111111111111111111111111111111111111111111111",
+		0,
+	)
+	refInput := shelley.NewShelleyTransactionInput(
+		"2222222222222222222222222222222222222222222222222222222222222222",
+		0,
+	)
+
+	witnessedTx := func(
+		signed bool,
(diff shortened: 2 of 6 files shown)


ledger/common/script/native.go
```diff
@@ -0,0 +1,161 @@
...
+// returned unchanged.
+func NewTxScriptViewSkippingUnresolved(
+	tx lcommon.Transaction,
+	ls lcommon.LedgerState,
+) (TxScriptView, error) {
+	view, err := NewTxScriptView(tx, ls)
+	if err != nil {
+		if errors.Is(err, lcommon.ErrInputResolution) ||
+			errors.Is(err, lcommon.ErrReferenceInputResolution) {
+			return TxScriptView{}, nil
```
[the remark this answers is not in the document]
chrisguiney: Checked and not a defect, but the reasoning was undocumented, so `585f342` adds `TestUnresolvableInputsAreRejectedBeforeNativeScripts` to pin it.

An unresolvable input cannot reach the native-script rule with the transaction still live. All three eras register, in this order:

| # | Babbage | Conway | Dijkstra |
|---|---|---|---|
| 1 | `UtxoValidateBadInputsUtxo` | same | `conway.UtxoValidateBadInputsUtxo` |
| 2 | `UtxoValidateScriptWitnesses` | same | `conway.UtxoValidateScriptWitnesses` |
| 3 | `UtxoValidateNativeScripts` | same | own |

`shelley.UtxoValidateBadInputsUtxo` rejects an unresolvable *consumed* input. It only walks `tx.Inputs()`, so it does not cover a reference input — `common.ValidateScriptWitnesses` does, returning `ReferenceInputResolutionError`. Both run before the native-script rule, so the empty view is reached only on a transaction that is already being rejected. The new test asserts exactly that for both shapes in all three eras.

On the suggested remedy: a partial view would make `Needed` depend on which `UtxoById` lookups happened to succeed. `neededScripts` decides a spending purpose from the resolved output's address, so a consumed input that failed to resolve silently drops its script from `Needed` — an *under*-approximation, and one that varies with lookup outcome. For a rule that has to reach the same verdict on every node, one authoritative reporter of input-resolution failure is the safer shape, which is the choice `babbage.UtxoValidateInlineDatumsWithPlutusV1` already made before this PR; `script.NewTxScriptViewSkippingUnresolved` just gives that idiom one home instead of two.

Auditing the rest of the class, the three callers of the view differ deliberately: the V1 datum rule and the native-script rules skip, while `shelley.UtxoValidateWithdrawals` propagates the resolution error. Both are safe under the ordering above.

chrisguiney:
Bot review status: CodeRabbit reviewed `b432d18..5ae4afb` (6 files, run `5bad4c1d`) and generated no actionable comments. Cubic raised one P2 on the empty-view fallback in `script.NewTxScriptViewSkippingUnresolved`; answered in the thread and pinned by `TestUnresolvableInputsAreRejectedBeforeNativeScripts` in `585f342`. CodeRabbit's plan allows one included review per hour and reported 0 remaining after that run, so its pass on `585f342` may be rate-limited.

One pre-existing divergence this PR widens, flagged for the reviewer rather than changed here.

`script.ScriptPurposeCertifying.ScriptHash()` treats a Conway `RegistrationCertificate` with a script credential as needing its script, while `common.ValidateScriptWitnesses` treats the same script as *optional* (registration needs no authorization). Verified against this branch:

| transaction | `view.Needed` has the script | `ValidateScriptWitnesses` |
|---|---|---|
| `RegistrationCertificate` (Conway `reg_cert`) | yes | no error, script optional |
| `StakeRegistrationCertificate` (Shelley) | no | no error, script optional |

cardano-ledger agrees with the optional reading: `getScriptWitnessTxCert` returns `Nothing` for `ConwayRegCert`, so the script is neither required nor evaluated.

The effect is bounded and not new in kind. Before this PR the same script was already evaluated when it arrived in the witness set, since the old rule ran every witness native script regardless of need; this PR extends that to the reference-provided delivery of the same shape. Narrowing it belongs in `ScriptPurposeCertifying`, which also feeds `shelley.UtxoValidateWithdrawals`' Dijkstra partial-withdrawal gate and `babbage.UtxoValidateInlineDatumsWithPlutusV1`, so changing it needs conformance coverage those two rules do not currently have. Happy to file it as a follow-up issue if you agree it is worth one.

chrisguiney:
@coderabbitai full review

The CodeRabbit check on `585f342` reports "Review skipped: incremental reviews are disabled", so the follow-up test commit has not been reviewed. Requesting a full pass over the branch.

chrisguiney:
@coderabbitai full review

chrisguiney:
Bot review complete — correcting my earlier note, which said CodeRabbit's pass on `585f342` might stay rate-limited.

| bot | scope | result |
|---|---|---|
| CodeRabbit | full review, `b432d18..585f342`, all 6 files (run `3b76c147`) | no actionable comments |
| Cubic | `5ae4afb`, 6 files | 1 P2 — answered, disputed, pinned by a test |
| Cubic | `585f342`, 1 file | 0 issues |

The first CodeRabbit attempt was refused ("Review limit reached", action not completed). Its reset was 1 minute away rather than the usual hour, so a second `full review` after the window rolled over produced a real review of the current head. All 14 CI checks pass.

The Cubic thread on `native.go:80` is deliberately left unresolved: I disputed the finding rather than applying the suggested fix, so the reviewer should see both the finding and the reasoning. Ready for human review.

ledger/babbage/rules.go
```diff
@@ -1067,14 +1064,41 @@ func UtxoValidateScriptWitnesses(
...
+// to satisfy.
+//
+// Babbage is the first era in which a script can reach a transaction as a
+// reference script rather than a witness, so unlike the Alonzo rule this
+// replaces, the scripts to evaluate come from the resolved transaction view
+// rather than the witness set alone: a native script some script purpose
+// requires counts whether the witness set, a reference input, or the spent
+// input's own reference-script field supplies it -- the same three sources
+// UtxoValidateScriptWitnesses accepts a required script from, so a script that
+// rule counts as provided is a script this rule evaluates rather than ignores.
```
wolf31o2, line 1077: This overstates the alignment. `validateFailedBabbageScripts` restricts to needed ∩ provided:

```haskell
let scriptIsNeeded = scriptHash `Set.member` neededHashes
    scriptDoesNotValidate = not (validateNativeScript tx nativeScript)
 in scriptIsNeeded && scriptDoesNotValidate
```

`NativeScriptsToEvaluate` includes every witness-set native script, needed or not, then adds the needed reference-supplied ones. That matches Alonzo and earlier, and matches main, so it is not a regression and there is no accept/reject divergence -- `ValidateScriptWitnesses` rejects an unneeded witness script as extraneous and runs first, which is also how Haskell rejects it (`ExtraneousScriptWitnessesUTXOW`). Only the error identity differs.

Suggest saying the needed subset is added to the witness set rather than substituted for it, since "the same three sources `UtxoValidateScriptWitnesses` accepts a required script from" reads as a set equality that does not hold.

ledger/dijkstra/rules.go
```diff
@@ -1608,45 +1608,37 @@ func scriptRefIsNativeHash(
 	}
 }
 
+// UtxoValidateNativeScripts evaluates the native scripts this transaction has
+// to satisfy, with Dijkstra's guard credentials in scope for RequireGuard.
+//
+// The scripts to evaluate come from the resolved transaction view, as in
+// Babbage: a native script some script purpose requires counts whether the
```
wolf31o2, line 1615: "a native script some script purpose requires" is not yet true for the guard purpose. `neededScripts` walks spend, mint, cert, reward, vote and propose; there is no `ScriptPurposeGuarding` walk, though `ScriptPurposeGuarding` exists.

So a Dijkstra transaction whose `TxGuards.Credentials[i]` is a script credential H, where H's native script is present only as a reference script on a reference input and is unsatisfiable, is not evaluated here -- and `dijkstraGuardNeedsPlutusRedeemer` already returns "no Plutus redeemer needed" once `dijkstraGuardResolvesToNativeScript` accepts a `ScriptRef`-supplied one. The guard is bypassed.

Pre-existing (main evaluated witness-set scripts only) and Dijkstra is unreleased, so not blocking. Either narrow this comment and file a follow-up, or add the guard walk to `neededScripts` gated to Dijkstra.

ledger/common/script/native.go
```diff
@@ -0,0 +1,161 @@
...
+			env.KeyHashes[lcommon.Blake2b224Hash(bw.PublicKey)] = true
+		}
+	}
+	return env
+}
+
+// NewTxScriptViewSkippingUnresolved builds the transaction's script view and
+// reports an unresolvable input as an empty view rather than an error.
+//
+// UtxoValidateBadInputsUtxo already reports an unresolvable input with the
```
wolf31o2, line 68: `UtxoValidateBadInputsUtxo` covers consumed inputs only -- `shelley.UtxoValidateBadInputsUtxo` walks `tx.Inputs()` alone. For a reference input the rule named in the test's version of this rationale, `UtxoValidateScriptWitnesses`, returns nil immediately when `!tx.IsValid()` (`ledger/common/rules.go:268-270`), so on an `IsValid=false` transaction with an unresolvable reference input, `BadInputsUtxo`, `ScriptWitnesses` and `NativeScripts` all return nil.

The invariant does hold, via a different rule: `UtxoValidateRedeemerAndScriptWitnesses` is registered at index 6 in Babbage, Conway and Dijkstra, ahead of the native-script rule, and fails closed on reference-input resolution regardless of `IsValid`. Worth naming that rule here instead.

The decision itself is right: a partial view would make the needed set depend on which lookups happened to succeed.

ledger/common/reference_native_script_test.go
```diff
@@ -0,0 +1,420 @@
...
+			}
+		})
+	}
+}
+
+// The native-script rule reads an empty script view when an input cannot be
+// resolved, so it evaluates the witness set alone and contributes no
+// reference-provided script. That is safe only because a transaction with an
+// unresolvable input is rejected anyway, by a rule every era registers ahead
+// of the native-script rule: UtxoValidateBadInputsUtxo for a consumed input,
```
wolf31o2, line 320: Same attribution issue as the rationale in `native.go`: `UtxoValidateScriptWitnesses` does not cover the reference-input case for an `IsValid=false` transaction, because `common.ValidateScriptWitnesses` returns nil for those. The guarantee comes from `UtxoValidateRedeemerAndScriptWitnesses`.

`TestUnresolvableInputsAreRejectedBeforeNativeScripts` currently pins only the `IsValid=true` path; adding the `IsValid=false` unresolvable-reference-input case would pin the rule that actually provides the guarantee.

Also worth adding at rule level, since the fix's whole premise is coverage across purposes: a mint-policy or withdrawal script supplied only by a reference input. Both work today, but nothing in the suite pins them.

wolf31o2 (approved):
Approved.

Correct beyond what the tests cover: needed scripts are computed over all six purposes, and a mint-policy or withdrawal script supplied only by a reference input is now rejected when unsatisfied. `NewNativeScriptEnv` reproduces `evalTimelock`, including the absent upper bound. The failing-hash order is deterministic and the sort is total, since `seen` dedups by hash before sorting. All rule slices are byte-identical to `origin/main`.

Revert proof: restoring the three `rules.go` files from `origin/main` fails exactly 6 subtests, and that run compiled.

Three doc comments claim more than the code does; inline.

ledger/common/script/native.go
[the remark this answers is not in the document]
wolf31o2: Confirmed not a defect, and the disposition is right: a partial view would make the needed-script set depend on which `UtxoById` lookups happened to succeed, because `neededScripts` derives the spend purpose from the resolved output's address. An outcome-dependent under-approximation is strictly worse for a rule that has to reach the same verdict on every node. Verified separately that the empty-view branch is reachable when the rule is called in isolation, and that no caller in the workspace does so -- Dingo consumes the full ordered slice.

One correction to the ordering table: `UtxoValidateScriptWitnesses` does not cover the reference-input case for an `IsValid=false` transaction. `common.ValidateScriptWitnesses` returns nil at `ledger/common/rules.go:268-270` when `!tx.IsValid()`, and `shelley.UtxoValidateBadInputsUtxo` walks only `tx.Inputs()`, so on such a transaction with an unresolvable reference input all three of `BadInputsUtxo`, `ScriptWitnesses` and `NativeScripts` return nil.

The invariant still holds, through `UtxoValidateRedeemerAndScriptWitnesses` at index 6 in all three eras, which fails closed on reference-input resolution regardless of `IsValid`. Worth naming that rule in the rationale and extending `TestUnresolvableInputsAreRejectedBeforeNativeScripts` to the `IsValid=false` case, which it does not currently pin.

ledger/common/script/native.go
wolf31o2, line 68: The rest of this argument (`UtxoValidateScriptWitnesses` covers the reference-input half) fails when `tx.IsValid()` is false: `common.ValidateScriptWitnesses` returns nil right after `ValidateWithdrawalAddresses`, and `shelley.UtxoValidateBadInputsUtxo` iterates only `tx.Inputs()`.

Proven against this head with `BadInputsUtxo` + `ScriptWitnesses` + `NativeScripts` selected out of each era's production slice: an `IsValid=false` transaction with one unresolvable reference input returns nil in Babbage, Conway and Dijkstra. With a second, resolvable reference input carrying a needed-but-unsatisfied native script, the same transaction also returns nil, so the whole-view drop re-opens #2149 for that transaction shape.

Bounded in practice — `IsValid=false` declares phase-2 failure, so inputs are not consumed — but the comment states an invariant that does not hold. Suggest narrowing it to the `IsValid=true` case and filing the reference-input coverage gap in `UtxoValidateBadInputsUtxo` separately.

ledger/common/reference_native_script_test.go
```diff
@@ -0,0 +1,420 @@
...
+// UtxoValidateScriptWitnesses for a reference input, which
```
wolf31o2, line 321: `mockledger.NewTransactionBuilder()` defaults `valid` to true, so both subtests run with `IsValid=true` and this branch of the claim is never exercised. Adding `tx.WithValid(false)` to the reference-input case makes it fail in all three eras: nothing rejects the transaction.

Exact missing case: `IsValid=false`, one resolvable script-address input, one unresolvable reference input — expected reject, currently accepted.

ledger/common/script/native.go
```diff
@@ -0,0 +1,161 @@
...
+			return TxScriptView{}, nil
+		}
+		return TxScriptView{}, err
+	}
+	return view, nil
+}
+
+// NativeScriptsToEvaluate returns the native scripts a phase-1 rule must run
+// for this transaction: every native script in the witness set, plus every
+// native script some script purpose requires that only a reference script
```
wolf31o2, line 89: Nit: `view.Needed` comes from `availableScripts`, which also folds in `SubTransactionWitnessSetsFromTransaction`. So a needed native script supplied only by a Dijkstra sub-transaction witness set is returned here too, and evaluated against `NewNativeScriptEnv(tx, slot)`, whose key hashes come from `tx.Witnesses()` alone. Unreachable today because `conway.UtxoValidateScriptWitnesses` runs first and counts only `tx.Witnesses()` as explicitly provided — except under the `IsValid=false` early return above. Worth saying "that the witness set does not supply" rather than "that only a reference script supplies".

wolf31o2 (commented):
Fix and tests are correct; nothing blocking. Verified fails-without-fix locally: reverting the Babbage body to `alonzo.UtxoValidateNativeScripts` fails the two unsatisfied subtests in Babbage and Conway while the three accept-cases still pass.

One non-blocking correction: the unresolvable-input safety argument documented in `native.go` and pinned by `TestUnresolvableInputsAreRejectedBeforeNativeScripts` does not hold for `IsValid=false`. Details inline. Underlying gap (`UtxoValidateBadInputsUtxo` ignores reference inputs) is pre-existing and belongs in a separate issue.

PR #2163, opened 2026-09-01
fix(ledger): enforce required redeemers for reference-script plutus spends

### LINKED ISSUE #2147: ledger: enforce required redeemers for reference-script Plutus spends
## Scope
Ensure UTxO validation rejects Plutus script spends that provide a reference script but no corresponding redeemer, across all supported eras including Dijkstra.

## Acceptance criteria
- Shared ledger validation enforces the required-redeemer implication for reference-script inputs.
- Conway and Dijkstra entry points behave consistently.
- Regression tests cover missing redeemer, valid redeemer, and reference-script/non-reference-script paths.
- Normal, race, vet, lint, and conformance checks pass.

## Compatibility
Preserve existing valid transaction behavior and avoid duplicating era-specific guards.

arepala-uml:
### Summary
- Adds a single shared validation function that checks every Plutus script-address
  input — whether its script is provided as an explicit witness or a CIP-33
  reference script — has a matching spend redeemer at its canonical sorted-input
  position.
- Closes a gap where a reference script satisfied existing "script is present"
  checks without ever requiring a redeemer, letting a script-locked input spend
  with no script execution and no error at all.
- Wired into Babbage, Conway, and Dijkstra via one implementation — no
  duplicated era-specific guards.

### Changes by file

**`ledger/common/script/redeemers.go`** (new)
- `ValidateRequiredRedeemers(tx, ls)`: walks `SortInputs(tx.Inputs())`, and for
  each script-address input whose resolved script (witness or reference) is
  Plutus, requires a spend redeemer at that index. Skips invalid txs, nil
  ledger state, and non-Plutus (native) scripts. Lives in the `script`
  subpackage since it reuses `SortInputs`/`NewTxScriptView` (avoids an import
  cycle with `ledger/common`).

**`ledger/common/errors.go`**
- Added `MissingRedeemerForScriptError` (mirrors `ExtraneousRedeemerError`'s
  style), returned by the new check.

**`ledger/babbage/errors.go`**
- Added `MissingRedeemerForScriptError` as an alias to the common type.

**`ledger/babbage/rules.go`**
- Added `UtxoValidateRequiredRedeemers` wrapper (delegates to
  `script.ValidateRequiredRedeemers`), registered in `UtxoValidationRules`
  right after `UtxoValidateScriptWitnesses`. Babbage doesn't execute Plutus,
  but must still reject an unexecuted reference-script-backed spend.

**`ledger/conway/errors.go`**
- Converted the pre-existing, never-wired-up `MissingRedeemerForScriptError`
  into a type alias to `common.MissingRedeemerForScriptError` (was dead code
  since PR #1418).

**`ledger/conway/rules.go`**
- Added `UtxoValidateRequiredRedeemers` wrapper, registered in
  `UtxoValidationRules` right after `UtxoValidateScriptWitnesses`.

**`ledger/dijkstra/rules.go`**
- Registered `conway.UtxoValidateRequiredRedeemers` directly in
  `UtxoValidationRules` (same reuse pattern as `conway.UtxoValidateScriptWitnesses`)
  — no Dijkstra-specific wrapper needed.

**`ledger/common/script/redeemers_test.go`** (new)
- Unit tests for the shared function: missing/valid redeemer via reference
  script (both attached-to-spent-output and separate-reference-input shapes),
  missing redeemer via explicit witness script, mixed script inputs (only one
  redeemed), non-script address skipped, native reference script skipped,
  invalid tx skipped, nil ledger state skipped.

**`ledger/babbage/required_redeemers_test.go`** (new)
- Wiring test (`UtxoValidateRequiredRedeemers` present in
  `UtxoValidationRules`) + missing/valid redeemer regression test.

**`ledger/conway/rules_test.go`**
- Added wiring test + missing/valid/non-script regression tests for
  `conway.UtxoValidateRequiredRedeemers`.

**`ledger/dijkstra/required_redeemers_test.go`** (new)
- Confirms Dijkstra registers Conway's shared function and behaves
  identically through the Dijkstra transaction type.

### Testing
- `go build ./...`, `go vet ./...` — clean
- `golangci-lint run ./...` — 0 issues
- `go test ./...` — all pass
- `go test -race ./ledger/...` — all pass
- Ledger rule conformance — 315/315 vectors pass (unchanged from baseline)

Closes #2147 

---
## Summary by cubic
Requires every Plutus script-address spend — whether backed by an explicit witness or a CIP-33 reference script — to include its matching spend redeemer across Babbage, Conway, and Dijkstra, including transactions marked invalid (`IsValid=false`). Previously, a reference script could satisfy script-presence checks without a redeemer, letting a Plutus-locked input spend unexecuted. Closes #2147.

**Bug Fixes**

- Adds shared validation using canonical sorted-input indexes and `MissingRedeemerForScriptError`.
- Adds regression coverage for reference scripts, witness scripts, mixed inputs, missing scripts, native scripts, invalid transactions, and PlutusV3/V4.

<sup>Written for commit 75e91a1cdc859cb3f47e9801f58e7573c12f675c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2163?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Transactions spending Plutus script addresses now require the appropriate spend redeemer.
  - This validation applies to scripts supplied directly or through reference scripts.
  - Transactions missing a required redeemer are rejected with a clear validation error identifying the affected script and input.
  - Key-address and native-script transactions remain unaffected.
  - Consistent redeemer validation is now enforced across Babbage, Conway, and Dijkstra transaction validation.

### DIFF
--- a/ledger/common/script/redeemers.go
+++ b/ledger/common/script/redeemers.go
@@ -1,0 +1,96 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package script
+
+import (
+	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
+)
+
+// ValidateRequiredRedeemers checks that every script-address input whose
+// spending script is Plutus and resolvable -- whether provided as an
+// explicit witness script or as a CIP-33 reference script -- has a matching
+// spend redeemer at its canonical sorted-input position.
+//
+// This closes a gap left by the existing script/redeemer checks:
+// ValidateScriptWitnesses only confirms the script itself is reachable
+// (explicit witness or reference script), and UtxoValidatePlutusScripts only
+// executes redeemers that already exist in the witness set. A reference
+// script satisfies the former without ever requiring a redeemer, so a
+// script-locked input backed by a reference script -- or, just as easily, an
+// explicit witness script missing its own redeemer while other redeemers are
+// present -- previously spent with no script execution and no error at all.
+//
+// Eras share this single implementation rather than each re-deriving the
+// sorted-input walk and its script-availability lookup.
+func ValidateRequiredRedeemers(
+	tx lcommon.Transaction,
+	ls lcommon.LedgerState,
+) error {
+	if ls == nil || !tx.IsValid() {
+		return nil
+	}
+	view, err := NewTxScriptView(tx, ls)
+	if err != nil {
+		return err
+	}
+	if len(view.Available) == 0 {
+		return nil
+	}
+	var redeemers lcommon.TransactionWitnessRedeemers
+	if wits := tx.Witnesses(); wits != nil {
+		redeemers = wits.Redeemers()
+	}
+	spendIndexes := make(map[uint32]struct{})
+	if redeemers != nil {
+		for _, idx := range redeemers.Indexes(lcommon.RedeemerTagSpend) {
+			spendIndexes[uint32(idx)] = struct{}{} // #nosec G115 -- input count is bounded
+		}
+	}
+	byId := make(map[string]lcommon.Utxo, len(view.ResolvedInputs))
+	for _, utxo := range view.ResolvedInputs {
+		if utxo.Id != nil {
+			byId[utxo.Id.String()] = utxo
+		}
+	}
+	for idx, input := range SortInputs(tx.Inputs()) {
+		utxo, ok := byId[input.String()]
+		if !ok || utxo.Output == nil {
+			continue
+		}
+		addr := utxo.Output.Address()
+		if addr.Type()&lcommon.AddressTypeScriptBit == 0 {
+			continue
+		}
+		scriptHash := lcommon.ScriptHash(addr.PaymentKeyHash())
+		s, available := view.Available[scriptHash]
+		if !available {
+			// A script that is missing entirely is
+			// ValidateScriptWitnesses's error to report, not this check's.
+			continue
+		}
+		if _, isPlutus := lcommon.PlutusScriptVersion(s); !isPlutus {
+			continue
+		}
+		index := uint32(idx) // #nosec G115 -- input count is bounded
+		if _, ok := spendIndexes[index]; !ok {
+			return lcommon.MissingRedeemerForScriptError{
+				ScriptHash: scriptHash,
+				Tag:        lcommon.RedeemerTagSpend,
+				Index:      index,
+			}
+		}
+	}
+	return nil
+}
--- a/ledger/common/script/redeemers.go
+++ b/ledger/common/script/redeemers.go
@@ -34,13 +34,31 @@ import (
 //
 // Eras share this single implementation rather than each re-deriving the
 // sorted-input walk and its script-availability lookup.
+//
+// This check applies regardless of the transaction's IsValid flag.
+// AlonzoUTXOW's hasExactSetOfRedeemers (the source of AlonzoUtxowPredFailure
+// MissingRedeemers) is a witnessing-completeness check, not a phase-2
+// execution check: IsValid is read only by UTXOS to decide how an already
+// phase-1-valid transaction is applied to the ledger state, never by UTXOW to
+// decide which witnesses are required. A transaction the submitter marked
+// invalid still must carry every redeemer its script-locked inputs require.
 func ValidateRequiredRedeemers(
 	tx lcommon.Transaction,
 	ls lcommon.LedgerState,
 ) error {
-	if tx == nil || ls == nil || !tx.IsValid() {
+	if tx == nil || ls == nil {
 		return nil
 	}
+	// NewTxScriptView resolves every input, so an unresolvable one surfaces
+	// here as InputResolutionError/ReferenceInputResolutionError even though
+	// shelley.UtxoValidateBadInputsUtxo and
+	// common.ValidateRedeemerAndScriptWitnesses -- both registered ahead of
+	// this rule in every era's rule list -- already report the same failure
+	// more specifically. Not reachable in production today because of that
+	// ordering, but reachable by any caller that runs this check first.
+	// TODO(#2162): switch to NewTxScriptViewSkippingUnresolved once it
+	// lands, so this rule -- which only ever reads the resolved view --
+	// stops being a second, competing source of input-resolution failures.
 	view, err := NewTxScriptView(tx, ls)
 	if err != nil {
 		return err
--- a/ledger/common/script/redeemers.go
+++ b/ledger/common/script/redeemers.go
@@ -38,7 +38,7 @@ func ValidateRequiredRedeemers(
 	tx lcommon.Transaction,
 	ls lcommon.LedgerState,
 ) error {
-	if ls == nil || !tx.IsValid() {
+	if tx == nil || ls == nil || !tx.IsValid() {
 		return nil
 	}
 	view, err := NewTxScriptView(tx, ls)
--- a/ledger/common/script/redeemers_test.go
+++ b/ledger/common/script/redeemers_test.go
@@ -314,6 +314,74 @@ func TestValidateRequiredRedeemersMixedInputs(t *testing.T) {
 	require.Equal(t, uint32(1), missingErr.Index)
 }
 
+// TestValidateRequiredRedeemersMixedInputsBothRedeemed is
+// TestValidateRequiredRedeemersMixedInputs's positive counterpart: the same
+// two script-address inputs (one witness-backed, one reference-backed),
+// each correctly redeemed at its own sorted-input index, must pass.
+func TestValidateRequiredRedeemersMixedInputsBothRedeemed(t *testing.T) {
+	v1 := common.PlutusV1Script{0x01}
+	v2 := common.PlutusV2Script{0x02}
+
+	// Sorts after "aaaa...", so v1's input lands at sorted index 1.
+	inputWitness := shelley.NewShelleyTransactionInput(
+		"bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb",
+		0,
+	)
+	inputRef := shelley.NewShelleyTransactionInput(
+		"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+		0,
+	)
+
+	utxoWitness := common.Utxo{
+		Id: inputWitness,
+		Output: &babbage.BabbageTransactionOutput{
+			OutputAddress: scriptAddress(t, v1),
+			OutputAmount:  mary.MaryTransactionOutputValue{Amount: 1000},
+		},
+	}
+	utxoRef := common.Utxo{
+		Id: inputRef,
+		Output: &babbage.BabbageTransactionOutput{
+			OutputAddress: scriptAddress(t, v2),
+			OutputAmount:  mary.MaryTransactionOutputValue{Amount: 1000},
+			TxOutScriptRef: &common.ScriptRef{
+				Type:   common.ScriptRefTypePlutusV2,
+				Script: v2,
+			},
+		},
+	}
+	ls := ledgerStateWithUtxos(utxoWitness, utxoRef)
+
+	tx := &conway.ConwayTransaction{
+		Body: conway.ConwayTransactionBody{
+			TxInputs: conway.NewConwayTransactionInputSet(
+				[]shelley.ShelleyTransactionInput{inputWitness, inputRef},
+			),
+		},
+		WitnessSet: conway.ConwayTransactionWitnessSet{
+			WsPlutusV1Scripts: cbor.NewSetType(
+				[]common.PlutusV1Script{v1},
+				false,
+			),
+			WsRedeemers: conway.ConwayRedeemers{
+				Redeemers: map[common.RedeemerKey]common.RedeemerValue{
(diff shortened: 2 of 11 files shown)


arepala-uml:
@cubic-dev-ai

ledger/common/script/redeemers.go
```diff
@@ -0,0 +1,96 @@
...
+// explicit witness script missing its own redeemer while other redeemers are
+// present -- previously spent with no script execution and no error at all.
+//
+// Eras share this single implementation rather than each re-deriving the
+// sorted-input walk and its script-availability lookup.
+func ValidateRequiredRedeemers(
+	tx lcommon.Transaction,
+	ls lcommon.LedgerState,
+) error {
+	if tx == nil || ls == nil || !tx.IsValid() {
```
wolf31o2: Merge blocker. `!tx.IsValid()` gates a phase-1 UTXOW condition on a phase-2 flag.

`MissingRedeemers` comes from `hasExactSetOfRedeemers` in UTXOW; `isValid` is read only by UTXOS. So `isValid == false` + a Plutus reference-script-locked input + no spend redeemer is rejected by cardano-node and accepted here.

Dingo's `validateConwayRequiredPlutusRedeemers` is placed above its own `if !tx.IsValid()` skip for exactly this reason. Either drop the term, or keep it and document the deviation with an issue reference so the two implementations do not silently disagree.

The `tx == nil || ls == nil` guards are right and address the earlier nil-panic finding.

ledger/common/script/redeemers.go
```diff
@@ -0,0 +1,96 @@
...
+		return nil
+	}
+	view, err := NewTxScriptView(tx, ls)
```
wolf31o2, line 62: Consistency with #2162, which introduces `NewTxScriptViewSkippingUnresolved` for precisely this position, on the rationale that "a rule that only reads the resolved view must not become a second, competing source of input-resolution failures".

Not reachable today: `shelley.UtxoValidateBadInputsUtxo` covers consumed inputs and `common.ValidateRedeemerAndScriptWitnesses` covers reference inputs, both registered ahead of this rule, and this function's own `tx == nil || ls == nil || !tx.IsValid()` guard mirrors theirs. It becomes reachable if the rule is ever moved earlier. Worth switching once #2162 lands.

ledger/common/script/redeemers_test.go
```diff
@@ -0,0 +1,444 @@
...
+	}
+
+	require.NoError(t, script.ValidateRequiredRedeemers(tx, ls))
+}
+
+// TestValidateRequiredRedeemersSkipsInvalidTx mirrors
+// ValidateScriptWitnesses: a phase-2-invalid transaction is expected to fail
+// phase-2 validation, so phase-1 checks like this one must not additionally
+// reject it for a missing redeemer.
+func TestValidateRequiredRedeemersSkipsInvalidTx(t *testing.T) {
```
wolf31o2: This test is what turns a shared pre-existing convention into an asserted deviation from `AlonzoUtxowPredFailure.MissingRedeemers`. The doc comment states the phase-2 rationale as settled, but UTXOW does not gate on `isValid`, and Dingo's implementation of this same rule asserts the opposite.

If the skip stays, this comment should say it encodes a known divergence and cite an issue.

Absence cases missing elsewhere in this file, each a distinct branch:
- script available from neither witness nor reference (`redeemers.go:78-82`) returning nil rather than reporting a missing redeemer;
- the resolution-error contract at `redeemers.go:44-47`, pinned either way;
- two Plutus script inputs both correctly redeemed at sorted indices 0 and 1 (`MixedInputs` covers only the negative);
- a native script from the witness set, not only via a reference script;
- a `PlutusV3Script` or `PlutusV4Script` case; only V1 and V2 appear.

wolf31o2 (changes requested):
Requesting changes on one item.

`!tx.IsValid()` at `redeemers.go:41` gates a phase-1 UTXOW condition on a phase-2 flag. `MissingRedeemers` comes from `hasExactSetOfRedeemers` in UTXOW; `isValid` is read only by UTXOS. A transaction with `isValid == false`, a Plutus reference-script-locked input and no spend redeemer is rejected by cardano-node and accepted here. `TestValidateRequiredRedeemersSkipsInvalidTx` pins that as intended.

Verified correct: the spend index basis matches `rdptr`/`Ord TxIn` and every other spend-pointer resolution in this repo; the `conway.MissingRedeemerForScriptError` alias is behavior-compatible; the behavioral tests fail without the fix.

On the conformance claim: `ouroboros-mock@v0.17.0` `ConformanceValidationRules` does not contain this rule and the harness hardcodes that slice, so 315/315 says nothing about it. Short-circuiting `ValidateRequiredRedeemers` to `return nil` still yields 315/315.

ledger/common/script/redeemers.go
arepala-uml: Fixed in d6204e6: dropped `!tx.IsValid()` entirely rather than documenting it as a deviation. You're right that `hasExactSetOfRedeemers` is a UTXOW witnessing-completeness check and `isValid` is read only by UTXOS — there's no ledger-spec basis for gating this on the flag. The guard is now just `tx == nil || ls == nil`, and the doc comment states the AlonzoUTXOW rationale directly instead of citing a (wrong) phase-1/phase-2 split.

`TestValidateRequiredRedeemersSkipsInvalidTx` is gone; replaced with `TestValidateRequiredRedeemersAppliesRegardlessOfIsValid`, which asserts `MissingRedeemerForScriptError` IS still returned for an `isValid == false` tx. Ran the full suite after the change — that was the only test that needed to flip; nothing else in the repo depended on the old gating.

ledger/common/script/redeemers.go
arepala-uml: Agreed, and it's not implementable yet: `NewTxScriptViewSkippingUnresolved` doesn't exist in this repo (#2162 hasn't landed) — confirmed by grep before touching anything here.

Added a `TODO(#2162)` comment at this call site spelling out today's contract (resolution errors surface here, redundantly with `UtxoValidateBadInputsUtxo`/`ValidateRedeemerAndScriptWitnesses` which are registered ahead of this rule in every era) and the planned switch once it lands.

Also added `TestValidateRequiredRedeemersPropagatesInputResolutionError`, which pins the current behavior (an unresolvable input surfaces as `common.InputResolutionError`) explicitly, so the future swap to skip-unresolved semantics has to touch this test deliberately rather than changing behavior silently underneath it.

ledger/common/script/redeemers_test.go
arepala-uml: Renamed/flipped: `TestValidateRequiredRedeemersSkipsInvalidTx` is now `TestValidateRequiredRedeemersAppliesRegardlessOfIsValid`, asserting the error IS returned for an `isValid == false` tx, with a doc comment stating the AlonzoUTXOW rationale instead of the old (incorrect) phase-1/phase-2 framing.

Added all five missing branches you listed:
- script available from neither witness nor reference → `TestValidateRequiredRedeemersSkipsScriptEntirelyMissing`
- the resolution-error contract → `TestValidateRequiredRedeemersPropagatesInputResolutionError`
- both script inputs correctly redeemed at sorted indices 0 and 1 → `TestValidateRequiredRedeemersMixedInputsBothRedeemed`
- native script from the witness set (not just reference) → `TestValidateRequiredRedeemersSkipsNativeWitnessScript`
- PlutusV3/PlutusV4 → `TestValidateRequiredRedeemersMissingRedeemerPlutusV3AndV4` (table-driven, both versions)

All in d6204e6. Full suite (build/vet/lint/test/-race/conformance) still green.

arepala-uml:
Pushed d6204e6 addressing the requested changes:

- Dropped `!tx.IsValid()` from `ValidateRequiredRedeemers` — confirmed the ledger-spec basis is wrong (UTXOW's redeemer-completeness check isn't gated by `isValid`, which UTXOS reads instead) and fixed it rather than documenting a deviation.
- Added a `TODO(#2162)` at the `NewTxScriptView` call noting the planned switch to `NewTxScriptViewSkippingUnresolved` once that lands, plus a test pinning today's resolution-error contract so that switch is deliberate.
- Added the five missing test branches (script entirely absent, resolution-error propagation, both-inputs-redeemed positive case, native witness script, PlutusV3/V4).

On the conformance point: agreed, and noted for the record — `ouroboros-mock@v0.17.0`'s `ConformanceValidationRules` is a hardcoded external slice that doesn't include this rule, so 315/315 was never evidence this specific check is conformant, only that nothing already-covered regressed. That's an upstream gap (would need a PR to `ouroboros-mock` plus a dependency bump here), out of scope for this PR — flagging rather than leaving the impression it's covered.

Full suite (build/vet/lint/test/-race/conformance) green on top of this commit.

ledger/common/script/redeemers.go
wolf31o2: Confirmed at d6204e6. Verified by re-adding the gate: `TestValidateRequiredRedeemersAppliesRegardlessOfIsValid` is the only failure, and the revert compiled.

ledger/common/script/redeemers.go
wolf31o2: `TODO(#2162)` is the right disposition, and pinning today's behaviour with `TestValidateRequiredRedeemersPropagatesInputResolutionError` means the switch will be a deliberate change rather than a silent one.

ledger/common/script/redeemers_test.go
wolf31o2: Confirmed — it now asserts the specific script hash and index, and the five absence cases are all present.

wolf31o2 (dismissed):
Approved at d6204e6. The blocker is fixed at the cause: the `!tx.IsValid()` gate is gone rather than documented as a deviation, and the doc comment now states the correct rationale — `hasExactSetOfRedeemers` is a UTXOW witnessing-completeness check, `isValid` is read only by UTXOS.

Revert proof: re-adding the gate makes `TestValidateRequiredRedeemersAppliesRegardlessOfIsValid` the only failure, and that run compiled. The test asserts the specific script hash and index rather than just an error.

All five absence cases are in: multi-input both-redeemed, native script from the witness set, script reachable from neither witness nor reference, the input-resolution contract, and PlutusV3/V4.

`NewTxScriptView` staying put with a `TODO(#2162)` is the right call, since `NewTxScriptViewSkippingUnresolved` does not exist in this repo yet and the ordering makes the redundancy unreachable in production.

Merge order still applies: #2162 first, then rebase this over it, at which point the TODO can be closed.

ledger/common/script/redeemers.go
```diff
@@ -0,0 +1,114 @@
...
+	if len(view.Available) == 0 {
+		return nil
+	}
+	var redeemers lcommon.TransactionWitnessRedeemers
+	if wits := tx.Witnesses(); wits != nil {
+		redeemers = wits.Redeemers()
+	}
+	spendIndexes := make(map[uint32]struct{})
+	if redeemers != nil {
+		for _, idx := range redeemers.Indexes(lcommon.RedeemerTagSpend) {
```
wolf31o2, line 75: Sub-transaction asymmetry in Dijkstra: `availableScripts` folds `SubTransactionWitnessSetsFromTransaction(tx)` into `view.Available`, but the redeemer set here is read only from the top-level `tx.Witnesses().Redeemers()`.

Reproduced: a top-level script-address input whose PlutusV3 script is supplied *only* by a sub-transaction witness set, with the spend redeemer also in that sub-transaction witness set and none at top level, is rejected —

```
missing redeemer for script def51d5e19af4b8693ef6966aea6d0dd15499ba03ec40a6156cf99a8: tag=0, index=0
```

Either script availability should be top-level-only for this check, or the redeemer set should include sub-transaction redeemers. Not a mainnet risk (Dijkstra is unreleased), but the contract should be decided rather than inherited from `availableScripts`.

ledger/common/script/redeemers.go
```diff
@@ -0,0 +1,114 @@
...
+			spendIndexes[uint32(idx)] = struct{}{} // #nosec G115 -- input count is bounded
+		}
+	}
+	byId := make(map[string]lcommon.Utxo, len(view.ResolvedInputs))
+	for _, utxo := range view.ResolvedInputs {
+		if utxo.Id != nil {
+			byId[utxo.Id.String()] = utxo
+		}
+	}
+	for idx, input := range SortInputs(tx.Inputs()) {
```
wolf31o2, line 85: Scope, for the record: this covers only the spend purpose. cardano-ledger's `hasExactSetOfRedeemers` covers mint/cert/reward/vote/propose too, and those still have no required-redeemer check in any era here — a Plutus minting policy with no mint redeemer still spends unexecuted. That direction is under-requiring, so no accept-case risk, but #2147's `MissingRedeemers` gap is only partly closed and the remaining purposes need their own issue.

Same loop: `tx.Inputs()` excludes Dijkstra sub-transaction inputs, so a script-address input inside a sub-transaction is never walked.

wolf31o2 (commented):
Reviewed the accept-case risk (over-requiring) as the priority. Nothing blocking.

Verified against cardano-ledger `hasExactSetOfRedeemers` semantics: the neededness filter is right (script must be in `Available` *and* Plutus, so an absent script and a native script both fall through), the sorted-input index space is identical to the one Conway's executor uses at `ledger/conway/rules.go:2808`, era gating is correct (Alonzo untouched; Babbage already fails closed on any redeemer via `ValidateUnsupportedPlutusExecution`), and dropping the `IsValid` gate in d6204e6 is correct since UTXOW never reads `IsValid`. `SortInputs` not deduplicating is unreachable here because `UtxoValidateNoDuplicateInputs` precedes this rule in all three slices. No overlap with #2143/#2162 — neither constructs `MissingRedeemerForScriptError`, and it was dead code on main.

Two non-blocking items inline. One test note: nothing drives this from a production `UtxoValidationRules` slice — the wiring tests only pin registration order. I ran the false-rejection classes through Conway's real slice (native reference script; script-credential `StakeRegistrationCertificate` with no redeemer; Plutus mint-only; correctly-redeemed reference-script spend with declared order != sorted order) and all pass, with the no-redeemer case firing at sorted index 1. Worth adding at least the native-reference-script and script-credential-certificate cases as slice-level accept tests, since those are the classes an over-requiring regression would land in.

wolf31o2 (approved):

PR #2170, opened 2026-09-02
fix(script): validity-interval upper-bound closure is era-dependent (Conway exclusive, pre-Conway closed)

chadaibbhot-git:
## Problem
`TimeRange.ToPlutusData` rendered the validity-interval upper bound with a single closure for all Plutus versions/eras. cardano-ledger's `transValidityInterval` makes the upper-bound closure **era-dependent**: pre-Conway (Alonzo/Babbage) Plutus V1/V2 render a TTL-only (`invalidHereafter`, upper-only) interval as a **closed** upper bound `UpperBound (Finite t) True`, while Conway renders it **exclusive** `UpperBound (Finite t) False` (cardano-ledger#3043). A single closure diverges from the node on one side, causing ExUnits / script-evaluation divergence.

## Fix
Gate the closure on the target era. Thread a `strictValidityUpperBound bool` through `NewTxInfoV1FromTransaction` / `NewTxInfoV2FromTransaction` (Plutus V3 is always strict). `TimeRange` carries a `strictUpperBound` field; the finite upper-bound closure is computed as `!lowerBoundPresent && !strictUpperBound` — closed only for a pre-Conway upper-only interval, exclusive when strict (Conway/Dijkstra) or when both bounds are present. Callers in `ledger/conway/rules.go` and `ledger/dijkstra/rules.go` pass `true`.

## Tests
- New `context_timerange_test.go`: pre-Conway (closed) vs Conway (exclusive) rendering for upper-only, both-bounds, and lower-only intervals.
- `context_validity_test.go`: V1/V2 Alonzo/Babbage reverted to the inclusive expectation.
- `context_test.go`: V3 golden keeps the exclusive bound.

Refs cardano-ledger#3043.

_(Re-files #2159, which GitHub auto-closed after its branch lost common history with `main` — it had been based on a vendored snapshot rather than a fork of main. Same change, rebuilt cleanly on current main.)_

---
## Summary by cubic
Fixes the Plutus validity-interval upper bound so it matches cardano-ledger per era, preventing ExUnits and script-evaluation divergence. Pre-Conway V1/V2 keep a closed upper-only bound; Conway and later now render it exclusive (cardano-ledger#3043).

**Migration**

- `NewTxInfoV1FromTransaction` and `NewTxInfoV2FromTransaction` now take a `strictValidityUpperBound bool` argument; pass `false` for Alonzo/Babbage and `true` for Conway and later.
- `NewTxInfoV3FromTransaction` is unchanged and always uses the exclusive bound.

<sup>Written for commit e3a446a028de705d44a7c1ddef3398c122b31876. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2170?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected validity-interval encoding for script validation.
  - In Conway and later eras, upper-only validity intervals are now treated as exclusive.
  - In Alonzo and Babbage eras, upper-only intervals retain inclusive behavior.
  - Two-sided, lower-only, and unbounded validity intervals continue to use consistent encoding across eras.
  - Added coverage to verify validity-interval behavior across supported eras.

### DIFF
--- a/ledger/common/script/context.go
+++ b/ledger/common/script/context.go
@@ -194,12 +194,18 @@ func (t TxInfoV1) ToPlutusData() data.PlutusData {
 	)
 }
 
+// NewTxInfoV1FromTransaction builds a Plutus V1 TxInfo. strictValidityUpperBound
+// selects the era-dependent encoding of a finite validity-interval upper bound:
+// pass true in the Conway era or later (EXCLUSIVE upper bound in all cases) and
+// false in Alonzo/Babbage (CLOSED upper bound for an upper-only interval). See
+// the TimeRange.strictUpperBound documentation and cardano-ledger#3043.
 func NewTxInfoV1FromTransaction(
 	slotState lcommon.SlotState,
 	tx lcommon.Transaction,
 	resolvedInputs []lcommon.Utxo,
+	strictValidityUpperBound bool,
 ) (TxInfoV1, error) {
-	validityRange, err := validityRangeInfo(slotState, tx)
+	validityRange, err := validityRangeInfo(slotState, tx, strictValidityUpperBound)
 	if err != nil {
 		return TxInfoV1{}, err
 	}
@@ -307,12 +313,18 @@ func (t TxInfoV2) ToPlutusData() data.PlutusData {
 	)
 }
 
+// NewTxInfoV2FromTransaction builds a Plutus V2 TxInfo. strictValidityUpperBound
+// selects the era-dependent encoding of a finite validity-interval upper bound:
+// pass true in the Conway era or later (EXCLUSIVE upper bound in all cases) and
+// false in Babbage (CLOSED upper bound for an upper-only interval). See the
+// TimeRange.strictUpperBound documentation and cardano-ledger#3043.
 func NewTxInfoV2FromTransaction(
 	slotState lcommon.SlotState,
 	tx lcommon.Transaction,
 	resolvedInputs []lcommon.Utxo,
+	strictValidityUpperBound bool,
 ) (TxInfoV2, error) {
-	validityRange, err := validityRangeInfo(slotState, tx)
+	validityRange, err := validityRangeInfo(slotState, tx, strictValidityUpperBound)
 	if err != nil {
 		return TxInfoV2{}, err
 	}
@@ -409,7 +421,9 @@ func NewTxInfoV3FromTransaction(
 	tx lcommon.Transaction,
 	resolvedInputs []lcommon.Utxo,
 ) (TxInfoV3, error) {
-	validityRange, err := validityRangeInfo(slotState, tx)
+	// Plutus V3 only exists in the Conway era and later, where cardano-ledger
+	// always uses an EXCLUSIVE (strict) validity-interval upper bound.
+	validityRange, err := validityRangeInfo(slotState, tx, true)
 	if err != nil {
 		return TxInfoV3{}, err
 	}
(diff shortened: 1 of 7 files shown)


ledger/common/script/context.go
```diff
@@ -517,10 +551,17 @@ func (t TimeRange) ToPlutusData() data.PlutusData {
...
+			// ERA-DEPENDENT translation (see the strictUpperBound field):
+			//   - both bounds present (lowerBoundPresent): EXCLUSIVE (false)
+			//     in every era (transVITime / transValidityInterval both use
+			//     strictUpperBound for a two-sided interval).
+			//   - upper-only interval (no lower bound): EXCLUSIVE (false) in
+			//     Conway and later (Conway.transValidityInterval,
+			//     cardano-ledger#3043), INCLUSIVE (true) in Alonzo/Babbage
+			//     (transVITime uses PV1.to).
+			// i.e. closed iff it is an upper-only, pre-Conway interval.
+			!t.lowerBoundPresent && !t.strictUpperBound,
```
wolf31o2, line 564: Worth recording where the era decision actually gets made, because it is not in this repository.

`NewTxInfoV1FromTransaction`/`V2` are called from `ledger/conway/rules.go` and `ledger/dijkstra/rules.go` only, both passing `true`. Nothing in gouroboros calls them for Alonzo or Babbage, so the `false` branch this PR adds has no production caller here and is exercised only by the new tests.

The caller that does exercise both is dingo, which builds V1/V2 TxInfo from `ledger/eras/alonzo.go`, `babbage.go`, and `conway.go`. The added parameter breaks that build, which is the right failure mode — it forces an explicit per-era choice rather than silently inheriting one.

The risk is in how that choice gets made. Passing `true` uniformly to fix the compile error would repair Conway and simultaneously break Alonzo and Babbage, which are correct on `main` today and would start rendering an exclusive bound where the node uses `PV1.to`. Trading a Conway divergence for a pre-Conway one is easy to do by accident and hard to notice, since the pre-Conway path has no failing test here to catch it.

Worth calling out in the release note for this change, or filing the dingo-side update alongside it so the two land together.

wolf31o2 (approved):
Approving. I checked the era claim against cardano-ledger rather than the description, since a script-context change alters what every validator sees.

**The reference confirms both halves, including the lower-bound dependence.**

`Alonzo/Plutus/TxInfo.hs`:

```haskell
ValidityInterval SNothing (SJust i) -> PV1.to <$> transSlotToPOSIXTime i
ValidityInterval (SJust i) (SJust j) -> ... PV1.Interval (PV1.lowerBound t1) (PV1.strictUpperBound t2)
```

`Conway/TxInfo.hs` overrides it:

```haskell
ValidityInterval SNothing (SJust i) -> do
  t <- transSlotToPOSIXTime i
  pure $ PV1.Interval (PV1.LowerBound PV1.NegInf True) (PV1.strictUpperBound t)
```

So an upper-only interval is closed pre-Conway (`PV1.to`) and exclusive in Conway, while a two-sided interval is exclusive in both. `Babbage/TxInfo.hs` defines no override — it calls `Alonzo.transValidityInterval` directly — so "pre-Conway (Alonzo/Babbage)" is accurate. `!t.lowerBoundPresent && !t.strictUpperBound` is exactly that truth table.

**This is the root cause behind blinklabs-io/dingo#3627 / #3735.** `main`'s expression is `!t.lowerBoundPresent` — the Alonzo rule applied to every era — so a Conway TTL-only transaction gets a closed upper bound where the node uses an exclusive one. That is precisely the TTL-only Conway shape those issues report, and it explains a script-context difference rather than a cost-model one.

Both directions are pinned. Reverting to `!t.lowerBoundPresent` fails `TestTimeRangeToPlutusDataUpperBoundEraDependent/conway_ttl_only` and the `TestScriptContextV3` golden; forcing the bound always-exclusive fails `.../preconway_ttl_only` and `TestValidityRangeMatchesCardanoLedger/V1` and `/V2`. Neither probe fails the other's tests.

`go test ./ledger/...` green.

One consequence worth stating explicitly, inline.

PR #2173, opened 2026-09-02
fix(leiosfetch): answer unconfigured requests instead of holding agency

wolf31o2:
Keep block-range delivery alive through the terminal response.

Closes #3623

Tests: `go test -race ./protocol/leiosfetch -count=1`

### DIFF
--- a/protocol/leiosfetch/README.md
+++ b/protocol/leiosfetch/README.md
@@ -208,18 +208,24 @@ Errors from configured callbacks and transport failures are still propagated.
 
 ## Abandoned requests
 
-`BlockRequest` and `BlockTxsRequest` are bounded by the caller's context. A
-request whose context expires is abandoned: its delivery channel is cleared so
-a late response is dropped rather than mis-delivered, and the slot stays busy
-so the next request cannot be correlated with the outstanding response.
+`BlockRequest`, `BlockTxsRequest`, `VotesRequest`, and `BlockRangeRequest` are
+bounded by the caller's context. These requests share one connection-wide slot
+because their responses carry no request identifier. A request whose context
+expires is abandoned: its delivery channel is cleared so a late response is
+dropped rather than mis-delivered, and the slot stays busy so the next request
+cannot be correlated with the outstanding response. For `BlockRangeRequest`,
+this applies to the whole streaming exchange; the terminal response drains the
+abandoned slot.
 
 A later request waits a bounded grace period for that response to drain. If it
 arrives, the connection continues normally. If it does not, the exchange is
 desynchronised beyond recovery and the client fails the connection with
 `ErrRequestSlotAbandoned` so peer governance can drop and replace the peer.
 This is deliberately narrower than a protocol-level state timeout, which would
-also fire for a healthy relay that merely responded slowly, and is why
-`StateBlock` and `StateBlockTxs` still carry no `StateMap` timeout.
+also fire for a healthy relay that merely responded slowly. `StateBlock` and
+`StateBlockTxs` still carry no `StateMap` timeout; `StateVotes` and
+`StateBlockRange` retain their protocol-level timeout for exchanges that never
+return agency.
 
 ## Notes
 
--- a/protocol/leiosfetch/README.md
+++ b/protocol/leiosfetch/README.md
@@ -188,16 +188,38 @@ whole node-to-node connection.
 
 ## Optional server responders
 
-Leios fetch is optional, but not every request has a safe empty response. An
-unconfigured `VotesRequestFunc` returns `Votes` with an empty CBOR array,
-returning the protocol to `Idle` without a connection-level error. When a
-block, block-transactions, or block-range callback is unconfigured, the server
-enters the corresponding server-agency state and leaves the request pending.
-It sends no response because the available absence replies are either
-placeholder wire IDs or ambiguous with a real response. The requester cannot
-start another Leios fetch exchange, but other mini-protocols on the bearer
-remain usable. Errors from configured callbacks and transport failures are
-still propagated.
+Leios fetch is optional, but an unconfigured responder must never retain server
+agency. A server that accepts a request and never answers wedges the
+requester's client for the life of the connection: `protocol.sendLoop` waits
+for agency that only the missing response returns, so the requester can neither
+issue another Leios fetch request nor detect the condition.
+
+- An unconfigured `VotesRequestFunc` returns `Votes` with an empty CBOR array.
+- An unconfigured `BlockRequestFunc` returns `NoBlock`, and an unconfigured
+  `BlockTxsRequestFunc` returns `NoBlockTxs`. Both wire IDs are still
+  placeholders, but the configured not-available path already emits them, so
+  declining here adds no wire risk that the normal path does not already take.
+- An unconfigured `BlockRangeRequestFunc` returns a protocol error, because
+  `LastBlockAndTxsInRange` carries a mandatory block and there is no absence
+  reply to send. A connection-level error is diagnosable and lets the
+  requester's peer governance replace the peer; a silent hang is neither.
+
+Errors from configured callbacks and transport failures are still propagated.
+
+## Abandoned requests
+
+`BlockRequest` and `BlockTxsRequest` are bounded by the caller's context. A
+request whose context expires is abandoned: its delivery channel is cleared so
+a late response is dropped rather than mis-delivered, and the slot stays busy
+so the next request cannot be correlated with the outstanding response.
+
+A later request waits a bounded grace period for that response to drain. If it
+arrives, the connection continues normally. If it does not, the exchange is
+desynchronised beyond recovery and the client fails the connection with
+`ErrRequestSlotAbandoned` so peer governance can drop and replace the peer.
+This is deliberately narrower than a protocol-level state timeout, which would
+also fire for a healthy relay that merely responded slowly, and is why
+`StateBlock` and `StateBlockTxs` still carry no `StateMap` timeout.
 
 ## Notes
 
--- a/protocol/leiosfetch/client.go
+++ b/protocol/leiosfetch/client.go
@@ -39,7 +39,7 @@ type Client struct {
 }
 
 // requestSlot serializes a single outstanding request/response exchange for a
-// ping-pong leios-fetch state (Block or BlockTxs) and correlates the response
+// ping-pong or streaming leios-fetch state and correlates the response
 // with the caller that is actually waiting for it.
 //
 // These mini-protocol states carry no request identifier and the underlying
@@ -479,10 +479,10 @@ func (c *Client) BlockRangeRequest(
 				return ret, nil
 			}
 		case <-ctx.Done():
-			// release, not abandon, for the reason given in VotesRequest:
-			// a range response reaches blockRangeResultChan, never deliver,
-			// so an abandoned slot here would never drain.
-			c.blockRequestSlot.release(w)
+			// Range responses are delivered through the shared slot, including
+			// the terminal response. Keep the slot abandoned until that response
+			// drains so it cannot be delivered to a later request.
+			c.blockRequestSlot.abandon(w)
 			return nil, ctx.Err()
 		case <-c.DoneChan():
 			c.blockRequestSlot.release(w)
--- a/protocol/leiosfetch/client.go
+++ b/protocol/leiosfetch/client.go
@@ -33,11 +33,9 @@ type Client struct {
 	onceStop        sync.Once
 	// Block and BlockTxs share one slot because both requests use the same
 	// connection-wide agency and have no request identifier.
-	blockRequestSlot     requestSlot
-	blockSlot            requestSlot
-	blockTxsSlot         requestSlot
-	votesResultChan      chan protocol.Message
-	blockRangeResultChan chan protocol.Message
+	blockRequestSlot requestSlot
+	blockSlot        requestSlot
+	blockTxsSlot     requestSlot
 }
 
 // requestSlot serializes a single outstanding request/response exchange for a
@@ -220,8 +218,6 @@ func NewClient(protoOptions protocol.ProtocolOptions, cfg *Config) *Client {
 		// the response always has somewhere to land even if the caller has
 		// not reached its receive yet. Unbuffered, a response whose caller
 		// gave up on its context blocked the protocol receive loop forever.
-		votesResultChan:      make(chan protocol.Message, 1),
-		blockRangeResultChan: make(chan protocol.Message, 1),
 	}
 	c.callbackContext = CallbackContext{
 		Client:       c,
@@ -277,15 +273,6 @@ func (c *Client) Start() {
 			)
 		c.Protocol.Start()
 		// Start goroutine to cleanup resources on protocol shutdown.
-		// The Block/BlockTxs request slots are unblocked directly by the
-		// per-request select watching DoneChan (see BlockRequest /
-		// BlockTxsRequest), so only the shared Votes/BlockRange channels are
-		// closed here.
-		go func() {
-			<-c.DoneChan()
-			close(c.votesResultChan)
-			close(c.blockRangeResultChan)
-		}()
 	})
 }
 
@@ -444,8 +431,7 @@ func (c *Client) VotesRequest(
 		return nil, err
 	}
 	select {
-	case resp, ok := <-c.votesResultChan:
-		c.blockRequestSlot.release(w)
+	case resp, ok := <-w:
 		if !ok {
 			return nil, protocol.ErrProtocolShuttingDown
 		}
@@ -482,7 +468,7 @@ func (c *Client) BlockRangeRequest(
 	ret := make([]protocol.Message, 0, 20)
 	for {
 		select {
-		case resp, ok := <-c.blockRangeResultChan:
+		case resp, ok := <-w:
 			if !ok {
 				c.blockRequestSlot.release(w)
 				return nil, protocol.ErrProtocolShuttingDown
@@ -590,25 +576,19 @@ func (c *Client) handleNoBlockTxs(msg protocol.Message) {
 // waiting. The send must not block: the protocol receive loop runs it, and a
 // caller whose context expired is gone for good.
 func (c *Client) handleVotes(msg protocol.Message) {
-	select {
-	case c.votesResultChan <- msg:
-	default:
+	if !c.blockRequestSlot.deliver(msg) {
 		c.logDroppedResponse(msg)
 	}
 }
 
 func (c *Client) handleNextBlockAndTxsInRange(msg protocol.Message) {
-	select {
-	case c.blockRangeResultChan <- msg:
-	default:
+	if !c.blockRequestSlot.deliver(msg) {
 		c.logDroppedResponse(msg)
 	}
 }
 
 func (c *Client) handleLastBlockAndTxsInRange(msg protocol.Message) {
-	select {
-	case c.blockRangeResultChan <- msg:
-	default:
+	if !c.blockRequestSlot.deliver(msg) {
 		c.logDroppedResponse(msg)
 	}
 }
--- a/protocol/leiosfetch/client.go
+++ b/protocol/leiosfetch/client.go
@@ -16,6 +16,7 @@ package leiosfetch
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"time"
@@ -119,8 +120,9 @@ func (s *requestSlot) acquire(
 // abandon clears the live waiter so that a late response is dropped by deliver
 // rather than mis-delivered, while leaving the slot busy until that response
 // arrives (or the protocol shuts down). Subsequent acquire calls have a bounded
-// grace period while the slot is abandoned instead of parking indefinitely. It
-// is used when a caller's context expires before a response is received.
+// grace period while the slot is abandoned instead of parking indefinitely, and
+// on its expiry the client fails the connection (see acquireSlot). It is used
+// when a caller's context expires before a response is received.
 //
 // The caller passes the delivery channel it registered in acquire. The slot is
 // only mutated while that channel is still the live waiter (s.waiter == w). If
@@ -212,12 +214,17 @@ func NewClient(protoOptions protocol.ProtocolOptions, cfg *Config) *Client {
 	// Update state map with timeout
 	stateMap := StateMap.Copy()
 	// NOTE: StateBlock and StateBlockTxs intentionally do NOT get a
-	// protocol-level timeout. A missing response to a BlockRequest /
+	// protocol-level timeout. A slow response to a BlockRequest /
 	// BlockTxsRequest must fail only that individual request (bounded by the
 	// caller-supplied context), never tear down the shared multiplexed
 	// connection. The protocol-level timeout fires p.SendError(), which is
 	// fatal to every mini-protocol on the same bearer (chainsync, blockfetch,
-	// etc.), so it must not be wired for these two states.
+	// etc.), so it must not be wired for these two states: it would fire for
+	// a healthy relay that merely responded later than the timeout.
+	//
+	// A response that never arrives at all is a different case and IS fatal,
+	// but only once acquireSlot has proven the exchange is desynchronised --
+	// see acquireSlot.
 	if entry, ok := stateMap[StateVotes]; ok {
 		entry.Timeout = c.config.Timeout
 		stateMap[StateVotes] = entry
@@ -281,6 +288,44 @@ func (c *Client) Stop() error {
 	return err
 }
 
+// acquireSlot acquires slot for a new request in the given server-agency
+// state.
+//
+// A slot that is still holding an abandoned request's response means the peer
+// took leios-fetch agency and never returned it. That is not recoverable on
+// this connection: protocol.sendLoop waits on sendReadyChan for agency the
+// state map only grants once the missing response arrives, so no later
+// leios-fetch request can ever be written to this bearer. Keeping the bearer
+// alive therefore keeps a permanently dead leios-fetch client attached to a
+// peer that still looks healthy, which is exactly how dingo issue #3623
+// stalled a node's ledger indefinitely.
+//
+// Fail the connection instead, so the consumer's peer governance drops and
+// replaces the peer. This does not reintroduce the mis-delivery hazard the
+// slot exists to prevent: the slot is never reused, the connection is
+// discarded.
+func (c *Client) acquireSlot(
+	ctx context.Context,
+	slot *requestSlot,
+	state protocol.State,
+) (chan protocol.Message, error) {
+	w, err := slot.acquire(ctx, c.DoneChan())
+	if err == nil {
+		return w, nil
+	}
+	if errors.Is(err, ErrRequestSlotAbandoned) {
+		c.SendError(
+			fmt.Errorf(
+				"%s: peer retained agency in state %s after an abandoned request: %w",
+				ProtocolName,
+				state,
+				err,
+			),
+		)
+	}
+	return nil, err
+}
+
 // BlockRequest fetches the requested EB identified by the specified point.
 //
 // The wait for a response is bounded by the provided context. If the context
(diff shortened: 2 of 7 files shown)


protocol/leiosfetch/client.go
```diff
@@ -26,10 +27,13 @@ import (
...
-	callbackContext      CallbackContext
-	onceStart            sync.Once
-	onceStop             sync.Once
+	config          *Config
+	callbackContext CallbackContext
+	onceStart       sync.Once
+	onceStop        sync.Once
+	// Block and BlockTxs share one slot because both requests use the same
+	// connection-wide agency and have no request identifier.
+	blockRequestSlot     requestSlot
```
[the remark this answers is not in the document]
wolf31o2: Confirmed and fixed in e147ef6. Both methods sent directly with no slot, took no context, and read unbuffered result channels, so after an abandoned block request they blocked forever with no diagnosis — `acquireSlot` was never reached.

Both now take a context and acquire the shared admission slot. Their responses do not arrive through the slot, so success releases it explicitly; a context expiring with the response outstanding abandons instead, since the peer still holds agency and a late reply must still drain it.

Neither method had a production caller or a test. Added two driven through the real client path — `TestVotesRequestAfterAbandonedBlockRequestFailsFast` and `TestBlockRangeRequestAfterAbandonedBlockRequestFailsFast` — which park and fail without the change (reverted build compiled first).

protocol/leiosfetch/client.go
```diff
@@ -346,51 +405,97 @@ func (c *Client) BlockTxsRequest(
...
 	msg := NewMsgVotesRequest(voteIds)
-	if err := c.SendMessage(msg); err != nil {
+	if err := c.SendMessageContext(ctx, msg); err != nil {
+		c.blockRequestSlot.release(w)
 		return nil, err
 	}
-	resp, ok := <-c.votesResultChan
-	if !ok {
+	select {
+	case resp, ok := <-c.votesResultChan:
```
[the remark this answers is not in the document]
wolf31o2: Confirmed — this was introduced by my own previous commit. `abandon` keeps the slot busy until a late response drains it through `deliver`, and a votes or range response goes to its own result channel and never reaches `deliver`, so the slot stayed permanently busy and failed every later request on the connection.

Fixed in 2463098: those two states `release` on context expiry instead. Their responses are not slot-correlated, so the mis-delivery `abandon` guards cannot arise, and a late one is dropped by the handler.

The same review turned up a second half: both result channels were unbuffered with blocking sends, so a response whose caller had given up blocked the protocol receive loop. They are now capacity 1 — the caller registers before its request is sent — and the handlers drop a response nobody is waiting for rather than parking the loop.

protocol/leiosfetch/leiosfetch_test.go
```diff
@@ -386,3 +543,148 @@ func TestBlockRequestSubsequentAfterAbandoned(t *testing.T) {
...
+
+// TestVotesRequestAfterAbandonedBlockRequestFailsFast pins that a votes
+// request goes through the same admission slot as a block request.
+//
+// The leios-fetch states share one connection-wide agency, so a block request
+// whose caller gave up leaves the peer holding agency and nothing further can
+// be written to the bearer. Before VotesRequest acquired the slot it skipped
+// that diagnosis entirely and blocked forever on votesResultChan with no
+// context and no recovery.
+func TestVotesRequestAfterAbandonedBlockRequestFailsFast(t *testing.T) {
```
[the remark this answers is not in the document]
wolf31o2: Agreed, fixed in 2463098 — the two tests are now one table-driven test parameterised by the request function, cutting roughly 135 duplicated lines.

protocol/leiosfetch/client.go
```diff
@@ -26,10 +27,13 @@ import (
...
-	onceStop             sync.Once
+	config          *Config
+	callbackContext CallbackContext
+	onceStart       sync.Once
+	onceStop        sync.Once
+	// Block and BlockTxs share one slot because both requests use the same
+	// connection-wide agency and have no request identifier.
+	blockRequestSlot     requestSlot
 	blockSlot            requestSlot
 	blockTxsSlot         requestSlot
```
cryptodj413: `blockSlot`/`blockTxsSlot` look dead as of this PR: `BlockRequest`/`BlockTxsRequest` now acquire `blockRequestSlot` exclusively (lines 360, 395), so nothing ever calls `.acquire()` on these two anymore. They only survive as always-false second `deliver()` attempts in `handleBlock`/`handleNoBlock`/`handleBlockTxs`/`handleNoBlockTxs` below (e.g. lines 552, 565, 571, 584). Worth deleting both the fields and the fallback calls in a follow-up, since `client_test.go`'s `TestClientMessageHandler` currently exercises message routing through these dead fields rather than the real `blockRequestSlot` path.

protocol/leiosfetch/client.go
```diff
@@ -346,51 +409,100 @@ func (c *Client) BlockTxsRequest(
...
-	if !ok {
+	select {
+	case resp, ok := <-c.votesResultChan:
+		c.blockRequestSlot.release(w)
+		if !ok {
+			return nil, protocol.ErrProtocolShuttingDown
+		}
+		return resp, nil
+	case <-ctx.Done():
+		c.blockRequestSlot.abandon(w)
```
cryptodj413, line 522: `abandon()` leaves `blockRequestSlot` busy+abandoned until something calls `deliver()` or `release()` on it. But a late votes reply only reaches `handleVotes` (line 592), which pushes into `votesResultChan` and never touches `blockRequestSlot`. So this abandonment is structurally undrainable: the *next* leios-fetch request on this connection (of any of the four kinds, since they all share this slot) will always burn the full `abandonedRequestWait` grace period and then fail the whole connection with `ErrRequestSlotAbandoned` — even when the peer answers correctly and promptly after this caller's context expired.

Reproduced: peer sends `MsgVotes` 200ms after a 50ms client-side timeout (state machine confirms the peer returned agency to `Idle`); the next `BlockRequest` still fails with `leios-fetch: peer retained agency in state Block after an abandoned request: ...`, tearing down the connection for a peer that behaved correctly. This is the same failure class dingo#3623 describes (silent permanent wedge), now reachable through a votes timeout rather than an unconfigured responder, plus it kills the connection unnecessarily.

Note `StateVotes` already carries a `StateMap` timeout (line 244-247), so a genuinely non-responsive peer is already caught there — this `abandon` path only ever fires for peers that *do* answer.

protocol/leiosfetch/client.go
```diff
@@ -281,6 +304,44 @@ func (c *Client) Stop() error {
...
+	state protocol.State,
+) (chan protocol.Message, error) {
+	w, err := slot.acquire(ctx, c.DoneChan())
+	if err == nil {
+		return w, nil
+	}
+	if errors.Is(err, ErrRequestSlotAbandoned) {
+		c.SendError(
+			fmt.Errorf(
+				"%s: peer retained agency in state %s after an abandoned request: %w",
```
cryptodj413, line 404: Minor: this formats the error with the *new* request's target `state` argument, not the state where agency was actually retained. In my Finding-A repro (see general review comment) this produced `"peer retained agency in state Block after an abandoned request"` for a wedge that actually originated in `StateVotes` on a connection where the peer had already returned agency — misleading in an operator's logs.

protocol/leiosfetch/README.md
```diff
@@ -188,16 +188,38 @@ whole node-to-node connection.
...
+  reply to send. A connection-level error is diagnosable and lets the
+  requester's peer governance replace the peer; a silent hang is neither.
+
+Errors from configured callbacks and transport failures are still propagated.
+
+## Abandoned requests
+
+`BlockRequest` and `BlockTxsRequest` are bounded by the caller's context. A
+request whose context expires is abandoned: its delivery channel is cleared so
+a late response is dropped rather than mis-delivered, and the slot stays busy
```
cryptodj413: This section was written for the first commit in the PR and not updated for the later two (`e147ef6`, `2463098`): it only describes Block/BlockTxs's abandon-then-grace-period policy. It doesn't mention that `VotesRequest`/`BlockRangeRequest` now acquire the *same* shared slot (so an abandoned block request fails them too), or the release-vs-abandon split between them, which is the source of the mis-delivery in the other inline comment on this PR.

cryptodj413 (changes requested):
Reviewed the full diff against the PR head, including tracing `protocol.Protocol`'s send/receive/state-machine code and writing executable reproductions (run with `-race`) rather than taking either the bots' or my own first read at face value.

**Two confirmed, reproduced merge blockers**, both rooted in the same design gap: `blockRequestSlot` now gates *admission* for all four request kinds (Block/BlockTxs/Votes/Range), but only Block/BlockTxs route their *response* back through it via `deliver()` (which always drains the slot). Votes and Range responses bypass it entirely (`handleVotes`, `handleNextBlockAndTxsInRange`, `handleLastBlockAndTxsInRange` only push into their own shared channels), so the slot's admission bookkeeping and its response delivery are out of sync for those two:

1. **`VotesRequest`'s `ctx.Done()` path (`abandon`) permanently wedges the shared slot, then kills a healthy connection** — see inline comment. Reproduced: peer answers `MsgVotes` 200ms after a 50ms client timeout (a *healthy* peer); the next leios-fetch request on the connection still burns the full `abandonedRequestWait` and then fails the whole multiplexed bearer with `ErrRequestSlotAbandoned`, because nothing ever drains an abandoned votes slot.
2. **`BlockRangeRequest`'s `ctx.Done()` path (`release`) causes confirmed stale-data mis-delivery to the next caller** — this is the still-open Cubic P1 at client.go:499; replied to that thread with a concrete repro confirming it. A second, unrelated `BlockRangeRequest` received the *first* (abandoned) request's stale terminal message as if it were its own, with `err == nil`.

Both are worse than the bug this PR fixes: a silent wrong answer (2) or a silent permanent wedge with the connection then torn down for a peer that behaved correctly (1), versus the original detectable hang.

**Also found:**
- No test exercises a `VotesRequest`/`BlockRangeRequest` whose own context expires with a response still outstanding, nor a successful end-to-end call to either — the new `release`-on-success paths and the buffered/non-blocking channel change are entirely uncovered by this PR's own tests.
- `blockSlot`/`blockTxsSlot` (client.go:37-38) are dead now that Block/BlockTxs moved onto `blockRequestSlot`; they only survive as always-false fallback targets in the four `handleBlock*`/`handleNoBlock*` handlers. Not part of this diff, but worth a heads-up: `client_test.go`'s `TestClientMessageHandler` currently drives routing through these dead fields, so it passes even with `blockRequestSlot.deliver()` deleted from all four handlers — it isn't covering the real path anymore.
- README's new "Abandoned requests" section (209-222) was written for the first commit and never updated for the later two: it doesn't mention that Votes/Range now share the same admission slot, or the release-vs-abandon split that's the source of finding 2.

No API/wire compatibility break: nothing in the blinklabs-io workspace calls `VotesRequest`/`BlockRangeRequest` today (only Block/BlockTxs, unchanged signatures), so the new `ctx` parameter is safe as an isolated change; dingo's own follow-up (dingo#3772) already flags that consuming this needs a go.mod bump plus a client-side test update.

Happy to help work through a fix for 1/2 — the cleanest direction looks like draining `blockRequestSlot` from the votes/range handlers too (so `abandon` becomes legitimately recoverable everywhere and both context paths can use it consistently), or correlating each votes/range response per-request the way Block/BlockTxs already do via `deliver()`, rather than through one shared, unscoped channel.

protocol/leiosfetch/client.go
```diff
@@ -346,51 +409,100 @@ func (c *Client) BlockTxsRequest(
...
+			ret = append(ret, resp)
+			if _, ok := resp.(*MsgLastBlockAndTxsInRange); ok {
+				c.blockRequestSlot.release(w)
+				return ret, nil
+			}
+		case <-ctx.Done():
+			// release, not abandon, for the reason given in VotesRequest:
+			// a range response reaches blockRangeResultChan, never deliver,
+			// so an abandoned slot here would never drain.
+			c.blockRequestSlot.release(w)
```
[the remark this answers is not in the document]
cryptodj413: Confirmed with a reproduction — this is a real, live bug, not a false positive.

Repro: caller A sends `BlockRangeRequest` with a 100ms context; the peer sleeps 300ms then sends the terminal `MsgLastBlockAndTxsInRange` for A's (already-abandoned) exchange. Immediately after A's timeout, caller B sends a second, unrelated `BlockRangeRequest` (fresh 2s context, different range) on the same client. Result:

```
req1 err: context deadline exceeded
req2 err: <nil>   blockRaw: 82 01 02   (A's stale terminal message, not B's own 82 03 04)
```

Root cause: `release()` frees `blockRequestSlot` immediately, so B can acquire it and start reading `blockRangeResultChan` before A's late response has drained. Since that channel is shared and unscoped (not per-request, unlike Block/BlockTxs's `deliver()`-based correlation), B silently receives A's data with `err == nil` — no error, no way to detect it. Filed as an inline comment and in the top-level review; there's a closely related second bug on `VotesRequest`'s side (uses `abandon` instead, which is undrainable there and permanently wedges the slot).

protocol/leiosfetch/client.go
```diff
@@ -346,51 +396,99 @@ func (c *Client) BlockTxsRequest(
...
```
cryptodj413: This still `release()`s rather than `abandon()`s, and the justifying comment above ("never deliver") is no longer accurate after this commit's refactor: `handleNextBlockAndTxsInRange`/`handleLastBlockAndTxsInRange` now call `c.blockRequestSlot.deliver(msg)` (lines 585, 591), so a late range response *does* reach `deliver()` just like Block/BlockTxs.

Because this still releases immediately, a second `BlockRangeRequest` can acquire the slot and register as the new `s.waiter` before this (abandoned) request's late terminal message arrives — and `deliver()` will hand that stale message to the *second* request. Reproduced against this exact commit: a second, unrelated `BlockRangeRequest` (fresh context, different range) received the first (abandoned) request's stale `MsgLastBlockAndTxsInRange` payload with `err == nil`.

`VotesRequest`'s equivalent branch (line 440) already calls `abandon(w)` and — per the top-level review comment — that now correctly recovers when the peer answers within the grace period. This branch looks like it should match: `c.blockRequestSlot.abandon(w)` instead of `release(w)`.

cryptodj413 (dismissed):
Re-reviewed `3654866` ("correlate votes and range responses") against the two merge blockers from the previous review, with fresh reproductions run against this exact commit.

**Finding A (VotesRequest permanent wedge) — fixed, verified.** Routing votes responses through `blockRequestSlot.deliver()` makes `abandon()` properly drainable now. Repro: a `VotesRequest` whose 50ms context expires, followed by the peer answering correctly 200ms later (well within the peer's real behavior), no longer wedges the slot or kills the connection — a subsequent `BlockRequest` on the same client succeeds normally and no connection error is emitted. Thanks for the fix.

**Finding B (BlockRangeRequest stale mis-delivery) — still present.** The correlation mechanism changed (per-request channel via `deliver()` instead of a shared `blockRangeResultChan`), but `BlockRangeRequest`'s `ctx.Done()` branch still calls `release()`, not `abandon()` (see inline comment), so the same mis-delivery survives through the new mechanism: `release()` frees the slot immediately, letting a second, unrelated `BlockRangeRequest` become the slot's registered waiter before the first (abandoned) request's late terminal message arrives; `deliver()` then hands that stale message to the *second* request's channel. Reproduced against this exact commit:

```
req1 (100ms ctx, range [1,2]) err: context deadline exceeded
req2 (2s ctx, range [3,4])    err: <nil>   blockRaw: 82 01 02   (req1's stale data, not req2's own 82 03 04)
```

The comment justifying `release()` ("a range response reaches blockRangeResultChan, never deliver, so an abandoned slot here would never drain") is now factually incorrect post-refactor — range responses *do* reach `deliver()` now (via `handleNextBlockAndTxsInRange`/`handleLastBlockAndTxsInRange`), so `abandon()` should be just as drainable for `BlockRangeRequest` as it now is for `VotesRequest`. Switching this one branch from `release(w)` to `abandon(w)` looks like it would close this out, matching `VotesRequest`'s pattern exactly.

Still open from the previous round, unaddressed by this commit: the dead `blockSlot`/`blockTxsSlot` fields (`client_test.go`'s `TestClientMessageHandler` still acquires those instead of `blockRequestSlot` for the Block/BlockTxs cases, so it still isn't exercising the real production delivery path for those two message types), and the README "Abandoned requests" section (still describes only the pre-refactor Block/BlockTxs-only policy).

protocol/leiosfetch/client.go
cryptodj413: Verified fixed as of 3654866. Repro: VotesRequest with a 50ms context, peer answers correctly 200ms later (well within the grace period) — the abandoned slot now drains via `deliver()`, and a subsequent BlockRequest on the same client succeeds normally with no connection error. Thanks.

protocol/leiosfetch/client.go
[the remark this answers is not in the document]
cryptodj413: Still present post-3654866, now via the new per-request channel rather than the old shared one — see the fresh top-level review and the inline comment at the current line 485 for a repro against this exact commit. Root cause is unchanged: this branch still `release()`s instead of `abandon()`s, and the comment's premise for doing so is no longer true now that range responses reach `deliver()`.

wolf31o2:
Resolved the latest range-response review finding in 83e7b7e. Abandoned BlockRangeRequest calls now quarantine the shared request slot until the late terminal response drains, preventing stale data from reaching a later request. Added a delayed-response regression test and updated the abandoned-request documentation. Validation: focused race test, protocol-wide race tests, conformance tests, and make test passed; integration endpoint test skipped because no Cardano node is configured. Repository lint reported no source issues but the example phase was blocked by a concurrent golangci-lint lock.

protocol/leiosfetch/client.go
```diff
@@ -469,19 +565,28 @@ func (c *Client) handleNoBlockTxs(msg protocol.Message) {
...
 func (c *Client) handleVotes(msg protocol.Message) {
-	c.votesResultChan <- msg
+	if !c.blockRequestSlot.deliver(msg) {
+		c.logDroppedResponse(msg)
+	}
 }
 
 func (c *Client) handleNextBlockAndTxsInRange(msg protocol.Message) {
-	c.blockRangeResultChan <- msg
+	if !c.blockRequestSlot.deliver(msg) {
```
chrisguiney: `deliver` frees the slot on every message, but this state streams.

`requestSlot.deliver` (`client.go:178`) does `s.waiter = nil; s.freeLocked()` unconditionally. `StateMap` self-transitions `StateBlockRange` on `MessageTypeNextBlockAndTxsInRange` (`leiosfetch.go:96`), so a range legitimately delivers N messages through the one registration `BlockRangeRequest` made. After the first `Next`, the waiter is gone: message 2 lands on `w == nil`, returns false, and gets logged as a dropped response while the caller is still selecting on `w`.

The premise for this is stated at `client.go:187` — "receives at most one response per request (the state machine rejects a second server message before it reaches this handler)". That holds for Block, BlockTxs and Votes, but `StateBlockRange` is the one state where the state machine explicitly permits the second message.

**Reproduced, two ways.**

A range of Next/Next/Last never assembles — `BlockRangeRequest` returns `context deadline exceeded`. This test passes on `origin/main` and fails here:

```go
ouroboros_mock.ConversationEntryInput{
    ProtocolId:  leiosfetch.ProtocolId,
    MessageType: leiosfetch.MessageTypeBlockRangeRequest,
},
ouroboros_mock.ConversationEntryOutput{
    ProtocolId: leiosfetch.ProtocolId,
    IsResponse: true,
    Messages: []protocol.Message{
        leiosfetch.NewMsgNextBlockAndTxsInRange([]byte{0x82, 0x01, 0x01}, nil),
        leiosfetch.NewMsgNextBlockAndTxsInRange([]byte{0x82, 0x01, 0x02}, nil),
        leiosfetch.NewMsgLastBlockAndTxsInRange([]byte{0x82, 0x01, 0x03}, nil),
    },
},
// ... resp, err := client.BlockRangeRequest(ctx, start, end)
// require.NoError(t, err); require.Len(t, resp, 3)
```

Bisected with it: `e147ef6` ok, `2463098` ok, **`3654866` first failure**, still failing at head.

Worse, the freed slot is immediately reusable while the peer is still streaming. With a `BlockRequest` issued 250ms into a range, I got:

```
BlockRequest returned *leiosfetch.MsgNextBlockAndTxsInRange
```

That is the mis-delivery the slot was built to prevent, now reachable on the range path.

One thing worth deciding before the fix: whichever way the waiter is retained across a stream, `deliver` runs on the protocol receive loop, so the send has to stay non-blocking. A retained waiter on the current capacity-1 channel would let a server that streams faster than the caller drains block `recvLoop`. Happy to think through options with you if it helps — a per-request queue and a slot that only frees on the terminal message is the shape I would reach for, but you know this protocol better than I do.

protocol/leiosfetch/client.go
```diff
@@ -346,51 +394,99 @@ func (c *Client) BlockTxsRequest(
...
+				c.blockRequestSlot.release(w)
+				return nil, protocol.ErrProtocolShuttingDown
+			}
+			ret = append(ret, resp)
+			if _, ok := resp.(*MsgLastBlockAndTxsInRange); ok {
+				c.blockRequestSlot.release(w)
+				return ret, nil
+			}
+		case <-ctx.Done():
+			// Range responses are delivered through the shared slot, including
```
chrisguiney, line 564: Follows from the `deliver` finding above: this `abandon(w)` is a no-op for any range that already received a message.

`abandon` only mutates while `s.waiter == w` (`client.go:129`), and `deliver` cleared `s.waiter` on the first `Next`. So the quarantine `83e7b7e` added only applies to a range abandoned *before* its first message — which is the narrower half of what the earlier review thread asked for. Past that point the slot is already free, a later request can acquire it, and the peer's remaining range messages go to that request instead.

The comment here states the intended behaviour exactly right ("Keep the slot abandoned until that response drains"); it is `deliver` that does not implement it. I did try to get a clean end-to-end repro of the stale-terminal-message case specifically, but the connection tore down first in my harness, so I am flagging this as reasoning from the code plus the two repros above rather than something I measured.

protocol/leiosfetch/leiosfetch_test.go
```diff
@@ -386,3 +543,179 @@ func TestBlockRequestSubsequentAfterAbandoned(t *testing.T) {
 		},
 	)
 }
+
+// TestBlockRangeRequestSubsequentAfterAbandoned verifies that a late terminal
+// response from an abandoned range exchange cannot be delivered to a later
+// range request. Range responses use the shared request slot, so the first
+// request must remain quarantined until its terminal response drains.
+func TestBlockRangeRequestSubsequentAfterAbandoned(t *testing.T) {
```
chrisguiney, line 551: No test sends more than one message in a range.

This test and every other range case respond with a single `MsgLastBlockAndTxsInRange`, so the streaming path — the one thing `StateBlockRange` exists for — is never exercised. That is why the full suite, `-race -count=2`, lint and all of CI are green on a head where a two-message range does not work.

A Next/Next/Last case here would have caught `3654866` on the commit that introduced it. Snippet in the `handleNextBlockAndTxsInRange` comment if it saves you a few minutes.

protocol/leiosfetch/client.go
```diff
@@ -201,9 +211,11 @@ func NewClient(protoOptions protocol.ProtocolOptions, cfg *Config) *Client {
 		cfg = &tmpCfg
 	}
 	c := &Client{
-		config:               cfg,
-		votesResultChan:      make(chan protocol.Message),
-		blockRangeResultChan: make(chan protocol.Message),
+		config: cfg,
+		// Capacity 1: the caller registers before its request is sent, so
```
chrisguiney: Two comments got orphaned by the refactor and now describe code that is not here.

- This one describes a channel capacity, but `votesResultChan` and `blockRangeResultChan` are gone and the literal now sets only `config`. The capacity-1 rationale still applies, just to the channel `requestSlot.acquire` creates — it would read correctly there.
- `client.go:273`: `// Start goroutine to cleanup resources on protocol shutdown.` is now the last line of `Start()`, with no goroutine after it.

Trivial, but both are the kind of thing that misleads the next reader of this file.

protocol/leiosfetch/README.md
```diff
@@ -188,16 +188,44 @@ whole node-to-node connection.
...
+
+## Abandoned requests
+
+`BlockRequest`, `BlockTxsRequest`, `VotesRequest`, and `BlockRangeRequest` are
+bounded by the caller's context. These requests share one connection-wide slot
+because their responses carry no request identifier. A request whose context
+expires is abandoned: its delivery channel is cleared so a late response is
+dropped rather than mis-delivered, and the slot stays busy so the next request
+cannot be correlated with the outstanding response. For `BlockRangeRequest`,
+this applies to the whole streaming exchange; the terminal response drains the
```
chrisguiney, line 217: This documents behaviour the code does not currently have: `deliver` frees the slot on the *first* range message, not the terminal one (see the inline comment on `handleNextBlockAndTxsInRange`).

Worth re-checking this paragraph once the streaming fix lands — as written it is the right specification, so it may need no edit at all beyond confirming the code matches it.

chrisguiney (changes requested):
Thanks for the depth here — the server-side half is right, and I checked it against the state machine rather than taking the description on trust: `MsgLastBlockAndTxsInRange` genuinely carries a mandatory block, so a protocol error really is the only honest answer for an unconfigured range responder.

The client-side refactor that grew on top of it breaks the streaming range state. Details inline; the main one is at `handleNextBlockAndTxsInRange`.

## The short version

`requestSlot.deliver` (`client.go:178`) clears the waiter and frees the slot on **every** message, but `StateBlockRange` self-transitions on `MessageTypeNextBlockAndTxsInRange` (`leiosfetch.go:96`), so one range registration has to survive N messages. Two things follow, both reproduced:

1. A Next/Next/Last range makes `BlockRangeRequest` return `context deadline exceeded`. The same test passes on `origin/main`.
2. A `BlockRequest` issued while a range is streaming returns a `*MsgNextBlockAndTxsInRange` — the mis-delivery the slot exists to prevent.

Bisected across the branch with test 1: `e147ef6` ok, `2463098` ok, **`3654866` first failure**, still failing at `b136417`.

## Verified good

- Merging `blockSlot` and `blockTxsSlot` is right — those states share one connection-wide agency, so two slots were never buying anything.
- Dropping the `Start()` cleanup goroutine removes a goroutine rather than leaking one, and the tests keep `goleak`.
- The protocol-level timeout re-arms on the `StateBlockRange` self-transition (`setState` runs per transition), so a long range will not trip `config.Timeout`. The README's claim there holds.
- `SendError` on `ErrRequestSlotAbandoned` is a big hammer, but I went looking for a gentler option and could not find one: `sendLoop` only writes at `AgencyClient`, so a bearer whose peer never returns agency really is dead for leios-fetch.

## Checks run locally

`go build ./...`, `go vet ./protocol/...`, `gofmt`, `golangci-lint run ./protocol/leiosfetch/...` (0 issues), `go test ./...` and `go test -race -count=2 ./protocol/leiosfetch/` all pass — which is the problem: nothing in the suite sends more than one message in a range. Not run: conformance, devnet, nilaway locally (CI green).

## Smaller things, not inline

- `VotesRequest` and `BlockRangeRequest` gained a `ctx` parameter. That is a public API break; no in-org client consumer today (dingo wires only the server callbacks), so I am not asking you to avoid it — but it is worth a line in the description.
- `BlockRequest`, `BlockTxsRequest` and `VotesRequest` are now three near-identical bodies differing only in the message constructor and the not-found mapping. A single `doRequest(ctx, state, msg)` helper would make the shared-slot policy live in one place, which is also where a future rule would have to go. Nit, and fine as a follow-up.
- The PR description still describes only the first commit. The single-slot refactor, the `ctx` break and the votes/range routing are the bulk of the diff and are not mentioned.

## Process note

Cubic's last review was 09-02 20:13Z, so `3654866`, `83e7b7e` and `b136417` are unreviewed by it, and CodeRabbit's green check reads "Review skipped: incremental reviews are disabled". The regression landed in `3654866`, inside that gap — worth another bot pass before this goes back out for human review.

protocol/leiosfetch/client.go
chrisguiney: Confirming this one is still live at `b136417`, and I think the shared slot has made it worse rather than neutral: now that all four request kinds admit through `blockRequestSlot`, the `state` argument is the state the *new* caller wanted, and the abandoned request that actually wedged the exchange could have been any of the other three. An operator reading `state Block` has no way to tell which.

Cheap fix if you want one: have `acquire` return the state recorded when the slot was taken, or store it on `requestSlot` at acquire time, and format that instead. Not blocking from me.

protocol/leiosfetch/client.go
chrisguiney: Following up on this rather than opening a duplicate — the `release` -> `abandon` change did land in `83e7b7e`, but I do not think it takes effect for the case you described.

`abandon` only mutates while `s.waiter == w` (`client.go:129`), and `deliver` clears `s.waiter` on the *first* range message, because it frees the slot on every message rather than only on the terminal one. So for any range that already received a `Next`, the slot is free before the context expires and `abandon` is a no-op — the quarantine only covers a range abandoned before its first message.

I filed that root cause separately on `handleNextBlockAndTxsInRange`, with two repros: a Next/Next/Last range never assembles at all (passes on `origin/main`, bisects to `3654866`), and a `BlockRequest` issued mid-range comes back holding a `*MsgNextBlockAndTxsInRange`. Your read of this branch was right; it just needs the fix one level down.

chrisguiney (approved):

PR #2190, opened 2026-09-03
fix(script): encode validity upper bounds as exclusive

### LINKED ISSUE #2160: ScriptContext validity-interval upper bound is encoded as INCLUSIVE for TTL-only txs (should be EXCLUSIVE) — mis-computes Plutus ExUnits
## Summary

`TimeRange.ToPlutusData()` (`ledger/common/script/context.go`) encodes the Plutus
validity-interval **upper bound** closure as `!t.lowerBoundPresent`. For a
TTL-only transaction (`invalidHereafter` set, `invalidBefore` absent, so
`lowerBoundPresent == false`) this yields `closed = true`, producing
`UpperBound (Finite ttl) True` (INCLUSIVE).

cardano-ledger always translates the validity-interval upper bound with
`strictUpperBound` (EXCLUSIVE): `UpperBound (Finite ttl) False`. The in-code
comment ("cardano-ledger uses `to` (closed upper) for an upper-only interval")
is factually incorrect.

Because the ScriptContext handed to the script interpreter differs from the one
the real ledger builds, **scripts that inspect `txInfoValidRange` compute the
wrong result and the reported ExUnits diverge from the node.** This is a
consensus-relevant divergence for any tooling that evaluates scripts via
gouroboros (fee/ExUnits estimation, phase-2 validation, mempool simulation).

## Where

`ledger/common/script/context.go`, `TimeRange.ToPlutusData()`:

```go
bound(
    t.upperBound,
    t.upperBoundPresent,
    false,
    // cardano-ledger uses `to` (closed upper) for an
    // upper-only interval, but `strictUpperBound` when
    // both bounds are present.
    !t.lowerBoundPresent,   // <-- BUG: True (inclusive) when no lower bound
),
```

When both bounds are present `!lowerBoundPresent` is already `false`, so the
bug is invisible for two-sided intervals and only bites TTL-only transactions —
which is the overwhelmingly common on-chain shape.

## Reference (cardano-ledger)

cardano-ledger's transaction-info translation (`transVITime` /
`Cardano.Ledger.BaseTypes.Interval`) maps the `ValidityInterval` to a Plutus
`POSIXTimeRange` using `closedLowerBound` for the lower bound and
`strictUpperBound` for the upper bound. `strictUpperBound x = UpperBound x
False`. There is no code path that produces a closed finite upper bound; the
"closed upper for an upper-only interval" behavior does not exist in
cardano-ledger. (Plutus `Interval`/`UpperBound`: `plutus-ledger-api`,
`PlutusLedgerApi.V1.Interval`.)

## Proof — ExUnits on a real preview transaction

Preview tx `197f1340…`, spending redeemer (tag 0, index 2), Plutus V3 reference
script `2b91a7e6…`. Four independent evaluators agree the upper bound is
EXCLUSIVE:

| Evaluator | validity upper closure | CPU (steps) | Memory |
|---|---|---|---|
| gouroboros ScriptContext + plutigo (current `main`) | True (inclusive) | 955,578,798 | 3,316,001 |
| gouroboros (patched) + plutigo | **False (exclusive)** | **954,938,034** | **3,313,597** |
| independent `uplc` CEK on the same script | False (exclusive) | 954,938,034 | 3,313,597 |
| cardano-cli (real preview node) | — | 954,938,034 | 3,313,597 |
| tx-declared ExUnits (on-chain witness) | — | 954,938,034 | 3,313,597 |

Flipping ONLY this closure Bool (no other change) makes plutigo AND an
independent uplc CEK land on the cardano-cli / on-chain value exactly. The
control redeemer (tag 0, index 0), which does not read the validity range, is
unaffected: 47,909,674 / 146,410 both before and after.

## Additional evidence inside gouroboros itself

The existing regression test `TestValidityRangeMatchesCardanoLedger`
(`context_validity_test.go`) and the `Withdraw` golden in
`TestScriptContextV3` (`context_test.go`) both bake in the same
`!startPresent` / closed-upper assumption, so they currently "pass" against the
buggy encoder — i.e. the tests encode the same wrong convention rather than
catching it.

## Fix

Make the upper-bound closure `false` whenever the upper bound is present
(matching `strictUpperBound`), and correct the two tests' expectations. PR to
follow.

wolf31o2:
Fix TTL-only validity ranges to use exclusive upper bounds.

Closes #2160

Tests: `go test ./ledger/common/script`

### DIFF
--- a/ledger/common/script/context.go
+++ b/ledger/common/script/context.go
@@ -26,6 +26,9 @@ import (
 	"github.com/blinklabs-io/plutigo/data"
 )
 
+// eraIdConway is the first era with strict validity upper bounds.
+const eraIdConway = 6
+
 type ScriptContext interface {
 	isScriptContext()
 	ToPlutusData() data.PlutusData
@@ -194,8 +197,7 @@ func (t TxInfoV1) ToPlutusData() data.PlutusData {
 	)
 }
 
-// NewTxInfoV1FromTransaction builds a Plutus V1 TxInfo. The upper validity
-// bound is always exclusive, matching cardano-ledger's strictUpperBound.
+// NewTxInfoV1FromTransaction builds a Plutus V1 TxInfo.
 func NewTxInfoV1FromTransaction(
 	slotState lcommon.SlotState,
 	tx lcommon.Transaction,
@@ -250,8 +252,7 @@ func NewTxInfoV1FromTransaction(
 // StrictValidityUpperBoundForTransaction reports whether the transaction's
 // era uses an exclusive upper validity bound in Plutus script contexts.
 func StrictValidityUpperBoundForTransaction(tx lcommon.Transaction) bool {
-	_ = tx
-	return true
+	return tx.Type() >= eraIdConway
 }
 
 type TxInfoV2 struct {
@@ -317,8 +318,7 @@ func (t TxInfoV2) ToPlutusData() data.PlutusData {
 	)
 }
 
-// NewTxInfoV2FromTransaction builds a Plutus V2 TxInfo. The upper validity
-// bound is always exclusive, matching cardano-ledger's strictUpperBound.
+// NewTxInfoV2FromTransaction builds a Plutus V2 TxInfo.
 func NewTxInfoV2FromTransaction(
 	slotState lcommon.SlotState,
 	tx lcommon.Transaction,
@@ -488,6 +488,7 @@ type TimeRange struct {
 	upperBound        uint64
 	lowerBoundPresent bool
 	upperBoundPresent bool
+	strictUpperBound  bool
 }
 
 func (t TimeRange) ToPlutusData() data.PlutusData {
@@ -533,8 +534,7 @@ func (t TimeRange) ToPlutusData() data.PlutusData {
 			t.upperBound,
 			t.upperBoundPresent,
 			false,
-			// cardano-ledger's strictUpperBound is always exclusive.
-			false,
+			!t.lowerBoundPresent && !t.strictUpperBound,
 		),
 	)
 }
@@ -612,9 +612,10 @@ func sortedRedeemerKeys(
 func validityRangeInfo(
 	slotState lcommon.SlotState,
 	tx lcommon.Transaction,
-	_ bool,
+	strictValidityUpperBound bool,
 ) (TimeRange, error) {
 	var ret TimeRange
+	ret.strictUpperBound = strictValidityUpperBound
 	startSlot := tx.ValidityIntervalStart()
 	endSlot, upperBoundPresent := lcommon.TransactionValidityIntervalUpperBound(
 		tx,
--- a/ledger/common/script/context.go
+++ b/ledger/common/script/context.go
@@ -26,10 +26,6 @@ import (
 	"github.com/blinklabs-io/plutigo/data"
 )
 
-// eraIdConway is the first era with strict validity upper bounds. Keep this
-// numeric boundary here to avoid importing era packages into common/script.
-const eraIdConway = 6
-
 type ScriptContext interface {
 	isScriptContext()
 	ToPlutusData() data.PlutusData
@@ -198,11 +194,8 @@ func (t TxInfoV1) ToPlutusData() data.PlutusData {
 	)
 }
 
-// NewTxInfoV1FromTransaction builds a Plutus V1 TxInfo. strictValidityUpperBound
-// selects the era-dependent encoding of a finite validity-interval upper bound:
-// pass true in the Conway era or later (EXCLUSIVE upper bound in all cases) and
-// false in Alonzo/Babbage (CLOSED upper bound for an upper-only interval). See
-// the TimeRange.strictUpperBound documentation and cardano-ledger#3043.
+// NewTxInfoV1FromTransaction builds a Plutus V1 TxInfo. The upper validity
+// bound is always exclusive, matching cardano-ledger's strictUpperBound.
 func NewTxInfoV1FromTransaction(
 	slotState lcommon.SlotState,
 	tx lcommon.Transaction,
@@ -257,7 +250,8 @@ func NewTxInfoV1FromTransaction(
 // StrictValidityUpperBoundForTransaction reports whether the transaction's
 // era uses an exclusive upper validity bound in Plutus script contexts.
 func StrictValidityUpperBoundForTransaction(tx lcommon.Transaction) bool {
-	return tx.Type() >= eraIdConway
+	_ = tx
+	return true
 }
 
 type TxInfoV2 struct {
@@ -323,11 +317,8 @@ func (t TxInfoV2) ToPlutusData() data.PlutusData {
 	)
 }
 
-// NewTxInfoV2FromTransaction builds a Plutus V2 TxInfo. strictValidityUpperBound
-// selects the era-dependent encoding of a finite validity-interval upper bound:
-// pass true in the Conway era or later (EXCLUSIVE upper bound in all cases) and
-// false in Babbage (CLOSED upper bound for an upper-only interval). See the
-// TimeRange.strictUpperBound documentation and cardano-ledger#3043.
+// NewTxInfoV2FromTransaction builds a Plutus V2 TxInfo. The upper validity
+// bound is always exclusive, matching cardano-ledger's strictUpperBound.
 func NewTxInfoV2FromTransaction(
 	slotState lcommon.SlotState,
 	tx lcommon.Transaction,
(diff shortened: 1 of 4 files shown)


ledger/common/script/context.go
```diff
@@ -497,26 +488,7 @@ type TimeRange struct {
...
-	// but already use `strictUpperBound` (exclusive) when BOTH bounds are
-	// present. cardano-ledger#3043 could not change this pre-Conway behavior
-	// because it would alter historical on-chain script validation, so the
-	// corrected exclusive bound was gated to the Conway era.
-	//
-	// Set strictUpperBound = true when building a Plutus context in the Conway
-	// era or later, false for Alonzo/Babbage. Plutus V3 only exists in Conway
-	// and later, so V3 contexts always set this true.
-	strictUpperBound bool
+	strictUpperBound  bool
```
chrisguiney, line 491: This is where the change lands, and it is worth being explicit about what it is: after filtering comments out of this file's diff, the only remaining line is this field's alignment. `!t.lowerBoundPresent && !t.strictUpperBound` below is unchanged from `origin/main`.

So what this file actually does is drop the ~30 lines that explained *why* the encoding is era-dependent. Trimming those was defensible — they were long — but two of the facts in them are the ones I would most want to keep:

- **Why pre-Conway must stay inclusive.** cardano-ledger#3043 could not change Alonzo/Babbage `transVITime` because it would alter historical on-chain script validation, so the corrected exclusive bound was gated to Conway. Without that written down, the natural reading of this field is that the pre-Conway branch is a bug — which is exactly the change `ee76def` made in this PR before `a19d0de` took it back out.
- **Why the gate is on the era and not the Plutus version.** A V1/V2-keyed gate looks equivalent and is not: it would wrongly apply pre-Conway behaviour to Conway-era V1/V2 scripts. That one is genuinely hard to rediscover.

Would you be open to keeping those two, in a shorter form? Something like three lines pointing at cardano-ledger#3043 and at `transVITime` vs `transValidityInterval` would carry the load without the current length.

ledger/common/script/context_timerange_test.go
```diff
@@ -153,18 +127,17 @@ func TestTimeRangeToPlutusDataUpperBoundEraDependent(t *testing.T) {
...
 				lowerBound:        500,
 				lowerBoundPresent: true,
-				strictUpperBound:  false,
 			},
 			want: wholeRange(finiteBound(500, true), infBound(true)),
 		},
 		{
 			name: "unbounded - conway",
-			tr:   TimeRange{strictUpperBound: true},
+			tr:   TimeRange{},
```
chrisguiney, line 135: These two cases are now byte-identical — same `TimeRange{}`, same `want` — and the same is true of `"lower only (upper absent) - conway"` and `"- preconway"` just above, because `strictUpperBound` was dropped from both Conway variants.

Each pair reads as covering two eras and covers one twice. That is a little worse than it looks: `strictUpperBound` is what makes these Conway cases Conway, so if the era-invariant shapes ever stop being era-invariant, nothing here would notice.

Either restoring `strictUpperBound: true` on the two Conway variants, or collapsing each pair into one case named for the shape rather than the era, would fix it. I would lean toward restoring the field, since these are exactly the cases that prove the gate does *not* affect the era-invariant shapes.

ledger/common/script/context_timerange_test.go
```diff
@@ -56,38 +56,22 @@ func wholeRange(lower, upper data.PlutusData) data.PlutusData {
...
-//     in every case (Conway.transValidityInterval / cardano-ledger#3043).
-//   - Alonzo/Babbage (strictUpperBound == false): an upper-only interval uses
-//     PV1.to, a CLOSED/INCLUSIVE upper bound; a two-sided interval already uses
-//     strictUpperBound (EXCLUSIVE).
-//
-// The bug that motivated this test: gouroboros previously emitted
-// `!lowerBoundPresent` for every era, giving Conway-era TTL-only transactions
-// an INCLUSIVE upper bound and mis-computing script execution units.
-func TestTimeRangeToPlutusDataUpperBoundEraDependent(t *testing.T) {
+// TestTimeRangeToPlutusDataUpperBound pins era-dependent upper-bound encoding.
```
chrisguiney, line 59: The comment this replaces recorded the actual bug — that gouroboros previously emitted `!lowerBoundPresent` for every era, giving Conway-era TTL-only transactions an inclusive upper bound and mis-computing ExUnits — and it carried the era-vs-Plutus-version note I mentioned on `context.go`.

That text is the regression record for #2160. Removing it in the same PR that says `Closes #2160` reads backwards: the issue's own value is largely in that explanation, and this is the place a future reader would look for it.

The rename from `TestTimeRangeToPlutusDataUpperBoundEraDependent` is a small loss too — the old name said which property the table pins, and `...UpperBound` alone does not.

ledger/common/script/context_validity_test.go
```diff
@@ -72,7 +72,7 @@ func boolTag(value bool) uint64 {
 func expectedValidityRange(
 	start *uint64,
 	end *uint64,
-	upperBoundOnlyIsClosed bool,
+	upperClosed ...bool,
```
chrisguiney, line 75: A required `bool` became variadic here. I checked whether that could go wrong silently and it cannot: both current callers pass the argument, and a caller that omitted it would get `closed = false`, which mismatches the pre-Conway upper-only expectation and fails loudly. So this is not a trap.

It does still trade a compile-time check for a runtime default in a helper whose entire job is to state the expected closure, and I could not find a caller that benefits from the argument being optional. If there is one I have missed, I would rather know — otherwise keeping the plain `bool` seems strictly better here.

ledger/common/script/context_validity_era_test.go
```diff
@@ -208,7 +201,7 @@ func TestValidityRangeUpperBoundByEra(t *testing.T) {
 							)
 							require.NoError(t, err)
 							strictUpperBound := build.name == "V3" ||
-								!era.upperBoundOnlyIsClosed
+								era.name == "Conway" || era.name == "Dijkstra"
```
chrisguiney, line 204: Minor, and I checked before raising it: this fails loudly rather than silently. Adding a post-Conway era to the table, or renaming an existing one, produces a mismatch against the production encoding and the test fails — so correctness is not at risk.

It is a maintainability step back from the `upperBoundOnlyIsClosed` column it replaces, though. The expectation now lives ~110 lines from the table that defines the eras, and each era row no longer states its own expected behaviour, so adding one means editing two places instead of filling in a column. The old shape also documented the property per era at the point where the era is introduced, which is the part I would miss most.

chrisguiney (changes requested):
Thanks for picking this up — and for catching your own regression in `a19d0de` before anyone had to ask. That said, I do not think this PR should merge as it stands, and the reason is more about what it claims than about anything being broken.

## The net production change is one whitespace character

Filtering comment lines out of the `context.go` diff leaves exactly this:

```
-	strictUpperBound bool
+	strictUpperBound  bool
```

The closure expression `!t.lowerBoundPresent && !t.strictUpperBound` is byte-identical to `origin/main`, so the title ("encode validity upper bounds as exclusive") and the body ("Fix TTL-only validity ranges to use exclusive upper bounds") describe a change that is not in the diff.

That is because the two commits cancel. `ee76def` removed the era gate outright — deleted `eraIdConway`, stubbed the era check to `_ = tx; return true`, hardcoded the closure to `false`, and reduced `strictValidityUpperBound` to an ignored `_ bool`. `a19d0de` then restored all of it, line for line. Cubic flagged the intermediate state at P1 and it was right: that version would have given Alonzo/Babbage TTL-only transactions an exclusive upper bound, diverging from `transVITime`'s `PV1.to` for historical script validation — the same class of divergence #2160 describes, pointed the other way.

## `Closes #2160` does not hold

`924de23 fix(ledger): era-gate ScriptContext validity upper bound` is already on `origin/main` and is an ancestor of this branch, so the issue is fixed. Merging this would close #2160 against a commit that changed no behaviour and break the trail back to the commit that actually fixed it.

## Verified good

- **The era gate is still pinned.** I mutated `eraIdConway` from 6 to 7 and `TestValidityRangeUpperBoundByEra/Conway` fails, so that coverage survived the edit.
- `go build ./...`, `go vet ./ledger/...`, `gofmt`, `golangci-lint run ./ledger/common/script/...` (0 issues) and `go test ./...` all pass.
- No concurrency surface — pure value encoding, no goroutines or shared state — and no performance impact.

I also checked the two things that looked like they might be silent traps, and neither is: adding or renaming an era in the table fails loudly rather than passing with a wrong expectation, and a caller that omits the new variadic argument fails loudly on the pre-Conway cases. Both are maintainability points, not correctness ones, and I have marked them as such inline.

## Suggested disposition

Closing this and letting `924de23` close #2160 is the cleanest option. If the comment volume is the real motivation — and some of those blocks genuinely were long — a re-scoped PR titled as a docs/test cleanup would be welcome, keeping the two facts called out inline, restoring `strictUpperBound` on the Conway test cases so the duplicated pairs test different things, and dropping the `Closes` line. Happy to review that quickly.

## Process note

Cubic reviewed at 22:03Z, before `a19d0de` landed at 22:47Z, so it has not seen the head commit. Its one remaining open thread is now a false positive — the comment it flags reads correctly again after the revert — and CodeRabbit's green check says "Review skipped: incremental reviews are disabled".

ledger/common/script/context_validity_era_test.go
```diff
@@ -89,13 +87,11 @@ func TestValidityRangeEraIdsUnchanged(t *testing.T) {
 // on.
 func TestValidityRangeUpperBoundByEra(t *testing.T) {
 	for _, era := range []struct {
-		name                   string
-		upperBoundOnlyIsClosed bool
-		tx                     eraTxBuilder
+		name string
```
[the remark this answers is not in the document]
chrisguiney: Checked this against the current head rather than the commit it was raised on: it is a false positive now.

Cubic reviewed at 22:03Z, which was after `ee76def` (where the comment genuinely did contradict the code) but before `a19d0de` at 22:47Z restored the era gate. At `a19d0de` the comment reads "closed in Alonzo and Babbage, exclusive from Conway on", and `TestValidityRangeUpperBoundByEra` does expect exactly that per era, so no change is needed.

Worth noting cubic has not seen the head commit at all — its P1 finding on the pre-Conway regression was the useful one, and that is already addressed by the revert.

wolf31o2:
Disposition: this PR is redundant.

`924de23` is already on `main` and contains the actual era-gated validity-bound fix for #2160. After the reverted implementation changes, this branch has no net production behavior change; the remaining production diff is whitespace only.

Closing this PR so #2160 remains attributed to the commit that fixed it.

wolf31o2:
Superseded by the existing fix in `main` (`924de23`); no net production behavior remains in this branch.

PR #2192, opened 2026-09-04
fix(consensus): preserve leader threshold boundaries

### LINKED ISSUE #2093: consensus: verify leader-threshold boundary behavior
The leader-threshold comparison differs at the exact integer boundary.

Acceptance criteria:

- Align the equality case with the reference calculation.
- Add below, equal, and above-boundary vectors for each supported consensus mode.
- Exercise the production leader-election entry point.

wolf31o2:
## Summary

- preserve the strict real-valued leader cutoff with an integer ceiling threshold
- cover below, equal, and above boundaries through the production CPraos and TPraos entry point

## Testing

- `make test`
- `make lint`
- `make format`
- `make golines`
- `GOFLAGS=-buildvcs=false make build`
- focused consensus tests with `-race`

Closes #2093.

---
## Summary by cubic
Switches the leadership threshold from `floor(X)` to `ceil(X)` so integer leader values compare against the real-valued cutoff exactly, fixing off-by-one rejection when the cutoff is an exact integer.

- Applies the ceiling in the general precision-escalated path and the exact-rational fast path in `consensus/threshold.go`, plus the public `IsSlotLeaderWithMode` entry point for CPraos and TPraos.
- Adds boundary tests covering leader values immediately below, at, and above the cutoff; updates existing tests whose expected threshold shifted by one (e.g., near-one-f, full-stake and partial-stake exact cases).
- Closes #2093.

<sup>Written for commit a11d8d1e2aa9da1a9a6721fb714e82d61059533a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2192?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated leader eligibility thresholds to use ceiling-based rounding, improving accuracy at fractional and exact cutoff boundaries.
  - Corrected threshold behavior for both CPraos and TPraos consensus modes, including values exactly at the eligibility limit.

- **Tests**
  - Added coverage for threshold rounding, boundary conditions, and precision across supported consensus modes.

### DIFF
--- a/consensus/leader.go
+++ b/consensus/leader.go
@@ -32,7 +32,7 @@ type LeaderElectionResult struct {
 	Proof []byte
 	// Output is the VRF output (nil only on early-return cases like zero stake)
 	Output []byte
-	// Threshold is the leadership threshold that was used
+	// Threshold is the integer comparison threshold ceil(T) that was used
 	Threshold *big.Int
 }
 
--- a/consensus/leader_test.go
+++ b/consensus/leader_test.go
@@ -28,6 +28,18 @@ import (
 // Test seed (exactly 32 bytes)
 var testVRFSeed = []byte("test_vrf_seed_for_consensus!!!!!")
 
+type fixedOutputVRFSigner struct {
+	output []byte
+}
+
+func (s fixedOutputVRFSigner) Prove([]byte) ([]byte, []byte, error) {
+	return make([]byte, 80), append([]byte(nil), s.output...), nil
+}
+
+func (fixedOutputVRFSigner) PublicKey() []byte {
+	return make([]byte, 32)
+}
+
 func TestNewSimpleVRFSigner(t *testing.T) {
 	signer, err := NewSimpleVRFSigner(testVRFSeed)
 	require.NoError(t, err, "NewSimpleVRFSigner failed")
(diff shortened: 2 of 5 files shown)


cryptodj413:
## Review summary

No merge blockers. The core fix — comparing the integer VRF leader value against `ceil(T)` instead of `floor(T)`, where `T = upperBound*(1-(1-f)^σ)` — is algebraically correct: for integer `v` and real `T`, `v < ceil(T) ⟺ v < T` in every case (integral `T`, non-integral `T`), with no mirror-image over-admission at any boundary. Confirmed this matches upstream cardano-ledger's own stated predicate (`checkLeaderNatValue`: `p < 1-(1-f)^σ`, a strict real-valued `<`). The old `floor`-based code was a liveness bug (a genuinely eligible leader wrongly rejected at an integral cutoff), never a safety bug — worth stating in the PR description.

Verified independently:
- `ceilBigFloat`'s `accuracy == big.Below` check is the only case needing `+1`, confirmed against Go's actual `big.Float.Int` accuracy contract (probed empirically, not just read from docs).
- The new `QuoRem`-based ceiling in `exactOneMinusFPowerSigmaThreshold` is sound given `probabilityExact.Denom() > 0` (always true) and `thresholdNumerator ≥ 0` (holds because `f ∈ (0,1)` is enforced before this call) — this is a load-bearing but currently unstated invariant (see below).
- The precision-escalation loop's cap-reached predicate (`thresholdLo.Cmp(thresholdHi) == 0`) is still sound under `ceil` — monotonicity gives `ceil(lo) ≤ ceil(T) ≤ ceil(hi)`, same argument as before under `floor`.
- All three in-repo call sites of the changed threshold functions (`consensus/leader.go:168,254`, `consensus/validate.go:425`) feed the result into a strict `<` comparison, so all are correct under the new semantics.
- Reverted only `consensus/threshold.go` to `origin/main` in a separate worktree (keeping the head tests): `TestIsSlotLeaderWithModeRealThresholdBoundaries`, `TestLeaderEligibilityBoundaryAgainstIndependentReference`, and `TestCertifiedNatThresholdNearOneDenominatorPrecisionLoss` all genuinely fail without the fix and pass with it.
- The two exact-rational rows removed from `TestLeaderEligibilityBoundaryAgainstIndependentReference`'s differential table (`f=0.5,σ=1`, `f=0.75,σ=0.5`) were removed for a real reason, not to hide a regression: probing `refThreshold` directly shows the *reference* implementation is off by +1 at those exact integral cutoffs (its `refLn`/`refExp` residual sits just above the true value), which `ceil` now surfaces where `floor` used to mask it. Equivalent coverage survives via exact arithmetic in `threshold_test.go` (`TestCertifiedNatThresholdFullStakeExactHalf`, `TestCertifiedNatThresholdPartialStakeExactCutoff`).
- `golangci-lint run ./...`, `gofmt -s -l`, and `go test -race` are all clean; no new line in the diff exceeds the repo's 80-column convention.

### API/wire note (non-blocking)
The exported semantics of `CertifiedNatThreshold`, `CertifiedNatThresholdWithMode`, and `LeaderElectionResult.Threshold` shift by `+1` for essentially every non-integral `(f, σ)`. Every in-repo consumer is safe (all feed straight into a strict `<`), but an external consumer computing "largest eligible value" as `threshold-1`, or comparing with `<=`, would shift by one. Worth a release-note line since there's no CHANGELOG.

### Test coverage (non-blocking)
- `TestCertifiedNatThresholdFullStakeExactHalf` (`threshold_test.go:1318`) asserts only the threshold value, unlike its sibling `TestCertifiedNatThresholdPartialStakeExactCutoff`, which also runs the below/at eligibility assertions the deleted boundary row provided. Cheap to restore parity.
- `checkBelowThreshold`'s `FillBytes` (`threshold_test.go:50`) panics if `above = cutoff+1` overflows `bits` — not triggered by the current table, but `ceil` can now return exactly `upperBound` (confirmed reachable, see `TestCertifiedNatThresholdNearOneDenominatorPrecisionLoss`'s updated expected value), so a future high-`f` row could hit this. Test-only.
- `exactOneMinusFPowerSigmaThreshold`'s `QuoRem` ceiling relies on `thresholdNumerator ≥ 0`, which holds only because `f ∈ (0,1)` is enforced earlier in the call chain — that invariant isn't stated at the call site and isn't defended the way `ceilBigFloat` twelve lines below defends the same case for negatives. A one-line comment would stop a future relaxation of the `f` domain guard from silently breaking this with nothing catching it.

### Documentation (non-blocking)
- `threshold_boundary_test.go`'s trimmed NOTE (replacing the ~25-line comment describing gouroboros's remaining divergence from cardano-node) drops the still-true, still-larger residual: cardano-node's own `Fixed E34` internal precision (~113 bits, ~2^143 leader values wide) is a much bigger gap than the 1-integer (`2^-256`) gap this PR closes. The new four-line comment reads as if gouroboros is now bit-exact with cardano-node at the boundary, which it isn't (verified against `BaseTypes.hs`'s `FixedPoint`/`taylorExpCmp` quantization). Worth keeping two sentences of the old NOTE.
- `ceilBigFloat` panics (nil-deref) on ±Inf input — unreachable from current callers (verified: no finite-input path in `oneMinusFPowerSigmaBounds` can produce it), but worth an explicit "x must be finite" doc line since it's the kind of helper a future caller might reuse.

### Skipped checks
- No live Cardano devnet/`cardano-node` interop run available — the ~2^-113 residual-divergence claim above is established by reading upstream Haskell source (`checkLeaderNatValue`, `Fixed E34`/`taylorExpCmp`), not by executing a node. That boundary has probability ~10^-34 per check, so a devnet run wouldn't exercise it regardless.
- No Antithesis run; no fuzz target added for `ceilBigFloat` (none of the 19 existing fuzz targets cover `consensus/threshold.go` — a differential fuzz against exact `big.Rat` ceiling would be cheap future work, not a merge condition).
- `make test` not run verbatim (runs `go mod tidy` and iterates `examples/*`, unaffected by this diff); ran `go test ./...` on the root module instead.

cryptodj413 (approved):

PR #2193, opened 2026-09-04
fix(consensus): enforce Byron delegation epoch ranges

### LINKED ISSUE #2090: consensus: enforce Byron light-delegation epoch ranges
The legacy Byron `HeaderValidator` parses type-1 light and type-2 heavy proxy
certificate constraints as the same scalar value. Canonical light certificates
carry an inclusive `[epochLow, epochHigh]` interval and are valid only when the
header epoch is inside it. Heavy certificates carry an unconstrained omega.

Acceptance criteria:

- Decode the light certificate interval in its canonical wire shape.
- Require `epochLow <= headerEpoch <= epochHigh`.
- Reject malformed or inverted light intervals.
- Preserve unconstrained heavy-certificate omega behavior.
- Cover current, expired, premature, boundary, malformed, and heavy cases.

wolf31o2:
Fixes #2090.

Byron light proxy certificates now decode and enforce their inclusive epoch
range against the header epoch, including malformed and inverted ranges.
Heavy proxy certificates retain unconstrained omega semantics. Validation uses
the preserved wire encoding for certificate signatures.

Regression coverage includes current, expired, premature, boundary,
malformed, type-mismatch, heavy, and non-shortest CBOR cases.

### DIFF
--- a/consensus/byron/pbft_test.go
+++ b/consensus/byron/pbft_test.go
@@ -20,12 +20,13 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/blinklabs-io/gouroboros/cbor"
 	ledgerbyron "github.com/blinklabs-io/gouroboros/ledger/byron"
 	"github.com/blinklabs-io/gouroboros/ledger/common"
 	"github.com/stretchr/testify/require"
 )
 
-func realPBFTHeaderFixture(
+func heavyPBFTHeaderFixture(
 	t *testing.T,
 ) (*ledgerbyron.ByronMainBlockHeader, ByronConfig, PBFTIssuer) {
 	t.Helper()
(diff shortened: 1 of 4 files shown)


arepala-uml (approved):
Verified in an isolated worktree: `go build ./...`, `go test -race ./consensus/byron/...`, `go test ./internal/test/conformance/...`, `go vet`, and `golangci-lint run` all pass clean on the PR head.

Traced the full path (`validateProxySignature` → `proxyCertIndexCbor` → `validateProxyCertIndex` → `validateDelegationCertSignature`): the recovered-wire-bytes-vs-decoded-value check, inverted-range check, and header-epoch inclusion check are correctly ordered and covered (current/expired/premature/boundary/malformed/type-mismatch/heavy/non-shortest-CBOR). `TestValidateProxySignaturePreservesNonShortestEpochRange` fails without the wire-preservation fix, confirming it's load-bearing.

All prior CodeRabbit/Cubic findings on this branch (the type-1/type-2 constant inversion, the redundant header decode, the missing non-shortest-encoding test) were already fixed in later commits at HEAD.

Two non-blocking notes, not requesting changes:
- `consensus/byron/validate_test.go`: one new line (`require.Equal(t, []any{currentEpoch, currentEpoch}, ...)`) is 93 cols; `make golines` would reformat it.
- `VerifyDelegationCertificateSignature` is now reused for light-cert verification, but its signing format was only ever validated against a real *heavy* mainnet certificate — no live light-type (type-1) certificate exists on any chain to verify the assumption against.

PR #2198, opened 2026-09-04
fix(ledger): count a duplicated redeemer's budget once

chrisguiney:
Reported as blinklabs-io/dingo#3875.

The last-wins half of this work is already on `main` via #2200. This PR is now
only the second half: the execution-unit budget of a duplicated redeemer key.

## The defect

`UtxoValidateExUnitsTooBigUtxo` summed `WsRedeemers.Redeemers`, the raw wire
list, so a key appearing more than once contributed its budget once per copy.
cardano-ledger sums over the map:

```haskell
totExUnits tx = foldMap snd $ tx ^. witsTxL . rdmrsTxWitsL . unRedeemersL
```

A duplicated key contributes once there, because `unRedeemersL` is a `Map`.

## How it surfaced

After #2200 landed the last-wins fix, a Preview replay advanced 1231 slots and
stopped again:

```
ExUnits too big: total 5060693388/19936728 steps/memory, maximum 10000000000/14000000
```

Transaction `3ace3bc7f4c5…` at slot 12925989 carries the same `(mint, 0)`
redeemer six times at `[3322788, 843448898]`. `6 x 3322788 = 19936728` is
exactly the reported memory. Collapsed, the transaction is well inside both
limits.

## The change

Alonzo and Babbage now iterate `WsRedeemers.Iter()`, the collapsed view, for the
budget. Conway and Dijkstra already did; all four eras now agree. Slot 12925989
is a Babbage block, so Babbage is where it bites in practice.

Collapsing can only lower a sum, so no transaction that validated before can be
newly rejected by this.

The raw list is deliberately left in use for the script data hash, which has to
reproduce the bytes as they were signed.

## Tests

Both eras get a `duplicate redeemer key counts once` accept case: six copies of
one `(mint, 0)` redeemer at 3,000,000 memory / 3,000 steps against a
5,000,000 / 5,000 cap, so one copy fits and *two* already exceed it. Only a
collapse to exactly one entry passes; an over-count of any size fails, not just
a full six.

Reverting only the era's `rules.go` change fails that era's test, at six copies
(`18000000` memory) and at two (`6000000`).

The existing `ExUnits overflow` case in both suites built two redeemers that
both defaulted to key `(0, 0)`, so it depended on duplicates summing. Its
subject is addition overflowing, not duplicate handling, so the entries now
carry distinct indexes and exercise the same arithmetic.

`go build ./...`, `go test ./... -count=1` (39 packages) and `make lint` (0 issues) pass.

### DIFF
--- a/ledger/alonzo/rules.go
+++ b/ledger/alonzo/rules.go
@@ -233,8 +233,18 @@ func UtxoValidateExUnitsTooBigUtxo(
 	if !ok {
 		return errors.New("transaction is not expected type")
 	}
+	// Iterate the collapsed view rather than the raw list. The wire format is
+	// a list but the ledger holds a map, so a key appearing more than once
+	// contributes its budget once. Summing the raw list multiplies it by the
+	// number of copies: on Preview, transaction 3ace3bc7f4c5… at slot 12925989
+	// carries the same (mint, 0) redeemer six times at 3322788/843448898, and
+	// summing all six gives 19936728 memory against a 14000000 cap, rejecting
+	// a transaction the network accepted (blinklabs-io/dingo#3875).
+	//
+	// The raw list is still the right thing for the script data hash, which has
+	// to reproduce the bytes as they were signed.
 	var totalSteps, totalMemory int64
-	for _, redeemer := range tmpTx.WitnessSet.WsRedeemers.Redeemers {
+	for _, redeemer := range tmpTx.WitnessSet.WsRedeemers.Iter() {
 		newSteps, ok := common.AddInt64Checked(
 			totalSteps,
 			redeemer.ExUnits.Steps,
--- a/ledger/babbage/rules.go
+++ b/ledger/babbage/rules.go
@@ -1034,8 +1034,13 @@ func UtxoValidateExUnitsTooBigUtxo(
 	if !ok {
 		return errors.New("transaction is not expected type")
 	}
+	// Iterate the collapsed view rather than the raw list: the wire format is a
+	// list but the ledger holds a map, so a key appearing more than once
+	// contributes its budget once. See the Alonzo rule of the same name; this
+	// is where it bites in practice, since Preview transaction 3ace3bc7f4c5 at
+	// slot 12925989 is a Babbage-era block (blinklabs-io/dingo#3875).
 	var totalSteps, totalMemory int64
-	for _, redeemer := range tmpTx.WitnessSet.WsRedeemers.Redeemers {
+	for _, redeemer := range tmpTx.WitnessSet.WsRedeemers.Iter() {
 		newSteps, ok := common.AddInt64Checked(
 			totalSteps,
 			redeemer.ExUnits.Steps,
--- a/ledger/babbage/rules_test.go
+++ b/ledger/babbage/rules_test.go
@@ -1336,6 +1336,38 @@ func TestUtxoValidateExUnitsTooBigUtxo(t *testing.T) {
 			Steps:  5_000,
 		},
 	}
+	// A duplicated key contributes its budget once, as it does in the map the
+	// ledger holds. Preview transaction 3ace3bc7f4c5 at slot 12925989 is a
+	// Babbage block carrying the same (mint, 0) redeemer six times; summing the
+	// raw list rejected a transaction the network accepted
+	// (blinklabs-io/dingo#3875).
+	t.Run(
+		"duplicate redeemer key counts once",
+		func(t *testing.T) {
+			dup := alonzo.AlonzoRedeemer{
+				Tag:   common.RedeemerTagMint,
+				Index: 0,
+				ExUnits: common.ExUnits{
+					Memory: 3_000_000,
+					Steps:  3_000,
+				},
+			}
+			testTx.WitnessSet.WsRedeemers = alonzo.AlonzoRedeemers{
+				Redeemers: []alonzo.AlonzoRedeemer{dup, dup, dup, dup, dup, dup},
+			}
+			// One copy is inside the 5,000,000 / 5,000 cap and two already
+			// exceed it, so only a collapse to exactly one entry passes: an
+			// over-count of any size, not just all six, fails here.
+			if err := babbage.UtxoValidateExUnitsTooBigUtxo(
+				testTx, testSlot, testLedgerState, testProtocolParams,
+			); err != nil {
+				t.Errorf(
+					"six copies of one redeemer must count once, got: %v", err,
+				)
+			}
+		},
+	)
+
 	// Ex-units too large
 	t.Run(
 		"ExUnits too large",
@@ -1391,15 +1423,20 @@ func TestUtxoValidateExUnitsTooBigUtxo(t *testing.T) {
 	t.Run(
 		"ExUnits overflow",
 		func(t *testing.T) {
+			// Distinct keys: this case is about addition overflowing, and
+			// two entries sharing a key would collapse to one before the
+			// sum ever happens.
 			testTx.WitnessSet.WsRedeemers = alonzo.AlonzoRedeemers{
 				Redeemers: []alonzo.AlonzoRedeemer{
 					{
+						Index: 0,
 						ExUnits: common.ExUnits{
 							Memory: math.MaxInt64 - 10,
 							Steps:  math.MaxInt64 - 10,
 						},
 					},
 					{
+						Index: 1,
 						ExUnits: common.ExUnits{
 							Memory: 100,
 							Steps:  100,
--- a/ledger/alonzo/rules_test.go
+++ b/ledger/alonzo/rules_test.go
@@ -1256,6 +1256,39 @@ func TestUtxoValidateExUnitsTooBigUtxo(t *testing.T) {
 			Steps:  5_000,
 		},
 	}
+	// A duplicated key contributes its budget once, as it does in the map the
+	// ledger holds. Preview transaction 3ace3bc7f4c5 at slot 12925989 carries
+	// the same (mint, 0) redeemer six times; summing the raw list rejected a
+	// transaction the network accepted (blinklabs-io/dingo#3875). That block is
+	// Babbage, but the rule is duplicated per era and the Alonzo copy summed
+	// the raw list the same way.
+	t.Run(
+		"duplicate redeemer key counts once",
+		func(t *testing.T) {
+			dup := alonzo.AlonzoRedeemer{
+				Tag:   common.RedeemerTagMint,
+				Index: 0,
+				ExUnits: common.ExUnits{
+					Memory: 3_000_000,
+					Steps:  3_000,
+				},
+			}
+			testTx.WitnessSet.WsRedeemers = alonzo.AlonzoRedeemers{
+				Redeemers: []alonzo.AlonzoRedeemer{dup, dup, dup, dup, dup, dup},
+			}
+			// One copy is inside the 5,000,000 / 5,000 cap and two already
+			// exceed it, so only a collapse to exactly one entry passes: an
+			// over-count of any size, not just all six, fails here.
+			if err := alonzo.UtxoValidateExUnitsTooBigUtxo(
+				testTx, testSlot, testLedgerState, testProtocolParams,
+			); err != nil {
+				t.Errorf(
+					"six copies of one redeemer must count once, got: %v", err,
+				)
+			}
+		},
+	)
+
 	// Ex-units too large
 	t.Run(
 		"ExUnits too large",
@@ -1311,15 +1344,20 @@ func TestUtxoValidateExUnitsTooBigUtxo(t *testing.T) {
 	t.Run(
 		"ExUnits overflow",
 		func(t *testing.T) {
+			// Distinct keys: this case is about addition overflowing, and
+			// two entries sharing a key would collapse to one before the
+			// sum ever happens.
 			testTx.WitnessSet.WsRedeemers = alonzo.AlonzoRedeemers{
 				Redeemers: []alonzo.AlonzoRedeemer{
 					{
+						Index: 0,
 						ExUnits: common.ExUnits{
 							Memory: math.MaxInt64 - 10,
 							Steps:  math.MaxInt64 - 10,
 						},
 					},
 					{
+						Index: 1,
 						ExUnits: common.ExUnits{
 							Memory: 100,
 							Steps:  100,


chrisguiney:
**Acceptance on the real chain, and a correction: the first two commits were not enough.**

Restarting the wedged node on its own data directory, so it had only to get past the blocks that stopped it:

```
committed tip 12924731  ->  12932637
plutus disagreements  0   (was 786)
ExUnits rejections    0
```

Both blocks now apply: 12924758 (the redeemer the script errored on) and 12925989 (the budget summed six times).

## What the first attempt missed

Fixing `Value` and `Iter` cleared the Plutus divergence, and the node immediately advanced 1231 slots and stopped again on a different transaction:

```
ExUnits too big: total 5060693388/19936728 steps/memory, maximum 10000000000/14000000
```

`3ace3bc7f4c5…` carries the same `(mint, 0)` redeemer six times at `[3322788, 843448898]`, and `6 x 3322788 = 19936728` is exactly the reported memory. `UtxoValidateExUnitsTooBigUtxo` sums the raw list, which `Iter` does not touch.

Then the second attempt fixed only the **Alonzo** copy of that rule. There are four — Alonzo, Babbage, Conway, Dijkstra — and Preview epoch 149 is Babbage, so the node behaved identically and I had to be told again by the same log line. Conway and Dijkstra already iterate the collapsed view; all four now agree, and I checked that none still sums the raw list rather than assuming.

## One thing deliberately left on the raw list

The script data hash re-encodes `WsRedeemers.Redeemers` directly, and it has to reproduce the bytes as they were signed. Collapsing there would change the hash, so that call site is untouched.

## A latent flaw in two existing tests

The overflow case in both the Alonzo and Babbage tests built two redeemers that both defaulted to key `(0, 0)`, so it depended on duplicates summing. Its subject is addition overflowing rather than duplicate handling, so the entries now carry distinct indexes and exercise the same arithmetic.

ledger/babbage/rules.go
```diff
@@ -1227,6 +1172,27 @@ func UtxoValidateScriptDataHash(
 		return common.MissingScriptDataHashError{}
 	}
 
+	// The language views cover the Plutus scripts some script purpose of this
+	// transaction requires, not every script it can reach. A reference script
+	// on a spent or referenced input that no purpose needs is inert data (see
+	// the comment above), and counting it adds a view the producer did not,
+	// which rejects a canonical transaction on a hash it never declared
+	// (gouroboros #2188).
+	view, err := script.NewTxScriptView(tx, ls)
```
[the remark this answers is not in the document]
chrisguiney: This finding was against code that is not part of this change — it came from an already-merged commit my branch was accidentally carrying.

I branched from tag `v0.202.5` and cherry-picked the #2189 script-data-hash fix so a local node could build against it, not noticing that `main` already contained it. The PR therefore showed that commit and roughly 250 lines of unrelated diff, including the `NewTxScriptView` error handling you are pointing at, and it also failed DCO because that commit is not mine to sign off.

Rebased onto current `main`, so the PR is now three commits and six files, all redeemer-related. `ledger/babbage/rules.go:1186` is untouched by it.

If the concern about consumed-input versus reference-input error precedence is real against `main` as it stands, it is worth its own issue rather than being lost here — I have not evaluated it, since it is not my code and not in this diff.

chrisguiney:
@coderabbitai review

ledger/conway/rules.go
```diff
@@ -1617,6 +1562,27 @@ func UtxoValidateScriptDataHash(
...
 
+	// The language views cover the Plutus scripts some script purpose of this
+	// transaction requires, not every script it can reach. A reference script
+	// on a spent or referenced input that no purpose needs is inert data (see
+	// the comment above), and counting it adds a view the producer did not,
+	// which rejects a canonical transaction on a hash it never declared
+	// (gouroboros #2188).
+	view, err := script.NewTxScriptView(tx, ls)
+	if err != nil {
+		if errors.Is(err, common.ErrInputResolution) {
```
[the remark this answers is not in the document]
chrisguiney: Same as the sibling thread: this is against code that is not part of this change.

`ledger/conway/rules.go` is not among the files this PR touches. It appeared in the diff because my branch was carrying an already-merged commit — I branched from tag `v0.202.5` and cherry-picked the #2189 script-data-hash fix so a local node could build, without noticing `main` already had it.

Timing confirms it: both of your inline findings were created at 18:06:47Z, and the rebase landed at 18:53:17Z, so they describe the pre-rebase diff.

The PR is now three commits across six files — `ledger/alonzo/{alonzo,rules,rules_test,duplicate_redeemer_test}.go` and `ledger/babbage/{rules,rules_test}.go` — all redeemer-related. A fresh pass over that would be genuinely useful; the current findings cannot be acted on here.

If the spent-input versus reference-input precedence concern holds against `main` as it stands, it deserves its own issue rather than being lost on an unrelated PR. I have not evaluated it.

chrisguiney:
**Note for reviewers, human and bot:** this branch was force-pushed at 18:53Z to drop an already-merged commit it was accidentally carrying.

I branched from tag `v0.202.5` and cherry-picked the #2189 script-data-hash fix so a local node could build against it, not noticing that `main` already contained it. The PR therefore showed four commits and nine files, failed DCO on a commit that is not mine to sign off, and drew two bot findings against `ledger/conway/rules.go` and `ledger/babbage/rules.go:1339` — none of which this change touches.

Rebased onto current `main`:

```
before: 4 commits, 9 files
after:  3 commits, 6 files
```

The two existing inline findings (created 18:06Z) predate that push and refer to the old diff; I have replied on both threads rather than leaving them unanswered. A fresh review over the current six files would be welcome.

`go build ./...`, `go vet ./ledger/...` and `go test ./...` (39 packages) pass on the rebased branch, and all three commits carry sign-off.

wolf31o2 (commented):
Half of this is already on `main`. #2200 landed as 9c07b06 and implements the same two things in `ledger/alonzo/alonzo.go`: last-wins in `Value`, duplicate-collapsing in `Iter`. This branch merged `main` and then replaced both with its own versions, so the alonzo.go diff against `main` is a rewrite with no behavioural difference -- and `Value` loses the early exit in the process. Those hunks should go.

What is *not* superseded, and is the reason to keep this PR open, is the `UtxoValidateExUnitsTooBigUtxo` change in `ledger/alonzo/rules.go` and `ledger/babbage/rules.go`. That is new and correct.

Both halves check out against cardano-ledger rather than by inference:

- `totExUnits tx = foldMap snd $ tx ^. witsTxL . rdmrsTxWitsL . unRedeemersL` -- the sum is over the `Map`, so a duplicated key contributes once.
- For PV < 9, `RedeemersRaw`'s decoder is `mapTraverseableDecoderA (decodeList decodeAnnElement) (RedeemersRaw . Map.fromList)`. `Map.fromList` keeps the last value, so last-wins is the decoder's behaviour, not just what the canonical block implies.

Nothing here can newly reject: collapsing can only lower the summed ex-units. One test gap inline.

ledger/alonzo/alonzo.go
```diff
@@ -761,21 +782,36 @@ func (r AlonzoRedeemers) Indexes(tag common.RedeemerTag) []uint {
 	return ret
 }
 
+// Value returns the redeemer for a key, taking the LAST entry when the list
```
wolf31o2: `main` already has this via #2200 (9c07b06), which walks the list backward and returns on the first match from the end. This version scans forward and overwrites, which is the same result with no early exit. Same for `Iter` above: `main` builds a `map[RedeemerKey]RedeemerValue` and sorts the keys; this rebuilds a last-index map and sorts the surviving entries. Identical output, since the keys are unique after collapsing either way.

Suggest rebasing and dropping both hunks so the diff is just the ex-units fix -- that is the part that is actually new, and it reviews much better on its own.

ledger/alonzo/rules.go
```diff
@@ -233,8 +233,18 @@ func UtxoValidateExUnitsTooBigUtxo(
...
+	// number of copies: on Preview, transaction 3ace3bc7f4c5… at slot 12925989
+	// carries the same (mint, 0) redeemer six times at 3322788/843448898, and
+	// summing all six gives 19936728 memory against a 14000000 cap, rejecting
+	// a transaction the network accepted (blinklabs-io/dingo#3875).
+	//
+	// The raw list is still the right thing for the script data hash, which has
+	// to reproduce the bytes as they were signed.
 	var totalSteps, totalMemory int64
-	for _, redeemer := range tmpTx.WitnessSet.WsRedeemers.Redeemers {
+	for _, redeemer := range tmpTx.WitnessSet.WsRedeemers.Iter() {
```
wolf31o2, line 247: This is the real fix and it matches `totExUnits`. But the accept-case for it -- "duplicate redeemer key counts once" -- is only added to `ledger/babbage/rules_test.go`. `ledger/alonzo/rules_test.go` gets the two `Index:` disambiguations for the overflow case and nothing else, so this line has no test of its own: revert just this hunk and the alonzo suite still passes.

Missing case, exactly: the same six-copies-of-one-`(mint, 0)`-redeemer subtest in `ledger/alonzo/rules_test.go`, sized so one copy is inside the cap and six are not.

Both new subtests also call the era rule function bare rather than resolving it from `UtxoValidationRules`, so nothing pins that the rule is still wired into the production slice. Consistent with the rest of the file, so not worth changing here, but it means the coverage is of the function and not of the rule.

ledger/alonzo/alonzo.go
chrisguiney: Agreed, and done. Rebased onto current `main` and dropped the commit entirely rather than the two hunks, since `main`'s #2200 (9c07b06) supersedes all of it.

`git diff origin/main -- ledger/alonzo/alonzo.go` is now empty. Your read of the difference was right: my `Value` set a variable and kept scanning, so it walked the whole list on every lookup; `main` returns on the first match from the end.

I also dropped my `ledger/alonzo/duplicate_redeemer_test.go` addition. It decoded the real six-copy fixture and summed `Iter()`, but `main`'s `TestAlonzoRedeemersDuplicateKeyLastWins` already asserts `Iter` yields a duplicated key exactly once, which determines that sum. It tested the decoder, not the rule, so the coverage it was standing in for belongs in `rules_test.go` — where it now is.

The PR is two commits and four files: `ledger/{alonzo,babbage}/rules.go` and their tests.

ledger/alonzo/rules.go
chrisguiney: Correct, and I checked it rather than taking it on faith. With the new subtest removed and only the `Iter()` change in `ledger/alonzo/rules.go` reverted, the alonzo suite passes:

```
ok  github.com/blinklabs-io/gouroboros/ledger/alonzo  0.288s
```

Added the subtest you specified, mirroring the Babbage one: six copies of a single `(mint, 0)` redeemer at 1,000,000 memory / 1,000 steps, so one copy is inside the 5,000,000 / 5,000 cap and six are not. It sits before the `ExUnits overflow` case, which mutates `testProtocolParams.MaxTxExUnits` to `MaxInt64`.

With it in place and the same hunk reverted:

```
--- FAIL: TestUtxoValidateExUnitsTooBigUtxo/duplicate_redeemer_key_counts_once
    rules_test.go:1283: six copies of one redeemer must count once, got:
    ExUnits too big: total 6000/6000000 steps/memory, maximum 5000/5000000 steps/memory
```

Only that subtest fails, and `ledger/babbage` still passes, so the failure is attributable to the alonzo hunk alone. Restored, both suites pass.

On the second point: taken, and left alone. Both subtests still call the era function directly, so nothing here pins that the rule is still in `UtxoValidationRules`. Worth a separate change that covers the whole file rather than two new subtests.

chrisguiney:
Rebased onto current `main` (`88145ad`) and force-pushed. Both of @wolf31o2's findings are addressed; the PR is now half its previous size.

**Dropped the duplicated last-wins work.** `main` already carries it via #2200 (`9c07b06`), and my branch had merged `main` and then replaced both implementations with its own. `git diff origin/main -- ledger/alonzo/alonzo.go` is now empty. I dropped my `duplicate_redeemer_test.go` addition with it: `main`'s `TestAlonzoRedeemersDuplicateKeyLastWins` already asserts `Iter` yields a duplicated key exactly once, which determines the sum mine was checking, and it tested the decoder rather than the rule.

**Gave the surviving fix a test in its own era.** `ledger/alonzo/rules_test.go` had only the two `Index:` disambiguations, so the alonzo rule change had no test — I confirmed that by reverting it and watching the suite pass. It now has the same `duplicate redeemer key counts once` accept case as Babbage, and with only the alonzo `rules.go` hunk reverted:

```
--- FAIL: TestUtxoValidateExUnitsTooBigUtxo/duplicate_redeemer_key_counts_once
    rules_test.go:1283: six copies of one redeemer must count once, got:
    ExUnits too big: total 6000/6000000 steps/memory, maximum 5000/5000000 steps/memory
```

`ledger/babbage` still passed during that revert, so the failure is attributable to the alonzo hunk alone.

**Diff is now four files:** `ledger/{alonzo,babbage}/rules.go` and their `rules_test.go`. `ledger/conway/rules.go` is no longer touched at all, so the two earlier cubic findings about `NewTxScriptView` error precedence are out of scope here for good — they are against code already on `main` and want their own issue.

Not changed, per review: both new subtests call the era rule function directly rather than resolving it from `UtxoValidationRules`, so they cover the function and not its wiring. That matches the rest of both files and is better handled file-wide than in two subtests.

`go build ./...`, `go test ./... -count=1` (39 packages, 0 failures) and `make lint` (0 issues) pass locally.

chrisguiney:
@coderabbitai review

@cubic-dev-ai review

Both bots skipped `637b33b`: CodeRabbit has auto-incremental reviews disabled, and cubic reports "Automatic AI review not started after branch rewrite". Requesting a pass over the rebased diff, which is four files.

ledger/babbage/rules_test.go
```diff
@@ -1336,6 +1336,36 @@ func TestUtxoValidateExUnitsTooBigUtxo(t *testing.T) {
...
+				t.Errorf(
+					"six copies of one redeemer must count once, got: %v", err,
+				)
+			}
+		},
+	)
+
 	// Ex-units too large
 	t.Run(
 		"ExUnits too large",
```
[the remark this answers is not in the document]
chrisguiney: Valid, and fixed in both eras. The old sizing only discriminated at the extreme:

```
per-copy 1_000_000 mem / 1_000 steps, cap 5_000_000 / 5_000
  1 copy  ->  1000000 / 1000   accept
  2 copies->  2000000 / 2000   accept   <- a broken collapse passes here
  ...
  5 copies->  5000000 / 5000   accept   <- and here
  6 copies->  6000000 / 6000   REJECT
```

Per-copy ExUnits are now 3_000_000 memory / 3_000 steps, exactly as you suggested, so one copy fits and two already exceed the cap:

```
  1 copy  ->  3000000 / 3000   accept
  2 copies->  6000000 / 6000   REJECT
```

Only a collapse to exactly one entry passes now.

Verified by mutation rather than by reading. With the `Iter()` change reverted in both `ledger/alonzo/rules.go` and `ledger/babbage/rules.go`:

- six copies, new sizing -> **FAIL** in both eras, `total 18000/18000000 steps/memory, maximum 5000/5000000`
- two copies, new sizing -> **FAIL** in both eras, `total 6000/6000000 steps/memory, maximum 5000/5000000`
- two copies, old sizing -> **passes** in both eras

That last line is your finding reproduced directly: the same partial over-count that the old numbers waved through is now caught.

`go test ./... -count=1` (39 packages) and `make lint` (0 issues) pass with the hunks restored.

chrisguiney:
Pushed `4840dd9`, which addresses cubic's finding on the two tests I had just added.

Cubic was right that the accept case was loose: at 1,000,000 memory per copy against a 5,000,000 cap, only a full six-way over-count tripped it — a broken collapse that counted 2, 3, 4 or 5 copies would have passed. Per-copy ExUnits are now 3,000,000 memory / 3,000 steps, so one copy fits and two already exceed the cap, and only a collapse to exactly one entry passes.

Confirmed by mutation in both eras, with the `Iter()` change reverted in `rules.go`:

| copies | per-copy | result |
| --- | --- | --- |
| 6 | 3,000,000 (new) | FAIL — `total 18000/18000000, maximum 5000/5000000` |
| 2 | 3,000,000 (new) | FAIL — `total 6000/6000000, maximum 5000/5000000` |
| 2 | 1,000,000 (old) | passes — the blind spot cubic identified |

`go build ./...`, `go test ./... -count=1` (39 packages, 0 failures) and `make lint` (0 issues) pass on the final tree.

@wolf31o2 both of your threads are answered above and the diff is still the same four files; the only change since your review is the test sizing.

chrisguiney:
@cubic-dev-ai review

Another force-push (`4840dd9`), so the automatic review will have been skipped again. This one carries only the test-sizing fix for your previous finding.

wolf31o2 (approved):

PR #2200, opened 2026-09-04
fix(ledger): resolve duplicate redeemers last-wins

wolf31o2:
Supersedes #2198 and addresses the duplicate-redeemer failure in blinklabs-io/dingo#3875.

Alonzo and Babbage encode redeemers as lists, while cardano-ledger resolves them through a map. This change:

- resolves duplicate keys to the last list entry;
- yields each resolved key once, preventing repeated script evaluation and execution-unit accounting;
- covers the Preview duplicate-key shape plus a distinct-key control.

Validated with a focused fail-before/pass-after regression, affected-package tests, a focused race test, and `go vet`.

---
## Summary by cubic
Fixes duplicate redeemer handling in Alonzo so replay no longer fails on transactions with repeated keys. Duplicate keys now resolve to the last list entry and are yielded once, matching cardano-ledger's map-based behavior. This resolves the Preview replay failure in blinklabs-io/dingo#3875.

- `Iter()` now collapses duplicate keys to their last value before yielding sorted keys.
- `Value()` walks redeemers backward so the last entry wins.
- Adds a regression test using the Preview transaction shape from #3875 plus a distinct-key control.

<sup>Written for commit 3bfdb4aebc482cfad592e2abc5c70bd15318ba1f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2200?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ledger/alonzo/alonzo.go
+++ b/ledger/alonzo/alonzo.go
@@ -720,31 +720,31 @@ func (r AlonzoRedeemers) MarshalCBOR() ([]byte, error) {
 	return cbor.Encode(r.Redeemers)
 }
 
+// Iter projects the list-form encoding into the map held by cardano-ledger.
+// Duplicate keys therefore resolve to their last value and are yielded once.
 func (r AlonzoRedeemers) Iter() iter.Seq2[common.RedeemerKey, common.RedeemerValue] {
 	return func(yield func(common.RedeemerKey, common.RedeemerValue) bool) {
-		// Sort redeemers
-		sorted := make([]AlonzoRedeemer, len(r.Redeemers))
-		copy(sorted, r.Redeemers)
-		slices.SortFunc(
-			sorted,
-			func(a, b AlonzoRedeemer) int {
-				return common.CompareRedeemerKeys(
-					common.RedeemerKey{Tag: a.Tag, Index: a.Index},
-					common.RedeemerKey{Tag: b.Tag, Index: b.Index},
-				)
-			},
+		byKey := make(
+			map[common.RedeemerKey]common.RedeemerValue,
+			len(r.Redeemers),
 		)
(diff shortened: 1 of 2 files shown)


arepala-uml (approved):
Verified against current head (3bfdb4a), merged cleanly onto current main (which includes #2196):

- `Iter()` and `Value()` both confirmed to resolve duplicate (tag, index) redeemer keys to the last list entry, matching cardano-ledger's map-based last-wins semantics.
- Regression-test discipline confirmed: reverted `alonzo.go` alone while keeping the new tests — both `TestAlonzoRedeemersDuplicateKeyLastWins` and `TestAlonzoRedeemersDuplicateKeyPreservesDistinctKeys` fail without the fix (resolves to first entry, double-yields the duplicate key) and pass with it.
- `go build ./...`, `go vet ./...`, `golangci-lint run ./ledger/...` clean; full `go test ./...` passes across all packages.
- Blast-radius/class check: Babbage embeds `AlonzoRedeemers` directly and Conway's `legacyRedeemers` field is also `AlonzoRedeemers`, so both inherit this fix automatically. Conway's native map format already rejects duplicate keys at decode time, and Dijkstra decodes redeemers into a native Go map (no list-duplication path). No gaps in coverage.
- Cubic reports no issues; no CodeRabbit inline findings.

No merge blockers.

Issue #2098, opened 2026-08-25
ledger: use exact reward math and bound metadata nesting


wolf31o2:
Two common ledger helpers use unbounded or inexact representations:

- Reward calculation uses floating-point arithmetic where the ledger calculation is rational.
- Metadata JSON parsing has no explicit nesting limit.

Acceptance criteria:

- Use exact rational arithmetic and compare boundary results with reference vectors.
- Enforce a documented metadata nesting limit before descending recursively.
- Preserve valid existing reward and metadata behavior.
- Add rounding, large-denominator, nesting-boundary, and over-limit tests.

chrisguiney:
Fixed by #2201 (exact rational reward arithmetic, released in v0.202.10) and #2203 (MetadataJSONMaxNestingDepth enforced before descending, released in v0.204.1).

PR #2201, opened 2026-09-04
fix(ledger): compute pool rewards with exact rational arithmetic

chrisguiney:
Addresses the reward-calculation half of #2098. Metadata parsing is covered by #2203.

Changes:

- Replace floating-point pool performance, margin, and share arithmetic with exact big.Rat values.
- Compute delegator payouts with exact integer division.
- Apportion pool rewards deterministically by fractional remainder and pool ID, preserving the pot total.
- Isolate saturation-threshold values per calculation and enforce the allocation invariant defensively.

Validation:

- make test (race-enabled root and examples)
- go build ./...
- go vet ./...
- go test -count=1 ./internal/test/conformance/...
- go test -race -count=1 ./ledger/common
- git diff --check and gofmt -d

make lint was unavailable because golangci-lint is not installed in the validation environment.

### DIFF
--- a/ledger/common/rewards.go
+++ b/ledger/common/rewards.go
@@ -224,13 +224,14 @@ func calculateEta(totalBlocksInEpoch uint32, params RewardParameters) (*big.Rat,
 	return eta, nil
 }
 
-// poolSaturationThreshold (z0) is the relative stake at which a pool counts as
-// fully saturated. Held as an exact rational so the reward split does not
-// depend on the binary representation of 0.05. Read-only: never pass it as a
-// big.Rat receiver.
 // TODO(enhancement): Extract to a parameter for consistency with
 // CalculatePoolSaturation
-var poolSaturationThreshold = big.NewRat(1, 20)
+// poolSaturationThreshold returns (z0), the relative stake at which a pool
+// counts as fully saturated. Constructing a fresh value avoids sharing a
+// mutable big.Rat between concurrent reward calculations.
+func poolSaturationThreshold() *big.Rat {
+	return big.NewRat(1, 20)
+}
 
 // ratFromUint64 converts a lovelace amount to an exact rational
 func ratFromUint64(value uint64) *big.Rat {
@@ -431,6 +432,7 @@ func apportionRewardPot(
 	remainders := make(map[PoolKeyHash]*big.Rat, len(poolIDs))
 
 	potRat := ratFromUint64(pot)
+	potInt := new(big.Int).SetUint64(pot)
 	assigned := new(big.Int)
 	for _, poolID := range poolIDs {
 		// exact = pot * share / totalShare
@@ -440,8 +442,11 @@ func apportionRewardPot(
 			exact = new(big.Rat)
 		}
 		floor := new(big.Int).Quo(exact.Num(), exact.Denom())
-		if !floor.IsUint64() {
-			floor = new(big.Int).SetUint64(pot)
+		remaining := new(big.Int).Sub(potInt, assigned)
+		if remaining.Sign() <= 0 {
+			floor.SetInt64(0)
+		} else if floor.Cmp(remaining) > 0 {
+			floor.Set(remaining)
 		}
 		amounts[poolID] = floor.Uint64()
 		assigned.Add(assigned, floor)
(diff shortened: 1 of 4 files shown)


chrisguiney:
@coderabbitai full review

ledger/common/rewards.go
```diff
@@ -387,51 +416,108 @@ func CalculateRewards(
...
+	assigned := new(big.Int)
+	for _, poolID := range poolIDs {
+		// exact = pot * share / totalShare
+		exact := new(big.Rat).Quo(poolShares[poolID], totalShare)
+		exact.Mul(exact, potRat)
+		if exact.Sign() < 0 {
+			exact = new(big.Rat)
+		}
+		floor := new(big.Int).Quo(exact.Num(), exact.Denom())
+		if !floor.IsUint64() {
```
wolf31o2: This branch is unreachable and, if it were reached, would break the invariant the next step depends on. `exact = pot * share / totalShare` with every share non-negative (`calculatePoolShare` clamps margin to [0,1], `a0` to >= 0, and the denominator to > 0), so `exact` is in `[0, pot]` and the floor always fits a uint64. If it somehow did not, assigning the whole `pot` to one pool makes `assigned > pot`, `leftover` negative, and the amounts sum to more than the pot -- silently, because the early return swallows it.

Either drop it, or clamp to the still-unassigned remainder so the invariant holds by construction rather than by argument.

ledger/common/rewards.go
```diff
@@ -222,41 +224,88 @@ func calculateEta(totalBlocksInEpoch uint32, params RewardParameters) (*big.Rat,
 	return eta, nil
 }
 
+// poolSaturationThreshold (z0) is the relative stake at which a pool counts as
+// fully saturated. Held as an exact rational so the reward split does not
+// depend on the binary representation of 0.05. Read-only: never pass it as a
+// big.Rat receiver.
+// TODO(enhancement): Extract to a parameter for consistency with
+// CalculatePoolSaturation
+var poolSaturationThreshold = big.NewRat(1, 20)
```
wolf31o2: Safe today -- `big.Rat.Quo` does not mutate its operands -- and the comment says so. But a package-level mutable `*big.Rat` is one careless `.Set(` from a data race across concurrent `CalculateRewards` calls, and the compiler will not warn. A `func poolSaturationThreshold() *big.Rat { return big.NewRat(1, 20) }` costs nothing here and removes the hazard along with the comment that warns about it.

ledger/common/rewards_exact_test.go
```diff
@@ -0,0 +1,255 @@
...
+			first:  true,
+			second: true,
+		}},
+	)
+
+	assert.Equal(t, 2*firstStake, rewards.DelegatorRewards[first])
+	assert.Equal(t, 2*secondStake, rewards.DelegatorRewards[second])
+	assert.Equal(t, poolCost, rewards.OperatorRewards)
+}
+
```
wolf31o2, line 94: "Each fails on `main`" is not quite the fails-without-the-fix property for most of these. `calculatePoolPerformance` and `marginRat` changed return type, so `TestCalculatePoolPerformanceDistinguishesStakeAbovePow53`, `TestMarginRatIsBoundedAndExact` and `TestCalculateRewardsIsDeterministic` do not compile against `main` -- they cannot be run against the unfixed code at all.

The two that genuinely discriminate are `TestDistributePoolRewardsExactDelegatorShare` and `TestDistributePoolRewardsExactLargeDenominator`: `distributePoolRewards` keeps its signature, so reverting only its body reproduces the one-lovelace shortfall. Worth saying which is which in the description, so a later reader does not treat the type-change tests as regression proofs.

wolf31o2 (commented):
Nothing blocking. The representation change is faithful: every floor is `big.Int.Quo`/`Div` on a non-negative value, so it matches the old `uint64(float64)` truncation; `apportionRewardPot`'s largest-remainder split with a pool-ID tiebreak sums to the pot exactly and is order-independent; and the surviving `range delegatorStake` is safe because `assigned` is exact integer addition.

One framing point worth stating in the description: this formula is not cardano-ledger's. `rewardOnePool`/`maxPool` compute something else, and the leftover lovelace here go to the operator where the ledger returns them to the pot. Making the arithmetic exact does not make it agree with the node, and a reader of "exact rational arithmetic" could take it that way. The change is still right.

`CalculateRewards` is not on the transaction validation path in this module, so this cannot newly reject anything. Three inline notes. Branch is on `fab2538`, four commits behind `main`.

wolf31o2 (approved):

PR #2202, opened 2026-09-04
fix(ledger): order withdrawals by credential (#2187)

### LINKED ISSUE #2187: Withdrawals ordered by stake-address bytes, not by Credential
up`ledger/common/script/context.go:644` `withdrawalsInfo` and `purpose.go` `BuildScriptPurpose` order withdrawals by full stake-address bytes, so key-hash credentials (`0xe0`/`0xe1`) sort before script-hash (`0xf0`/`0xf1`).

cardano-ledger orders `RewardAccount` by `Credential`, declared `ScriptHashObj | KeyHashObj` (`Cardano/Ledger/Credential.hs`), so script credentials sort **first**.

For a transaction withdrawing from both key-hash and script-hash stake credentials, both the `txInfoWdrl` order and the redeemer-index to credential mapping diverge from the node. A Plutus validator reading withdrawals positionally, or resolving its own purpose by index, sees the wrong entry.

Found while diagnosing dingo #3780; not triggered by that block (all four credentials are script-hash). Verified against `Credential.hs` on master; not reproduced against a real transaction.

Same family as #2177 (`SortInputs` ordering).

Acceptance criteria: withdrawal ordering matches `Ord (Credential kr)` with script credentials first; a test covers a transaction withdrawing from both credential types and pins both the `txInfoWdrl` order and the redeemer index mapping.

wolf31o2:
Fixes #2187

Cardano ledger orders withdrawal reward accounts by `(network, Credential)`,
with script credentials before key credentials. Address-byte ordering reverses
that credential-type distinction.

- Share credential-based withdrawal ordering across TxInfo and redeemer-purpose consumers.
- Apply the ordering to Dijkstra V4 withdrawal TxInfo and reward redeemer indices.
- Add mixed key-hash/script-hash regression coverage for Plutus V1–V3 and Dijkstra V4.

Validation:

- `GOWORK=off go test -race -count=1 ./...` — pass
- `GOWORK=off go test -race -count=1 -v ./internal/test/conformance/...` — 315/315 vectors passed
- `GOWORK=off go vet ./...` — pass

## Summary by CodeRabbit

- **Bug Fixes**
  - Ensured withdrawal entries consistently follow credential order across transaction information and Plutus script versions.
  - Preserved deterministic ordering for reward withdrawals, including cases with both script and key credentials.
  - Standardized handling of nil and non-nil withdrawal addresses.

- **Tests**
  - Added coverage validating withdrawal ordering across Plutus V1–V4 and reward purpose indexing.

### DIFF
--- a/ledger/common/script/context.go
+++ b/ledger/common/script/context.go
@@ -709,7 +709,7 @@ func withdrawalsInfo(
 			},
 		)
 	}
-	sorted := sortWithdrawalAddresses(withdrawals)
+	sorted := SortWithdrawalAddresses(withdrawals)
 	ret = ret[:0]
 	for _, addr := range sorted {
 		ret = append(ret, KeyValuePair[*lcommon.Address, *big.Int]{
--- a/ledger/common/script/context_test.go
+++ b/ledger/common/script/context_test.go
@@ -15,6 +15,7 @@
 package script_test
 
 import (
+	"bytes"
 	"encoding/hex"
 	"errors"
 	"fmt"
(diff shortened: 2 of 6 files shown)


arepala-uml (approved):
Reviewed at head 5664d3c4. This fixes two independently-broken withdrawal-ordering implementations in Dijkstra:

- `dijkstraWithdrawalsV4` sorted by raw `CredType` ascending (key before script -- backwards from cardano-ledger's derived Ord, which places ScriptHashObj before KeyHashObj).
- `dijkstraRequiredScriptPurposes` sorted by raw address bytes, not credential at all -- since a reward header byte is 0xE_ for key hash and 0xF_ for script hash, this also inverted the real order.

Both now route through the shared `script.SortWithdrawalAddresses` (exported from the already-correct implementation previously used only for V1-V3 TxInfo).

Verified independently:
- Cross-checked the new shared function against the pre-existing reference implementation in `ledger/common/rules.go` (untouched by this PR), which already cites the actual cardano-ledger Haskell source paths for the Ord instances (Network Testnet-before-Mainnet, Credential ScriptHashObj-before-KeyHashObj). The logic matches exactly.
- Confirmed `dijkstraRequiredScriptPurposes` still indexes `RedeemerTagReward` off position in the full sorted list (only skipping non-script entries via `continue`), matching what `BuildScriptPurpose`'s reward case expects when indexing into the same sorted list.
- Searched for other ad-hoc withdrawal-sort implementations across the repo -- none remain; every consumer now funnels through the one shared function.
- The new regression tests deliberately give the script and key addresses identical hash bytes, isolating the credential-type dimension specifically -- exactly the axis both bugs got wrong.

CI and cubic are both clean. No merge blockers. Approving.

PR #2203, opened 2026-09-04
fix(ledger): bound metadata JSON nesting depth

chrisguiney:
Addresses the metadata-parsing half of #2098. Exact reward arithmetic is covered by #2201.

Metadata JSON parsing recursed once per nesting level with no bound, and the depth it accepted was not tied to what this repository's CBOR decoder accepts.

- Reject over-deep metadata JSON on entering a container, before descending, in both the object and array readers.
- Bound the parser by `cbor.MaxNestedLevels`, now exported so the JSON and CBOR limits cannot drift apart. The mandatory top-level object counts toward the bound, so the deepest document the parser accepts is exactly the deepest one `cbor.Decode` accepts.
- Pin that boundary by encoding every accepted document and decoding it back through both `cbor.Decode` and `DecodeMetadatumRaw`, over both container readers, alongside the first rejected depth and a far-over-limit payload.


chrisguiney:
**cubic P3 (`metadata_json.go:78`, duplicated nesting literal): valid, fixed in 91a8d03.**

Confirmed against the checkout: `cbor` held one literal (`cborMaxNestedLevels`, `cbor/decode.go:43`) used from both `decode.go` and `value.go`, so the only genuine duplication was the new `256` in `ledger/common`. A doc comment asserted the two matched with nothing enforcing it.

- Exported it as `cbor.MaxNestedLevels`, moving the fxamacker-default rationale onto the constant now that it is shared.
- `MetadataJSONMaxNestingDepth = cbor.MaxNestedLevels`, so the documented match is a compile-time fact rather than a comment.

No other repeated nesting bound exists in the tree (`grep` for `cborMaxNestedLevels`/`MaxNestedLevels` now resolves to the one constant).

Re-ran after the change: `go test -race -count=1 ./...` (39 packages), `golangci-lint run ./...` 0 issues, `nilaway ./ledger/common/ ./cbor/` clean, `gofmt -s` and the 80-column limit clean.

Note for the record: CodeRabbit has not reviewed this PR. Its check reports success, but the comment body is a "Review limit reached" rate-limit notice.

chrisguiney:
@coderabbitai full review

ledger/common/metadata_json.go
```diff
@@ -66,6 +66,28 @@ func NewShelleyAuxiliaryData(
...
+// The bound is the CBOR decoder's limit: metadata that cannot round trip
+// through CBOR is not usable as transaction metadata anyway. It is enforced
+// before descending because the readers below recurse once per level, and
+// exhausting the goroutine stack is a fatal error that recover cannot catch.
+const MetadataJSONMaxNestingDepth = cbor.MaxNestedLevels
+
+// checkMetadataJSONNestingDepth reports whether a container opened at depth
+// may be descended into.
+func checkMetadataJSONNestingDepth(depth int) error {
+	if depth >= MetadataJSONMaxNestingDepth {
```
wolf31o2: The two limits do not agree at the boundary. This is `depth >= MaxNestedLevels` with the top-level object at depth 0, so 256 containers may be open. `cbor.Value.unmarshalCBOR` is `depth > MaxNestedLevels` from a depth-0 top level, so it allows 257. A document with exactly 257 nested containers therefore round trips through `cbor.Value` but is rejected here -- which makes the stated rationale ("metadata that cannot round trip through CBOR is not usable") off by one.

Harmless in effect, since the stricter side is the JSON parser. But either match the CBOR convention or say in the doc comment why the JSON side is one container tighter.

While here: 256 has no counterpart in cardano-ledger -- it is gouroboros's own decoder setting. Worth a word in the doc comment so a later reader does not take `MetadataJSONMaxNestingDepth` for a protocol constant.

wolf31o2 (commented):
Nothing blocking. This bounds a cardano-cli JSON parser, not the block decode path, so a too-strict limit cannot reject a chain transaction -- the usual "a stricter bound rejects valid transactions" concern does not apply here. The fix checks before descending, which is the only thing that works against a stack overflow `recover` cannot catch, and the tests cover last-accepted, first-rejected, far-beyond and both schemas.

One off-by-one against the limit the change cites as its justification, inline. Branch is on `fab2538`, four commits behind `main`.

ledger/common/metadata_json.go
```diff
@@ -66,6 +66,27 @@ func NewShelleyAuxiliaryData(
...
+// container at that depth is allowed; deeper containers are rejected only
+// after they exceed the same bound used by CBOR. This is enforced before
+// descending because the readers below recurse once per level, and exhausting
+// the goroutine stack is a fatal error that recover cannot catch.
+const MetadataJSONMaxNestingDepth = cbor.MaxNestedLevels
+
+// checkMetadataJSONNestingDepth reports whether a container opened at depth
+// may be descended into.
+func checkMetadataJSONNestingDepth(depth int) error {
+	if depth > MetadataJSONMaxNestingDepth {
```
wolf31o2, line 81: `depth > MetadataJSONMaxNestingDepth` accepts 257 containers (top-level object at depth 0 plus 256 nested), one more than a decoder configured with `MaxNestedLevels: 256` accepts. The doc comment's parity claim does not hold in effect.

Measured on this head: `{"1":` + 256 `[` + 256 `]` + `}` parses successfully, and the encoded result does not decode -- `cbor.Decode` into `ShelleyAuxiliaryData` returns `cbor: exceeded max nested level 256`, and `DecodeMetadatumRaw` returns `failed to decode CBOR map in metadatum`. At 255 arrays both succeed.

`cbor.Value.unmarshalCBOR` also permits 257 levels, but it hands fxamacker one item at a time, so the full nesting never reaches the `MaxNestedLevels` gate. Matching its `>` convention therefore does not match `getDecMode`. `>=`, as in 91a8d03, is the boundary that makes the accepted maximum decodable.

ledger/common/metadata_json_nesting_test.go
```diff
@@ -0,0 +1,86 @@
...
+		`{"1":` + strings.Repeat("[", arrays) +
+			strings.Repeat("]", arrays) + `}`,
+	)
+}
+
+func TestParseMetadataJSONNoSchemaAcceptsMaximumNesting(t *testing.T) {
+	// The top-level object occupies depth 0, so the maximum number of nested
+	// arrays is the configured depth bound.
+	metadata, err := ParseCardanoCLIMetadataJSONNoSchema(
+		nestedNoSchemaJSON(MetadataJSONMaxNestingDepth),
```
wolf31o2, line 38: This locks in the off-by-one: `MetadataJSONMaxNestingDepth` arrays under the top-level object encodes to 257 CBOR levels, which this repository's decoder rejects. Under a `>=` boundary the last accepted input is `MetadataJSONMaxNestingDepth - 1` arrays, which round trips through `cbor.Encode`/`cbor.Decode` and `DecodeMetadatumRaw`.

Worth adding to this test: encode the accepted result and decode it back, so the accept boundary is pinned to what the decoder accepts rather than to the constant alone.

wolf31o2 (changes requested):
The bound is on both container readers and runs before descent, which is the only placement that works against a stack overflow `recover` cannot catch. The reject tests are real regression tests: with the check disabled locally, all three reject cases return no error.

One boundary defect inline. The maximum accepted depth is one container deeper than the decoder built from the same constant accepts, so the deepest document this parser accepts produces metadata this package cannot decode.

Validated on b32aa8c: `go build ./...`, `go vet ./ledger/common/... ./cbor/...`, `go test -count=1 ./ledger/common/... ./cbor/...` all pass. Not run: devnet, conformance beyond the packages above, `golangci-lint`.

ledger/common/metadata_json.go
arepala-uml: This accepts one container deeper than the real CBOR decode path (`cbor.Decode` via `getDecMode`'s `MaxNestedLevels: 256`) can decode -- confirmed by encoding and decoding metadata parsed at exactly `MetadataJSONMaxNestingDepth`, which fails with `cbor: exceeded max nested level 256`. At `MetadataJSONMaxNestingDepth - 1` the round trip succeeds. `cbor.Value.unmarshalCBOR`'s `>` convention isn't the right thing to match here -- that path decodes item-by-item and never exercises the real `MaxNestedLevels` gate, so its literal operator doesn't reflect what `cbor.Decode` actually accepts.

```suggestion
	if depth >= MetadataJSONMaxNestingDepth {
```

ledger/common/metadata_json_nesting_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+package common
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
```
arepala-uml: Needed for the round-trip regression test suggested below.

```suggestion
import (
	"strings"
	"testing"

	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"

	"github.com/blinklabs-io/gouroboros/cbor"
)
```

ledger/common/metadata_json_nesting_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+func TestParseMetadataJSONDetailedSchemaAcceptsModestNesting(t *testing.T) {
+	const levels = 8
+	payload := `{"1":` + strings.Repeat(`{"list":[`, levels) +
+		`{"int":1}` + strings.Repeat(`]}`, levels) + `}`
+
+	metadata, err := ParseCardanoCLIMetadataJSONDetailedSchema([]byte(payload))
+	require.NoError(t, err)
+	require.NotNil(t, metadata)
+}
```
arepala-uml: Suggested regression test: pins the accepted boundary to what the decoder actually accepts (per your comment above), rather than to the constant alone. Fails on this head (256 does not round trip); should pass once the boundary check is corrected.

```suggestion
}

func TestParseMetadataJSONNoSchemaAcceptedMaximumRoundTripsThroughCBOR(t *testing.T) {
	metadata, err := ParseCardanoCLIMetadataJSONNoSchema(
		nestedNoSchemaJSON(MetadataJSONMaxNestingDepth),
	)
	require.NoError(t, err)

	aux := NewShelleyAuxiliaryData(metadata)
	encoded, err := cbor.Encode(aux)
	require.NoError(t, err)

	var decodedAux ShelleyAuxiliaryData
	_, err = cbor.Decode(encoded, &decodedAux)
	require.NoError(
		t,
		err,
		"the documented maximum accepted nesting depth must round trip through CBOR",
	)
}
```

arepala-uml (dismissed):
Retracted — posted in error; this PR was not the one asked for review.

ledger/common/metadata_json.go
chrisguiney: Fixed in 7c28c5f: `depth >= MetadataJSONMaxNestingDepth`, on both container readers.

Reproduced your measurement on b32aa8c before moving the boundary. 256 arrays under the top-level object parsed, then `cbor.Decode` into `ShelleyAuxiliaryData` returned `cbor: exceeded max nested level 256` and `DecodeMetadatumRaw` returned `failed to decode CBOR map in metadatum`; 255 arrays round tripped through both. At 7c28c5f the parser rejects 256 arrays, so the deepest accepted document is 255 arrays plus the top-level object — 256 containers, exactly what `getDecMode` accepts.

One correction to the reasoning, not to the conclusion. `cbor.Value.unmarshalCBOR`'s own guard does permit 257 levels, but that is not reachable: fxamacker runs its well-formedness pass over the whole item before calling `Value.UnmarshalCBOR`, so the full nesting does reach the `MaxNestedLevels` gate. Measured with nested `0x81` arrays:

| levels | `Decode(&any)` | `Decode(&Value)` | `v.unmarshalCBOR(data, false, 0)` |
| --- | --- | --- | --- |
| 256 | ok | ok | ok |
| 257 | exceeded max nested level 256 | exceeded max nested level 256 | ok |
| 258 | exceeded max nested level 256 | exceeded max nested level 256 | exceeded maximum CBOR nesting depth: 256 |

So `cbor.Value` and `getDecMode` already agree at the observable boundary, and `value.go`'s `>` is a latent off-by-one in a guard the fxamacker check shadows. Left alone as out of scope. I also dropped the `cbor.Value` comparison from the doc comment rather than restate it, since the mechanism is not what the code does.

ledger/common/metadata_json_nesting_test.go
chrisguiney: Done in 7c28c5f. The accept test now uses `MetadataJSONMaxNestingDepth - 1` arrays, and the reject test's `first invalid` case is `MetadataJSONMaxNestingDepth`.

Added `TestParseMetadataJSONNoSchemaAcceptBoundaryIsDecodable`, which sweeps 254..257 arrays and requires every document the parser accepts to encode through `NewShelleyAuxiliaryData`/`cbor.Encode` and decode back through both `cbor.Decode` into `ShelleyAuxiliaryData` and `DecodeMetadatumRaw`, then pins the deepest accepted count at `MetadataJSONMaxNestingDepth - 1`. That makes it a property of the decoder rather than of the constant, so it fails if either side moves.

Fail-before, with the guard reverted to `>` in place and the tests unchanged:

```
--- FAIL: TestParseMetadataJSONNoSchemaAcceptBoundaryIsDecodable
    Received unexpected error: cbor: exceeded max nested level 256
    Messages: 256 arrays: parser accepted metadata that cbor.Decode rejects
--- FAIL: TestParseMetadataJSONNoSchemaRejectsExcessiveNesting/first_invalid
    An error is expected but got nil.
```

Restored, both pass.

ledger/common/metadata_json.go
chrisguiney: Both halves settled at 7c28c5f.

The boundary went back to `>=`, where this thread started. The measurement on the newer thread shows `>=` is what matches `getDecMode`, and that `cbor.Value`'s extra level is unreachable through `Decode` — so there was no convention on that side to match after all, and the JSON parser is not one container tighter than CBOR; it is exact.

The doc comment now says the bound is `cbor.MaxNestedLevels`, gouroboros's own CBOR decoder setting, with no counterpart in cardano-ledger and not a protocol constant.

Class audit while here: `cbor/diagnostic.go:92` is the only other guard of this shape, `depth > maxDiagnosticNestedLevels` against its own 256. Measured, `Diagnose` accepts 257 levels and rejects 258, so it is one level more permissive than `getDecMode`, and `diagnostic_api.go:79`'s description of the cap as 256 is correspondingly off by one. Harmless in effect — it is an inspection reader over raw bytes rather than a producer, so describing input the strict decoder rejects is the right direction — and unrelated to this change, so I left it. The display guards (`debug.go:28`, `diagnostic.go:627/848/962`) take a caller-supplied `MaxDepth` and are not bounds against `MaxNestedLevels`. No other member of the class.

ledger/common/metadata_json_nesting_test.go
chrisguiney: Extended in e08870f: the accept-boundary round trip and the reject cases now run over both container readers.

The object reader had no boundary coverage. With its depth check deleted and the array reader's left in place, `go test ./ledger/common/` stayed green, while 256 nested objects parsed and then failed `cbor.Decode` with `cbor: exceeded max nested level 256` and `DecodeMetadatumRaw` with `failed to decode CBOR map in metadatum` — the same accept-boundary defect this thread pinned for arrays.

Both mutations now fail. Deleting only the object guard fails the `object` subtests of `TestParseMetadataJSONNoSchemaAcceptBoundaryIsDecodable` and `TestParseMetadataJSONNoSchemaRejectsExcessiveNesting`; restoring `>` in `checkMetadataJSONNestingDepth` fails the `array` and `object` subtests of both.

chrisguiney:
The boundary defect was fixed in 7c28c5f, before this round. Confirmed by reverting `>=` to `>` in `checkMetadataJSONNestingDepth` in place: `TestParseMetadataJSONNoSchemaAcceptBoundaryIsDecodable` then fails with `cbor: exceeded max nested level 256` at 256 arrays, and the `first invalid` reject case stops erroring. At head the deepest accepted document is `MetadataJSONMaxNestingDepth - 1` arrays under the top-level object — 256 containers open, which is what `getDecMode` accepts.

e08870f closes a gap in the same class. Both container readers carry the depth check, but every nesting test drove the array reader. Deleting the check from `readMetadataJSONObject` alone left `./ledger/common` green, while 256 nested objects parsed and then failed `cbor.Decode` with `cbor: exceeded max nested level 256`. The accept-boundary round trip and the reject cases now run over both readers, so that mutation fails the `object` subtests and reverting `>=` fails both.

Covering the checks this review recorded as not run, at e08870f:

- `golangci-lint run ./ledger/common/... ./cbor/...` — 0 issues
- `go test -race -count=1 ./ledger/common/... ./cbor/...` — pass
- `go test -count=1 ./internal/test/conformance/...` — pass
- `go build ./...` and `go vet ./...`, plus build and vet in each `examples/*` module — pass
- merged with current `main`: conflict-free, builds and tests clean

`cbor.MaxNestedLevels` and `MetadataJSONMaxNestingDepth` are new exported constants and the decoder's configured value is unchanged at 256, so the surface change is additive.

chrisguiney:
d6cc772 is a conflict-free `origin/main` merge, bringing the branch level with `main`. It does not touch `ledger/common/metadata_json.go` or `ledger/common/metadata_json_nesting_test.go`, so the reviewed code is unchanged from e08870f.

Re-verified on the merged tree: `go test -race -count=1 ./ledger/common/... ./cbor/...` and `golangci-lint run ./ledger/common/... ./cbor/...` both pass, and both fail-before mutations still fail — deleting the depth check from `readMetadataJSONObject` fails the `object` subtests, and restoring `>` in `checkMetadataJSONNestingDepth` fails the `array` and `object` subtests.

wolf31o2 (approved):
Accept

PR #2205, opened 2026-09-04
fix(ledger): apply the collateral rules only to phase-2 transactions

chrisguiney:
Reported as blinklabs-io/dingo#3896.

Collateral pays for phase-2 script execution that fails. A transaction that runs
no phase-2 scripts has nothing for it to cover, but the collateral witness rules
were applied to any transaction that *declared* collateral — so declaring
collateral it did not need made an otherwise valid transaction invalid.

## How it surfaced

A from-genesis Preview replay wedged at slot 15148509 (epoch 175):

```
tx 9ce59ee0dc6abee0de759c33b4b0f59dedda2f2eb56b89b6978b5d098100ac44
  validation failure at slot 15148509:
  transaction: collateral input must be key-locked
```

114 occurrences, node no longer following the chain.

The transaction:

```
body     inputs, outputs, fee 177117, collateral [811c9029...674f#0]
witness  2 vkey witnesses
         1 native script
         no plutus_v1/v2/v3, no redeemers
```

Its collateral input really is script-locked — that part of the check is right:

```
addr_test1wrpe58z89f3kwtq2cslsqvvw9hzdep2jy2ykx9kty8xnamqycrwy6
header 0x70 -> type 7, enterprise(script)
```

Koios confirms the transaction is on chain at block height 670981 with no
collateral consumed, so it was phase-2 valid and the collateral was never
touched.

## The change

`ValidateCollateralVKeyWitnesses` previously returned early only when no
collateral was declared. It now also returns when the transaction runs no
phase-2 scripts.

**The condition is the presence of redeemers, not of Plutus scripts in the
witness set.** A script supplied by a reference input is not in the witness set,
so gating on that would skip the check for exactly the transactions that most
need it. Every phase-2 execution carries a redeemer wherever its script came
from.

## What I have not confirmed

**I have not verified this against cardano-ledger**, and that is worth a
reviewer's attention before merge.

What is certain is that the current behaviour is wrong: the block is canonical
and a node applying this rule cannot follow the chain. What I am inferring is
the precise condition the ledger uses — whether it gates on phase-2 scripts
being present, on a non-empty redeemer set, or on something else again.

Redeemers seemed the most defensible reading, and it is strictly narrower than
removing the check: a transaction that does run phase-2 scripts is still held to
the rule, which the second test asserts. But if the reference implementation
gates differently, this should follow it rather than the reading that happens to
make this block pass.

## Tests

`TestCollateralKeyLockedOnlyForPhase2` uses the real transaction's shape and the
real collateral address, and covers both directions: script-locked collateral is
accepted without phase-2 scripts, and still rejected with them, so the guard
cannot become a way to skip the rule. Without the change the first case fails
with `collateral input must be key-locked`.

---
## Summary by cubic
Applies the collateral witness rules only to transactions that run phase-2 scripts, fixing a node that stopped following the chain when a transaction declared unnecessary collateral.

The fix extends beyond the original key-locked rule: the `total_collateral` balance check is also gated, and redeemers found in sub-transaction witness sets count, so Dijkstra transactions cannot bypass the rules.

**Details**
- Gates the rules on the presence of redeemers rather than Plutus scripts in the witness set, so reference-input scripts are still covered.
- Counts redeemers in sub-transaction witness sets too, so Dijkstra transactions can't bypass the rule.
- Adds a regression test covering both accepted and rejected cases, resolving the rules from the production slice so the tests exercise what the node actually runs.

<sup>Written for commit aac12207f4068b5586f7203d0cae25694c35704b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2205?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ledger/babbage/collateral_phase2_test.go
+++ b/ledger/babbage/collateral_phase2_test.go
@@ -1,6 +1,7 @@
 package babbage_test
 
 import (
+	"reflect"
 	"testing"
 
 	"github.com/blinklabs-io/gouroboros/cbor"
@@ -12,6 +13,103 @@ import (
 	mockledger "github.com/blinklabs-io/ouroboros-mock/ledger"
 )
 
+// collateralFixtureTxId is the input the collateral fixtures spend as collateral.
+const collateralFixtureTxId = "811c9029fc79e6b552f54d857bf0db807c9f0d8b23dc9e1e37f382377018674f"
+
+// productionRule returns the entry of babbage.UtxoValidationRules whose
+// function identity matches want.
+//
+// The rules below must be exercised as they are wired, not as free functions:
+// a rule dropped from the production slice would still pass a test that called
+// it directly, and the point of these cases is what the node actually runs.
+func productionRule(
+	t *testing.T,
+	name string,
+	want common.UtxoValidationRuleFunc,
+) common.UtxoValidationRuleFunc {
+	t.Helper()
+	wantId := reflect.ValueOf(want).Pointer()
+	for _, rule := range babbage.UtxoValidationRules {
+		if reflect.ValueOf(rule).Pointer() == wantId {
+			return rule
+		}
+	}
+	t.Fatalf("%s is not wired into babbage.UtxoValidationRules", name)
+	return nil
+}
+
+// collateralFixtureLedgerState holds the single collateral UTxO the fixtures
+// spend: 100 ADA at an enterprise address with a script payment credential,
+// which is the shape that wedged the node.
+func collateralFixtureLedgerState(t *testing.T) common.LedgerState {
+	t.Helper()
+	// The real collateral address: enterprise with a script payment credential.
+	scriptAddr, err := common.NewAddress(
+		"addr_test1wrpe58z89f3kwtq2cslsqvvw9hzdep2jy2ykx9kty8xnamqycrwy6",
+	)
+	if err != nil {
+		t.Fatalf("fixture address: %v", err)
+	}
+	return mockledger.NewLedgerStateBuilder().WithUtxos([]common.Utxo{
+		{
+			Id: shelley.NewShelleyTransactionInput(collateralFixtureTxId, 0),
+			Output: babbage.BabbageTransactionOutput{
+				OutputAddress: scriptAddr,
+				OutputAmount: mary.MaryTransactionOutputValue{
+					Amount: 100_000_000,
+				},
+			},
+		},
+	}).Build()
+}
+
+// spendRedeemers is a witness-set redeemer map standing for phase-2 execution.
+func spendRedeemers() alonzo.AlonzoRedeemers {
+	return alonzo.AlonzoRedeemers{
+		Redeemers: []alonzo.AlonzoRedeemer{
+			{Tag: common.RedeemerTagSpend, Index: 0},
+		},
+	}
+}
+
+// collateralFixtureTx builds a transaction declaring the fixture collateral
+// input, optionally with a redeemer and a total_collateral field.
+func collateralFixtureTx(
+	withRedeemer bool,
+	totalCollateral uint64,
+) *babbage.BabbageTransaction {
+	wits := babbage.BabbageTransactionWitnessSet{}
+	wits.VkeyWitnesses = []common.VkeyWitness{
+		{Vkey: []byte{0x01}, Signature: []byte{0x02}},
+	}
+	if withRedeemer {
+		wits.WsRedeemers = spendRedeemers()
+	}
+	return &babbage.BabbageTransaction{
+		Body: babbage.BabbageTransactionBody{
+			TxTotalCollateral: totalCollateral,
+			TxCollateral: cbor.NewSetType(
+				[]shelley.ShelleyTransactionInput{
+					shelley.NewShelleyTransactionInput(collateralFixtureTxId, 0),
+				},
+				false,
+			),
+		},
+		WitnessSet: wits,
+	}
+}
+
+// withSubTxRedeemers wraps tx so that it also exposes a sub-transaction witness
+// set carrying redeemers, standing in for the Dijkstra shape.
+func withSubTxRedeemers(tx *babbage.BabbageTransaction) *subTxCarrier {
+	return &subTxCarrier{
+		BabbageTransaction: tx,
+		subWitnessSets: []common.TransactionWitnessSet{
+			babbage.BabbageTransactionWitnessSet{WsRedeemers: spendRedeemers()},
+		},
+	}
+}
+
 // TestCollateralKeyLockedOnlyForPhase2 is the blinklabs-io/dingo#3896
 // regression.
 //
@@ -22,54 +120,16 @@ import (
 // at an enterprise-script address. Holding it to the key-locked rule rejected a
 // canonical block and wedged the node.
 func TestCollateralKeyLockedOnlyForPhase2(t *testing.T) {
-	const txId = "811c9029fc79e6b552f54d857bf0db807c9f0d8b23dc9e1e37f382377018674f"
-	// The real collateral address: enterprise with a script payment credential.
-	scriptAddr, err := common.NewAddress(
-		"addr_test1wrpe58z89f3kwtq2cslsqvvw9hzdep2jy2ykx9kty8xnamqycrwy6",
+	ls := collateralFixtureLedgerState(t)
+	pp := &babbage.BabbageProtocolParameters{}
+	rule := productionRule(
+		t,
+		"UtxoValidateCollateralVKeyWitnesses",
+		babbage.UtxoValidateCollateralVKeyWitnesses,
 	)
-	if err != nil {
-		t.Fatalf("fixture address: %v", err)
-	}
-	input := shelley.NewShelleyTransactionInput(txId, 0)
-	utxos := []common.Utxo{
-		{
-			Id: input,
-			Output: babbage.BabbageTransactionOutput{
-				OutputAddress: scriptAddr,
-				OutputAmount: mary.MaryTransactionOutputValue{
-					Amount: 100_000_000,
-				},
-			},
-		},
-	}
-	ls := mockledger.NewLedgerStateBuilder().WithUtxos(utxos).Build()
-
-	newTx := func(withRedeemer bool) *babbage.BabbageTransaction {
-		wits := babbage.BabbageTransactionWitnessSet{}
-		wits.VkeyWitnesses = []common.VkeyWitness{
-			{Vkey: []byte{0x01}, Signature: []byte{0x02}},
-		}
-		if withRedeemer {
-			wits.WsRedeemers = alonzo.AlonzoRedeemers{
-				Redeemers: []alonzo.AlonzoRedeemer{
-					{Tag: common.RedeemerTagSpend, Index: 0},
-				},
-			}
-		}
-		return &babbage.BabbageTransaction{
-			Body: babbage.BabbageTransactionBody{
-				TxCollateral: cbor.NewSetType(
-					[]shelley.ShelleyTransactionInput{input}, false,
-				),
-			},
-			WitnessSet: wits,
-		}
-	}
 
 	t.Run("no phase-2 scripts: script collateral is accepted", func(t *testing.T) {
-		if err := babbage.UtxoValidateCollateralVKeyWitnesses(
-			newTx(false), 0, ls, &babbage.BabbageProtocolParameters{},
-		); err != nil {
+		if err := rule(collateralFixtureTx(false, 0), 0, ls, pp); err != nil {
 			t.Errorf(
 				"a transaction running no phase-2 scripts has nothing for "+
 					"collateral to cover, so it must not be held to the "+
@@ -83,21 +143,8 @@ func TestCollateralKeyLockedOnlyForPhase2(t *testing.T) {
 		// Reading only the top-level witness set would report no phase-2
 		// execution and skip the rule, which is the one direction this guard
 		// must never fail in.
-		tx := &subTxCarrier{
-			BabbageTransaction: newTx(false),
-			subWitnessSets: []common.TransactionWitnessSet{
-				babbage.BabbageTransactionWitnessSet{
-					WsRedeemers: alonzo.AlonzoRedeemers{
-						Redeemers: []alonzo.AlonzoRedeemer{
-							{Tag: common.RedeemerTagSpend, Index: 0},
-						},
-					},
-				},
-			},
-		}
-		if err := babbage.UtxoValidateCollateralVKeyWitnesses(
-			tx, 0, ls, &babbage.BabbageProtocolParameters{},
-		); err == nil {
+		tx := withSubTxRedeemers(collateralFixtureTx(false, 0))
+		if err := rule(tx, 0, ls, pp); err == nil {
 			t.Error(
 				"redeemers in a sub-transaction mean phase-2 execution, so " +
 					"the collateral rule must still apply",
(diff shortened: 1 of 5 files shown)


chrisguiney:
**Acceptance on the real chain.**

The wedged node was restarted on its own data directory, so it had only to get past the block that stopped it:

```
committed tip 15148378 -> 15184583
collateral rejections 0   (was 114)
```

Block 15148509 now applies and the node is following the chain again.

That confirms the change resolves the observed divergence. It does not confirm the *condition* is the one cardano-ledger uses — the caveat in the description stands, and a reviewer who knows that code should check it against the reference rather than against this block.

chrisguiney:
@coderabbitai review

ledger/common/witness.go
```diff
@@ -42,6 +42,23 @@ func ValidateCollateralVKeyWitnesses(
...
+	// script, no Plutus scripts and no redeemers, and a collateral input at an
+	// enterprise-script address. Holding it to the key-locked rule rejected a
+	// canonical block (blinklabs-io/dingo#3896).
+	//
+	// The presence of redeemers is the condition rather than the presence of
+	// Plutus scripts in the witness set: a script supplied by a reference input
+	// is not in the witness set, and gating on that would skip the check for
+	// exactly the transactions that most need it. Every phase-2 execution has a
+	// redeemer regardless of where its script came from.
+	if !transactionRunsPhase2Scripts(tx) {
```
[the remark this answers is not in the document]
chrisguiney: Valid, and it is the direction that matters — my guard could have become a way to skip the rule entirely, which is exactly what I wrote a test to prevent and then only tested at the top level.

`transactionRunsPhase2Scripts` read only `tx.Witnesses()`. A Dijkstra transaction carrying its redeemers in a sub-transaction would have reported no phase-2 execution, and the collateral rules would have been skipped for it.

Fixed in the head commit: it now aggregates over `SubTransactionWitnessSetsFromTransaction(tx)` as the other Dijkstra collateral rules do. The test gained a case using a transaction that exposes sub-transaction witness sets with redeemers only there, and it fails without the aggregation.

Worth noting the asymmetry in risk, since it shaped how I treated this: being too strict wedges a node on a canonical block, which is loud and gets found in an hour. Being too permissive silently skips a consensus check, which does not. This finding was the second kind.

wolf31o2: Two notes on the class this belongs to.

**One member is still ungated.** In the reference, `validateCollateral` / `validateTotalCollateral` is a group, and *all* of it sits behind the redeemer guard. gouroboros already gates three of them on `len(WsRedeemers.Redeemers) == 0` -- `UtxoValidateInsufficientCollateral`, `UtxoValidateCollateralContainsNonAda`, `UtxoValidateNoCollateralInputs`, in both alonzo and babbage. `babbage.UtxoValidateCollateralEqBalance` is not gated: it returns early only when `TotalCollateral()` is nil or zero. But it is Part 6 of `validateTotalCollateral`, inside the same guard. A transaction with declared collateral, a `total_collateral` field and no redeemers is still held to it, which is the same false-rejection shape this PR is closing. Suspected, not observed on chain -- but it is the one place left where the class is incomplete.

`babbage.UtxoValidateTooManyCollateralInputs` is correctly *outside* the guard: `validateTooManyCollateralInputs` is called from the UTXO transition, not from `feesOK`. No change wanted there.

**The guard now covers two different rules.** This helper does more than `validateScriptsNotPaidUTxO`, which only tests `vKeyLocked`. It also requires a matching vkey witness for each collateral input, and that requirement comes from UTXOW's `witsVKeyNeeded` (Alonzo includes collateral inputs), which is not redeemer-gated in the reference. So skipping it here is a missed rejection rather than a false one -- the safe direction, and not worth blocking on, but worth knowing the guard is broader than the reference's.

ledger/babbage/collateral_phase2_test.go
```diff
@@ -0,0 +1,131 @@
...
+			}
+		}
+		return &babbage.BabbageTransaction{
+			Body: babbage.BabbageTransactionBody{
+				TxCollateral: cbor.NewSetType(
+					[]shelley.ShelleyTransactionInput{input}, false,
+				),
+			},
+			WitnessSet: wits,
+		}
```
wolf31o2: Accept-case, reject-case and the sub-transaction direction are all here, and the accept-case fails without the change -- this is the right shape. Two smaller points: it drives `babbage.UtxoValidateCollateralVKeyWitnesses` directly rather than resolving it from `babbage.UtxoValidationRules`, so nothing pins that the rule is still in the production slice; and the fixture transaction has no `TxTotalCollateral`, so it would not notice the `UtxoValidateCollateralEqBalance` gap noted on `witness.go`.

wolf31o2 (commented):
Answering the "what I have not confirmed" section: the reading is right, and it is redeemers.

cardano-ledger `Alonzo.Rules.Utxo.feesOK`, Part 2:

```haskell
unless (null $ tx ^. witsTxL . rdmrsTxWitsL . unRedeemersL) $
  validateCollateral pp txBody utxoCollateral
```

Babbage's `feesOK` gates `validateTotalCollateral` with the identical guard, and Conway inherits Babbage. The gate is a non-empty redeemer *map*, not the presence of Plutus scripts in the witness set -- so gating on redeemers is exactly what the reference does, and the reference-input reasoning in the description is the right reason for it.

One remaining member of the same class inline. Branch is on `fab2538`, four commits behind `main`.

ledger/common/witness.go
chrisguiney: Both correct. Fixed in `5252120`.

**The ungated member.** `babbage.UtxoValidateCollateralEqBalance` now returns early on `!common.TransactionRunsPhase2Scripts(tx)` (`ledger/babbage/rules.go:653`), matching the guard the other three already carry.

It uses the interface-level helper rather than `tmpTx.WitnessSet.WsRedeemers` deliberately: Conway and Dijkstra both delegate into this one function, and the helper counts sub-transaction redeemers, so Dijkstra gets the same answer its own collateral rules get from `redeemerCount(tx)`.

I audited the class per era rather than assuming the copies agree:

| era | `CollateralEqBalance` | `TooManyCollateralInputs` |
|---|---|---|
| alonzo | not defined, not wired — no `total_collateral` field in the Alonzo body | not defined |
| babbage | was ungated, **now gated** | ungated, untouched |
| conway | thin delegate to babbage — inherits the gate | ungated, untouched |
| dijkstra | wires `conway.UtxoValidateCollateralEqBalance` — inherits the gate | ungated, untouched |

So babbage was the only place the rule has a body, and the other two eras route through it. The other three members were already gated in all four eras (alonzo/babbage on `len(...Redeemers) == 0`, conway on `WsRedeemers.Len() == 0`, dijkstra on `redeemerCount(tx) == 0`).

`UtxoValidateTooManyCollateralInputs` is unchanged in every era — the diff does not mention it.

Two existing fixtures had to change: `TestUtxoValidateCollateralEqBalance` in both babbage and conway built transactions with no redeemers, so under the new gate they would have passed while testing nothing. Each gained a redeemer.

**The broader guard.** Agreed, and it is now written down rather than left implicit — see the doc comment on `ValidateCollateralVKeyWitnesses` in `ledger/common/witness.go`. It names both halves (`validateScriptsNotPaidUTxO`, inside the guard; UTXOW's `witsVKeyNeeded`, not gated in the reference), states that gating both makes it a missed rejection rather than a false one, and gives the reason that matters — it can only accept something the reference rejects, so it cannot wedge a node on a canonical block. It also records that splitting the `witsVKeyNeeded` half back out to run ungated is the correct end state, left out of scope here.

Also rebased onto current `main` (`88145ad`).

ledger/babbage/collateral_phase2_test.go
chrisguiney: Both fixed in `aac1220`.

**Production wiring is now pinned.** The cases resolve the rule out of `babbage.UtxoValidationRules` by function identity and fail if it is not there, instead of calling the exported function directly:

```go
func productionRule(t *testing.T, name string, want common.UtxoValidationRuleFunc) common.UtxoValidationRuleFunc {
	wantId := reflect.ValueOf(want).Pointer()
	for _, rule := range babbage.UtxoValidationRules {
		if reflect.ValueOf(rule).Pointer() == wantId {
			return rule
		}
	}
	t.Fatalf("%s is not wired into babbage.UtxoValidationRules", name)
	return nil
}
```

**The `TxTotalCollateral` gap is covered.** New `TestCollateralEqBalanceOnlyForPhase2` uses a fixture that sets `TxTotalCollateral` to 1 ADA against a 100 ADA collateral input with no collateral return — a genuine mismatch, so the rule rejects it whenever it actually runs. Three cases: no redeemers accepts, redeemers reject, and redeemers only in a sub-transaction reject.

Each case was proven to fail without its fix. The mutations:

1. **Delete the redeemer guard** from `babbage.UtxoValidateCollateralEqBalance` →
   `TestCollateralEqBalanceOnlyForPhase2/no_phase-2_scripts` fails with `incorrect total collateral field: provided 100000000, total collateral 1000000`. This is the case that would have caught the gap you noted.
2. **Remove `UtxoValidateCollateralVKeyWitnesses` and `UtxoValidateCollateralEqBalance` from `UtxoValidationRules`** (leaving the functions defined) →
   both tests fail with `... is not wired into babbage.UtxoValidationRules`. Under the old shape they both still passed, which is exactly the hole.
3. **Make `TransactionRunsPhase2Scripts` read only the top-level witness set** →
   the sub-transaction case of *both* tests fails.

Full suite green and `make lint` at 0 issues on `aac1220`.

ledger/common/witness.go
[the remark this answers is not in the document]
chrisguiney: Re-verified against the current head (`aac1220`): still fixed.

The aggregation lives in `common.TransactionRunsPhase2Scripts`, `ledger/common/witness.go:147`:

```go
func TransactionRunsPhase2Scripts(tx Transaction) bool {
	if witnessSetHasRedeemers(tx.Witnesses()) {
		return true
	}
	for _, sub := range SubTransactionWitnessSetsFromTransaction(tx) {
		if witnessSetHasRedeemers(sub) {
			return true
		}
	}
	return false
}
```

Introduced in `aa37e7f` (was `22c1275` before the rebase onto `main`; the branch was rebased, so the old SHA no longer resolves).

Two things changed since your comment that are worth noting:

- The helper is now exported and is also the guard on `babbage.UtxoValidateCollateralEqBalance`, which Conway and Dijkstra both delegate into. So the newly gated rule counts sub-transaction redeemers for free, rather than repeating the top-level-only mistake in a second place.
- Coverage is still proven by mutation: reducing `TransactionRunsPhase2Scripts` to the top-level witness set fails the sub-transaction case of both `TestCollateralKeyLockedOnlyForPhase2` and `TestCollateralEqBalanceOnlyForPhase2`.

chrisguiney:
Addressed @wolf31o2's review. Head is now `aac1220`, rebased onto current `main` (`88145ad`).

**The ungated member of the class.** `babbage.UtxoValidateCollateralEqBalance` now sits behind the redeemer guard, as Part 6 of `validateTotalCollateral` inside `feesOK`. It used the interface-level helper rather than the Babbage-typed witness set, because Conway and Dijkstra delegate into this one function and the helper also counts sub-transaction redeemers.

Audited per era rather than assuming the copies agree — this family is duplicated and has drifted before:

| era | `CollateralEqBalance` | result |
|---|---|---|
| alonzo | not defined, not wired (no `total_collateral` in the Alonzo body) | nothing to do |
| babbage | had a body, ungated | **gated** |
| conway | thin delegate to babbage | inherits the gate |
| dijkstra | wires `conway.UtxoValidateCollateralEqBalance` | inherits the gate |

The other three members were already gated in all four eras. `UtxoValidateTooManyCollateralInputs` is untouched everywhere — it is called from the UTXO transition, not `feesOK`.

**The broader guard, documented.** No behaviour change. The doc comment on `ValidateCollateralVKeyWitnesses` now records that the helper covers both `validateScriptsNotPaidUTxO` (inside the guard) and UTXOW's `witsVKeyNeeded` (not redeemer-gated in the reference), so gating both is a missed rejection rather than a false one — it can only accept something the reference rejects, never reject a canonical block. It also notes that splitting the `witsVKeyNeeded` half back out is the correct end state, out of scope here.

**Tests.** The phase-2 cases now resolve the rule out of `babbage.UtxoValidationRules` by function identity and fail if it is not wired, so dropping a rule from the production slice can no longer pass. New `TestCollateralEqBalanceOnlyForPhase2` uses a fixture with `TxTotalCollateral` set to a genuine mismatch (1 ADA declared against a 100 ADA collateral input, no collateral return), covering accept / reject / sub-transaction.

Each new case was proven to fail without its fix:

1. Delete the guard from `UtxoValidateCollateralEqBalance` → the accept case fails with `incorrect total collateral field: provided 100000000, total collateral 1000000`.
2. Remove the two rules from `UtxoValidationRules`, leaving the functions defined → both tests fail with `... is not wired into babbage.UtxoValidationRules`; under the previous shape both still passed.
3. Reduce `TransactionRunsPhase2Scripts` to the top-level witness set → the sub-transaction case of both tests fails.

Two pre-existing fixtures changed: `TestUtxoValidateCollateralEqBalance` in babbage and conway built transactions with no redeemers, so under the new gate they would have passed while testing nothing. Each gained a redeemer.

**Validation on `aac1220`:** `go test ./... -count=1` passes (39 packages, exit 0); `make lint` reports 0 issues across the root module and all nine example modules, with a fresh `GOLANGCI_LINT_CACHE`.

**Not re-run:** the Preview replay from the earlier acceptance comment. That evidence still covers the key-locked rule, which is unchanged here. The `CollateralEqBalance` gate is the suspected-not-observed case @wolf31o2 flagged, and it has unit coverage only — I have no chain evidence of a transaction with `total_collateral`, declared collateral and no redeemers.

Cubic's sub-transaction finding was re-verified against this head and remains fixed; replied on the thread.

ledger/common/witness.go
```diff
@@ -34,6 +34,23 @@ type BootstrapWitness struct {
...
+//     witsVKeyNeeded (Alonzo adds collateral inputs to that set), which is not
+//     redeemer-gated in the reference.
+//
+// Gating both means a no-redeemer transaction with an unwitnessed key-locked
+// collateral input is not rejected here, where the reference would reject it in
+// UTXOW. That is a missed rejection, not a false one: it accepts a transaction
+// the reference rejects rather than rejecting one the reference accepts, so it
+// cannot wedge a node on a canonical block. Splitting the witsVKeyNeeded half
+// back out to run ungated is the correct end state; it is not done here because
+// this change is scoped to the false-rejection fix.
```
wolf31o2, line 53: Non-blocking, for a follow-up issue: this paragraph checks out against cardano-ledger `master`.

`getAlonzoWitsVKeyNeeded` is `getShelleyWitsVKeyNeeded certState utxo txBody \`Set.union\` reqSignerHashes`, and Shelley's `inputAuthors` folds over `spendableInputsTxBodyF`, which Alonzo defines as `allInputsTxBodyF = inputs \`Set.union\` collateralInputs` (`eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxBody.hs:243-248`). That set feeds UTXOW, not `feesOK`, so the reference does require a vkey witness per key-locked collateral input regardless of redeemers.

So the gap is real: after this change a no-redeemer transaction with an unwitnessed key-locked collateral input is accepted here. Safe direction, and gouroboros does not check spend-input authors at all today, so the asymmetry is narrower than it looks. Worth an issue so the split does not stay only in this comment.

wolf31o2 (approved):

ledger/common/witness.go
chrisguiney: Filed as blinklabs-io/gouroboros#2224, with your cardano-ledger derivation recorded — `getAlonzoWitsVKeyNeeded` unioning `reqSignerHashes` onto `getShelleyWitsVKeyNeeded`, and `inputAuthors` folding over Alonzo's `allInputsTxBodyF = inputs ∪ collateralInputs`, feeding UTXOW rather than `feesOK`.

It states the consequence as you framed it: a missed rejection rather than a false one, and narrower than it looks because gouroboros does not check spend-input authors at all today, so collateral is not uniquely unchecked.

Agreed that it should not live only in a comment. The helper's doc comment now records the current state and points at the split as the end state; the issue carries the reference derivation and the fix.

PR #2207, opened 2026-09-04
fix(ledger): charge the pool deposit when registering an already-retired pool

chrisguiney:
Reported as blinklabs-io/dingo#3908.

A pool whose retirement has taken effect is not currently registered, so
registering it again incurs the deposit. `PoolCurrentState` still returns the old
registration certificate — alongside the retirement epoch — and the value
conservation rules read only the registration, so a re-registration was treated
as a parameter update and charged nothing.

## How it surfaced

A from-genesis Preview replay wedged at slot 17020890 (epoch 197):

```
tx a82343228b1c4eaed145d988eaf8b5cb57e56d0883e3093fcfb5df07fa455b2f
  value not conserved: consumed 997574722, produced 497574722
```

`997574722 - 497574722 = 500000000` — the pool deposit, exactly.

The pool's history, from the node's own tables:

```
registered        slot 16988057   (epoch 196)
retired           slot 16988068   (epoch 196), targeting epoch 197
epoch 197 begins  slot 17020800
re-registered     slot 17020890   <- ninety slots after the retirement took effect
```

The chain charged a fresh 500 ADA deposit. The node did not, computed a produced
value one deposit short, rejected a canonical block, and stopped following the
chain — 159 rejections.

## The change

`PoolRegistrationDepositDue` puts the decision in one place: a deposit is due
when there is no registration at all, or when there is one but its retirement
epoch has been reached.

**The rule is duplicated per era**, and this is the second time in this session
that fixing one era left the others behaving as before, so all five call it —
Shelley, Mary, Alonzo, Babbage and Conway.

The current epoch comes from the optional `EpochState` capability, following the
pattern `UtxoValidateStakePoolRetirementWrongEpoch` already uses. Without it the
retirement bound cannot be evaluated and the pre-existing behaviour is kept.

Both possible errors are loud: charging a deposit that is not due and skipping
one that is both fail conservation. Neither silently admits an invalid
transaction, which is why degrading to the old behaviour is acceptable rather
than dangerous.

## Left alone deliberately

The remaining `PoolCurrentState` caller in `shelley/rules.go` is a VRF key
uniqueness check rather than a deposit computation. A retired pool's key
association is arguably still on record, so changing that would be a separate
judgement with its own reasoning.

I also considered making `PoolCurrentState` itself return no registration once
the retirement has taken effect, which is the more semantically accurate shape.
I did not, because it changes what every caller sees — including
`verify_block.go`, where a pool retiring at epoch N still legitimately produces
blocks until the boundary. That is a wider decision than this defect needs.

## Tests

`TestPoolRegistrationDepositDueAfterRetirement` covers never registered,
registered with no retirement, retirement still pending, retirement just taken
effect, long retired, and the degraded path with no `EpochState`. Suppressing
the epoch comparison fails the two retired cases, so it reproduces the defect
rather than restating the fix.

---
## Summary by cubic
Fixes value conservation so registering an already-retired pool charges the pool deposit, once per transaction. Previously, `PoolCurrentState` still returned the old registration certificate alongside the retirement epoch, so re-registration was treated as a parameter update and charged nothing — leaving produced value one deposit short and causing the node to reject canonical blocks (blinklabs-io/dingo#3908).

**Bug Fixes**
- Adds `PoolRegistrationDepositDue` in `ledger/common` and uses it in all five era rules (Shelley, Mary, Alonzo, Babbage, Conway), charging the deposit once per transaction per pool.
- Reads the current epoch through the optional `EpochState` capability; without it, keeps the pre-existing behavior since both possible errors fail conservation.
- Documents the `PoolCurrentState` guarantee that a retirement superseded by a later registration is reported as nil; leaves the VRF key uniqueness check in `shelley/rules.go` unchanged.
- Tests in `internal/ledgertest` run from all seven eras, including delegating Allegra and Dijkstra; cases cover never registered, pending retirement, retirement taken effect, long retired, no-`EpochState` degradation, and two registrations for the same pool being charged once.

<sup>Written for commit e17a41c1e5d4094b0cad2d368ae45571d14a1baf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2207?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected pool registration deposit handling when a pool has retired, ensuring re-registration requires the appropriate deposit.
  - Prevented duplicate deposit charges when multiple registrations for the same pool appear in one transaction.
  - Improved deposit assessment for pools with pending retirements or no retirement scheduled.
  - Applied consistent pool deposit accounting across supported ledger eras.


chrisguiney:
Correction to the validation claim, and a dependency worth stating plainly.

**This change alone does not fix Dingo.** Dingo's `LedgerView` does not
implement `EpochState`, so the type assertion in `PoolRegistrationDepositDue`
fails and the helper takes exactly the fallback path described above — the
pre-existing behaviour. I found this while adding the fix to the local build
tree: the node stayed wedged at slot 17020890.

The Dingo half is blinklabs-io/dingo#3911, which exposes the epoch mapping
Dingo's `LedgerState` already had. Both halves are required.

Re-validated with both: built v0.202.8 plus this change, ran against Dingo with
#3911, restarted on the data directory already wedged at 17020890. The node
passed the slot, crossed the epoch 197 boundary, and has replayed through epoch
207 since. Without #3911 it does not move.

This also makes the "degrading to the old behaviour is acceptable" argument
above worth re-reading. It is still true that neither error direction silently
admits an invalid transaction. But the degraded path is reached by a *silent*
optional type assertion, and the consumer that most needed this fix was
sitting on that path without anyone noticing. The safety of the fallback is
about correctness, not about whether the fix takes effect.

wolf31o2 (commented):
The rule matches cardano-ledger. `shelleyTotalDepositsTxBody` charges `numNewRegPoolCerts <x> ppPoolDeposit` where the set is `Set.fromList [hk | RegPoolTxCert (PoolParams {ppId = hk}) <- certs, not (isPoolRegistered hk)]`, `isPoolRegistered` is membership in `psStakePoolParams`, and POOLREAP removes a retiring pool at the *start* of its target epoch -- so `currentEpoch >= retirementEpoch` is the right bound. `Set.fromList` is also where the per-transaction dedup comes from, so `seenPoolRegistrations` matches.

Era coverage is complete: allegra delegates to shelley and dijkstra to conway, so the five edits cover all seven.

The deposit rule cannot newly reject on dingo's current `LedgerView`: `PoolCurrentState` already suppresses the retirement epoch when a later registration supersedes it, so a routine parameter update after a re-registration still sees `retirementEpoch == nil`. That dependency is undocumented, though -- see inline. Main issue is test placement. Branch is on `fab2538`, four commits behind `main`.

ledger/common/pool_deposit_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+// TestPoolRegistrationDepositDueAfterRetirement is the blinklabs-io/dingo#3908
+// regression.
+//
+// A pool whose retirement has taken effect is not currently registered, so
+// registering it again incurs the deposit. PoolCurrentState still returns the
+// old registration certificate, and reading only that treats the
+// re-registration as a parameter update and charges nothing — leaving the
+// produced value exactly one deposit short of what the chain charged.
+func TestPoolRegistrationDepositDueAfterRetirement(t *testing.T) {
```
wolf31o2, line 41: This tests the helper, not the changed call sites. The five era rules gained a second behaviour that nothing here exercises: `seenPoolRegistrations` now charges the deposit once per pool per transaction. Delete that map and every test in this PR still passes.

Missing case, exactly: a transaction carrying two `PoolRegistrationCertificate`s for the same previously-unregistered operator, driven through `shelley.UtxoValidateValueNotConservedUtxo` (resolved from `shelley.UtxoValidationRules`, not called bare), asserting the transaction conserves value with one `PoolDeposit` added to produced and fails with two. That is the assertion that reproduces the dedup rather than restating it.

A rule-level accept-case for the retired-pool path would be worth having in the same test -- the helper cases prove the predicate, not that the produced value moves by one deposit.

ledger/shelley/rules.go
```diff
@@ -363,11 +364,18 @@ func UtxoValidateValueNotConservedUtxo(
 	for _, cert := range tx.Certificates() {
 		switch tmpCert := cert.(type) {
 		case *common.PoolRegistrationCertificate:
-			reg, _, err := ls.PoolCurrentState(common.Blake2b224(tmpCert.Operator))
+			operator := common.Blake2b224(tmpCert.Operator)
+			if _, seen := seenPoolRegistrations[operator]; seen {
```
wolf31o2, line 368: Correctness here now rests on a `PoolCurrentState` guarantee that `ledger/common/state.go` does not state. Its doc says only "the epoch of a pending retirement certificate, if one exists". A provider that returned the latest retirement epoch unconditionally would make every parameter update *after* a re-registration see `currentEpoch >= retirementEpoch`, charge a second deposit, and fail conservation on a canonical block -- the same defect this PR fixes, inverted.

Dingo's `LedgerView` does the right thing today (`registrationSupersedesRetirement` compares registration and retirement slot/cert IDs), so this is not a live bug. But since `PoolRegistrationDepositDue` is the first consumer to depend on it, the guarantee belongs in the `PoolState` interface doc: a retirement superseded by a later registration must not be returned.

ledger/common/pool_deposit_test.go
chrisguiney: Correct, and the "delete that map" check confirms it: with `seenPoolRegistrations` removed from `ledger/shelley/rules.go`, every test in the previous head still passed.

Added `TestUtxoValidateValueNotConservedUtxoPoolDeposits` in `ledger/shelley/rules_test.go`. It resolves the validator from `shelley.UtxoValidationRules`, indexed by matching `common.UtxoValidationRuleValueNotConserved` in `shelley.UtxoValidationRuleDescriptors()`, so an unregistered or reordered rule fails there rather than being papered over by a direct function call.

Each case declares how many deposits the rule must add, then asserts three things: budgeting exactly that many conserves value; budgeting one more fails with `Consumed - Produced == +PoolDeposit`; and, where a deposit is expected, budgeting one less fails with `-PoolDeposit`. Bounding it from both sides is what makes "the produced value moves by exactly one deposit" an assertion rather than a restatement.

Cases: unregistered pool; live pool (parameter update, no deposit); retirement still pending (no deposit); retirement taken effect (one deposit) — the rule-level accept-case you asked for; and two `PoolRegistrationCertificate`s for the same operator against an unregistered pool, a retired pool, and a live pool.

Proof both ways, run against the new test:

- Delete `seenPoolRegistrations` from `ledger/shelley/rules.go`: `two_registrations_for_an_unregistered_pool` and `two_registrations_for_a_retired_pool` fail with `value not conserved: consumed 3000000000, produced 3500000000` — one deposit too many.
- Revert `PoolRegistrationDepositDue` to `return reg == nil`: `retirement_has_taken_effect` and `two_registrations_for_a_retired_pool` fail with `consumed 3000000000, produced 2500000000` — one deposit short, which is dingo#3908.

`go test ./... -count=1` passes and `make lint` reports 0 issues.

ledger/shelley/rules.go
chrisguiney: Agreed — the guarantee was load-bearing and unstated. Documented it on `PoolCurrentState` in the `PoolState` interface (`ledger/common/state.go`):

> The returned retirement epoch must be pending relative to the returned registration. A retirement that a later registration superseded is no longer pending and must be reported as nil, not as the pool's most recent retirement.
>
> `PoolRegistrationDepositDue` reads the pair together and treats a retirement epoch the current epoch has reached as evidence that the pool is no longer registered, so an implementation that returned the latest retirement unconditionally would charge a second pool deposit on every parameter update made after a re-registration and reject a canonical block for failing value conservation.

Also cross-referenced it from `PoolRegistrationDepositDue`, so the dependency is visible from the consumer as well as from the contract.

Documentation only. Dingo's `LedgerView` already satisfies it via `registrationSupersedesRetirement`, so nothing changed there.

ledger/common/pool_deposit.go
```diff
@@ -0,0 +1,64 @@
...
+// the retirement bound cannot be evaluated, so the pre-existing behaviour is
+// kept: the registration on record is taken at face value. Both possible errors
+// are loud -- charging a deposit that is not due and skipping one that is both
+// fail conservation -- so neither silently admits an invalid transaction.
+func PoolRegistrationDepositDue(
+	ls LedgerState,
+	slot uint64,
+	operator PoolKeyHash,
+) (bool, error) {
+	reg, retirementEpoch, err := ls.PoolCurrentState(operator)
```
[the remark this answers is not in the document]
chrisguiney: Valid, and already fixed on the current head — this comment was written against `dbd7c62`, before `a76cc65` ("fix: charge pool deposit once per transaction") landed.

The fix is `seenPoolRegistrations`, a `map[common.PoolKeyHash]struct{}` in each era's `UtxoValidateValueNotConservedUtxo`. The certificate loop skips an operator it has already charged, so a second registration for a pool the first certificate just registered incurs nothing. That is transaction-local state on the side that needs it; the helper itself stays a pure question about ledger state and is not the right place to hold per-transaction bookkeeping.

This matches cardano-ledger: `shelleyTotalDepositsTxBody` charges `numNewRegPoolCerts <x> ppPoolDeposit` over `Set.fromList [hk | RegPoolTxCert (PoolParams {ppId = hk}) <- certs, not (isPoolRegistered hk)]`, and the `Set` is where the dedup comes from.

What was missing was test coverage, which the reviewer raised separately. `91e916c` adds `TestUtxoValidateValueNotConservedUtxoPoolDeposits`, driving the rule as registered in `shelley.UtxoValidationRules`; deleting `seenPoolRegistrations` now fails the two-certificate cases with `consumed 3000000000, produced 3500000000`.

chrisguiney:
Pushed `91e916c` addressing both review findings.

**Test placement (dedup was untested).** The helper tests exercised `PoolRegistrationDepositDue`, not the era rules that call it — deleting `seenPoolRegistrations` left every test passing. `TestUtxoValidateValueNotConservedUtxoPoolDeposits` in `ledger/shelley/rules_test.go` now drives the validator resolved from `shelley.UtxoValidationRules`, indexed by matching `common.UtxoValidationRuleValueNotConserved` in `shelley.UtxoValidationRuleDescriptors()`, rather than calling the function bare.

Each case declares the number of deposits the rule must add, then asserts that budgeting exactly that many conserves value, that budgeting one more fails with `Consumed - Produced == +PoolDeposit`, and that budgeting one less (where a deposit is expected) fails with `-PoolDeposit`. Bounding it from both sides makes "moves by exactly one deposit" an assertion rather than a restatement. Cases cover the unregistered pool, the live-pool parameter update, a still-pending retirement, an effective retirement (the rule-level accept-case), and two `PoolRegistrationCertificate`s for the same operator against unregistered, retired, and live pools.

Both directions verified by mutation:

| Mutation | Failing cases | Symptom |
| --- | --- | --- |
| Delete `seenPoolRegistrations` | `two_registrations_for_an_unregistered_pool`, `two_registrations_for_a_retired_pool` | `consumed 3000000000, produced 3500000000` — one deposit too many |
| Revert helper to `return reg == nil` | `retirement_has_taken_effect`, `two_registrations_for_a_retired_pool` | `consumed 3000000000, produced 2500000000` — one deposit short (dingo#3908) |

**Unstated `PoolCurrentState` guarantee.** Documented on the `PoolState` interface in `ledger/common/state.go`: a retirement superseded by a later registration is no longer pending and must be reported as nil, not as the pool's most recent retirement, because an implementation returning the latest retirement unconditionally would charge a second deposit on every parameter update after a re-registration and reject a canonical block. Cross-referenced from `PoolRegistrationDepositDue`. Documentation only — Dingo's `LedgerView` already satisfies it via `registrationSupersedesRetirement`, and nothing changed there.

No production behaviour changed in this commit; it is tests plus interface documentation. The branch is also current with `main` (`fab2538`), not behind as noted earlier.

Validation: `go test ./... -count=1` passes; `make lint` reports 0 issues across the module and all nine examples.

Cubic thread on `ledger/common/pool_deposit.go:43` answered — it was written against `dbd7c62`, and `seenPoolRegistrations` in `a76cc65` is the fix.

ledger/common/pool_deposit.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	if reg == nil {
+		return true, nil
+	}
+	if retirementEpoch == nil {
+		return false, nil
+	}
+	epochState, ok := ls.(EpochState)
+	if !ok {
+		return false, nil
```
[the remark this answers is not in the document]
chrisguiney: Already addressed, in the consumer rather than here.

Dingo's `LedgerView` did not implement `common.EpochState`, which is exactly the gap described — the assertion failed, `PoolRegistrationDepositDue` returned its degraded answer, and the deposit was omitted. blinklabs-io/dingo#3911 adds `EpochForSlot` to `LedgerView`, delegating to the epoch cache the header path already uses.

Both halves are required and neither is sufficient alone. I confirmed that directly: with this PR applied but without #3911, the Preview replay stays wedged at slot 17020890. With both, it clears the slot, crosses the epoch 197 boundary, and has since replayed through epoch 475.

The replay coverage you ask for exists as a rule-level test on this PR rather than a 197-epoch sync: `TestUtxoValidateValueNotConservedUtxoPoolDeposits` drives the value-conservation rule resolved from `shelley.UtxoValidationRules`, and its `retirement_has_taken_effect` case fails with `consumed 3000000000, produced 2500000000` — one deposit short, the slot 17020890 shape — when the predicate is reverted.

ledger/shelley/rules.go
```diff
@@ -363,11 +364,18 @@ func UtxoValidateValueNotConservedUtxo(
...
+				continue
+			}
+			seenPoolRegistrations[operator] = struct{}{}
+			depositDue, err := common.PoolRegistrationDepositDue(
+				ls, slot, operator,
+			)
```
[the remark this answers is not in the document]
chrisguiney: Already addressed, in the consumer rather than here.

Dingo's `LedgerView` did not implement `common.EpochState`, which is exactly the gap described — the assertion failed, `PoolRegistrationDepositDue` returned its degraded answer, and the deposit was omitted. blinklabs-io/dingo#3911 adds `EpochForSlot` to `LedgerView`, delegating to the epoch cache the header path already uses.

Both halves are required and neither is sufficient alone. I confirmed that directly: with this PR applied but without #3911, the Preview replay stays wedged at slot 17020890. With both, it clears the slot, crosses the epoch 197 boundary, and has since replayed through epoch 475.

The replay coverage you ask for exists as a rule-level test on this PR rather than a 197-epoch sync: `TestUtxoValidateValueNotConservedUtxoPoolDeposits` drives the value-conservation rule resolved from `shelley.UtxoValidationRules`, and its `retirement_has_taken_effect` case fails with `consumed 3000000000, produced 2500000000` — one deposit short, the slot 17020890 shape — when the predicate is reverted.

ledger/conway/rules.go
```diff
@@ -2204,11 +2205,18 @@ func UtxoValidateValueNotConservedUtxo(
 	for _, cert := range tx.Certificates() {
 		switch tmpCert := cert.(type) {
 		case *common.PoolRegistrationCertificate:
-			reg, _, err := ls.PoolCurrentState(common.Blake2b224(tmpCert.Operator))
+			operator := common.Blake2b224(tmpCert.Operator)
+			if _, seen := seenPoolRegistrations[operator]; seen {
+				continue
+			}
+			seenPoolRegistrations[operator] = struct{}{}
+			depositDue, err := common.PoolRegistrationDepositDue(
```
wolf31o2, line 2213: Conway, Babbage, Alonzo and Mary carry this change with nothing behind it. Reverting all four to `reg, _, err := ls.PoolCurrentState(operator)` / `if reg == nil` and dropping the dedup map leaves `go build ./...` and `go test ./ledger/... -count=1` fully green -- eleven packages, zero failures. Only `ledger/shelley` fails when Shelley is reverted, via `TestUtxoValidateValueNotConservedUtxoPoolDeposits`.

Each era's `UtxoValidateValueNotConservedUtxo` is a separate function in a separate package with its own descriptor list. Allegra delegates to Shelley and Dijkstra to Conway, so those two are covered transitively; Mary, Alonzo, Babbage and Conway are not. The stated rationale for touching five eras is that this rule has already drifted per era, and the fence now stands around one of them.

dingo#3908 is Preview epoch 197, which is Conway. That is the era with no coverage.

`ledger/conway/certificate_deposits_test.go` already builds Conway transactions against `mockledger` and drives the registered rule (`TestCertificateRegistrationDepositsProductionPath`), so the Shelley table maps onto it directly: unregistered pool, live pool, pending retirement, effective retirement, two registrations for one operator. Conway is the minimum; Mary, Alonzo and Babbage are the same shape.

ledger/common/pool_deposit_test.go
```diff
@@ -0,0 +1,97 @@
+package common_test
```
wolf31o2, line 15: Non-blocking: every other `_test.go` in `ledger/common` opens with the Apache header. This file starts at `package common_test`.

wolf31o2 (changes requested):

wolf31o2 (approved):

ledger/conway/rules.go
chrisguiney: Confirmed and fixed in `e17a41c`. Your revert reproduced exactly: with all five call sites back on `reg == nil` and the dedup maps dropped, `go build ./...` and `go test ./ledger/... -count=1` were green apart from `ledger/shelley`. Only Shelley was covered.

The cases now live in `internal/ledgertest` as a shared table-driven driver, and all seven eras run it from their own package. Each era supplies its own `UtxoValidationRules` plus `UtxoValidationRuleDescriptors`, its own protocol parameters, and a builder for a transaction of its own type; the driver resolves the validator by id from that era's descriptors, so the production wiring is pinned per era rather than assumed. A new package was needed because `internal/test` is imported by in-package test files, and pulling `ouroboros-mock` in there created an import cycle.

On your delegation point: Mary, Alonzo, Babbage and Conway each carry a full copy of the rule body. Allegra delegates to Shelley and Dijkstra to Conway, so they own only the delegation and the pparams conversion — their rows pin that, and they also fail when the era they delegate to is reverted.

Two mutation classes, run per era, each verified to still compile so the result is semantic and not syntactic breakage:

**A. Full revert** (call site to `reg, _, err := ls.PoolCurrentState(...)` / `if reg == nil`, dedup map dropped) — byte-for-byte the pre-PR code from `fab2538`:

| mutated | `go build` | failing package(s) | failing subtests |
| --- | --- | --- | --- |
| shelley | OK | `ledger/shelley`, `ledger/allegra` | retirement_has_taken_effect, two_registrations_for_an_unregistered_pool, two_registrations_for_a_retired_pool |
| mary | OK | `ledger/mary` | same three |
| alonzo | OK | `ledger/alonzo` | same three |
| babbage | OK | `ledger/babbage` | same three |
| conway | OK | `ledger/conway`, `ledger/dijkstra` | same three |

All five reverted at once: seven packages fail, where before it was zero.

**B. Dedup-only** (predicate kept, only the per-transaction guard removed) — isolates the dedup from the predicate. Each of the five fails `two_registrations_for_an_unregistered_pool` and `two_registrations_for_a_retired_pool`, and `retirement_has_taken_effect` correctly does *not* fail, since that case is the predicate's.

`go test ./... -count=1` passes and `make lint` reports 0 issues. `nilaway` is unavailable rather than clean: under the go1.27.1 toolchain it produces no output and exits 0, so it analysed nothing.

ledger/common/pool_deposit_test.go
chrisguiney: Fixed in `e17a41c` — added the Apache header to `ledger/common/pool_deposit_test.go`, copied verbatim from `ledger/common/certs_test.go` so it matches its siblings exactly.

One correction to the finding, in your favour but worth stating precisely: it is not quite every other `_test.go` in the package. `ledger/common/metadata_test.go` also has no header. That one predates this PR and is untouched here, so I left it alone rather than widening the diff — flagging it in case you want it swept separately. Every file this PR adds or touches carries the header.

ledger/common/pool_deposit.go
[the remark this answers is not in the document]
chrisguiney: Follow-up now that the head has moved to `e17a41c`: this finding remains addressed, and the coverage behind it is materially stronger than when I last replied.

`seenPoolRegistrations` is still the fix, unchanged. What has changed is that the dedup is now verified in every era that carries it, not just Shelley. Removing only the per-transaction guard — leaving `PoolRegistrationDepositDue` in place — fails `two_registrations_for_an_unregistered_pool` and `two_registrations_for_a_retired_pool` in each of `ledger/shelley`, `ledger/mary`, `ledger/alonzo`, `ledger/babbage` and `ledger/conway`, and in `ledger/allegra` and `ledger/dijkstra` through delegation. Each mutation was checked to still compile, so those are behavioural failures rather than build breakage.

PR #2208, opened 2026-09-04
fix(ledger): preserve minPoolCost across Mary upgrade

### LINKED ISSUE #2184: mary.UpgradePParams hardcodes 340 ADA minPoolCost now that Shelley carries the value
#2165 added `ppMinPoolCost` to the Shelley and Allegra protocol parameter structs (they had 17 array fields where the reference has 18, so the real parameter array failed to decode outright).

`mary.UpgradePParams` still hardcodes the 340 ADA upgrade default. It was left alone in #2165 because it is deliberately tested, but now that the value is carried through from Shelley the hardcoded constant should come from the previous era parameters instead.

Acceptance criteria: the Mary upgrade derives `minPoolCost` from the Allegra parameters; the existing upgrade test is updated to assert propagation rather than the constant.

wolf31o2:
Resolves #2184.

Shelley carries `minPoolCost`, so the Mary protocol-parameter upgrade now preserves the value instead of replacing it with the historical 340 ADA constant. Regression coverage verifies propagation through Allegra, Mary, Alonzo, Babbage, and Conway.

## Summary by CodeRabbit

- **Bug Fixes**
  - Protocol parameter upgrades now preserve the configured minimum pool cost when moving from the Shelley era into Mary and later eras.
  - Prevents the minimum pool cost from being unexpectedly reset to a fixed default during upgrades.
  - Added coverage to verify that the configured value remains unchanged across the full sequence of protocol era upgrades.

### DIFF
--- a/ledger/mary/pparams.go
+++ b/ledger/mary/pparams.go
@@ -244,7 +244,7 @@ func UpgradePParams(
 		ProtocolMajor:      prevPParams.ProtocolMajor,
 		ProtocolMinor:      prevPParams.ProtocolMinor,
 		MinUtxoValue:       prevPParams.MinUtxoValue,
-		MinPoolCost:        340000000, // initial MinPoolCost at Mary on mainnet
+		MinPoolCost:        prevPParams.MinPoolCost,
 	}
 }
 
--- a/ledger/mary/pparams_test.go
+++ b/ledger/mary/pparams_test.go
@@ -614,6 +614,7 @@ func TestMaryUpgradePParams(t *testing.T) {
 				ProtocolMajor:      6,
 				ProtocolMinor:      0,
 				MinUtxoValue:       1000000,
+				MinPoolCost:        123456789,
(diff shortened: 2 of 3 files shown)


arepala-uml (approved):
`mary.UpgradePParams` was hardcoding `MinPoolCost: 340000000` instead of carrying forward `prevPParams.MinPoolCost` from Shelley, which does populate the field from genesis. Confirmed the bug on the pre-fix commit and verified every other era (Mary→Alonzo→Babbage→Conway, and the Shelley→Allegra alias) already propagates the field correctly — this was the only broken link.

New tests cover both the direct Mary upgrade case and cross-era propagation through Conway. Ran the affected package tests plus `go vet` on the PR head; all pass. CI (test, lint, nilaway, CodeQL) is green and both CodeRabbit and cubic reported no actionable findings.

No behavioral, wire-compatibility, or coverage gaps found.

PR #2209, opened 2026-09-04
fix(ledger): enforce pool metadata URL bounds

### LINKED ISSUE #2079: utxorpc: handle pool registrations without metadata
UTxO-RPC conversion dereferences pool metadata unconditionally even though pool
registration certificates may omit metadata. Valid registrations without it
cannot be converted.

Acceptance criteria:

- Represent omitted pool metadata without dereferencing a nil value.
- Preserve populated metadata fields unchanged.
- Cover registrations with and without metadata.

The correction is included in #2209.

### LINKED ISSUE #2183: PoolMetadata.Url has no length bound
The reference `DecCBOR Url` bounds the URL at 64 bytes before decoder version 9 and 128 bytes from version 9. gouroboros applies no bound, so it accepts pool metadata URLs cardano-node rejects at decode.

Decode-level, not a POOL predicate, so it was out of scope for #2165.

Acceptance criteria: the bound is enforced at decode with the version-dependent limit; tests cover 64/65 and 128/129 at the relevant decoder versions.

wolf31o2:
Enforce pool metadata URL limits for on-chain data: 64 bytes before protocol
version 9 and 128 bytes thereafter. Preserve absent metadata in UTxO-RPC
conversion, including protobuf round trips.

Keep JSON and genesis URL values unbounded, matching the reference ledger's
`Url` JSON representation. Cover standard and extra genesis pools without
applying the CBOR constraint to configuration data. JSON output uses the
reference's lowercase `url` and `hash` keys; decoding accepts the previous
capitalized keys too.

Closes #2183. Closes #2079.

## Summary by cubic

Adds URL boundary and UTF-8 byte-length tests across Shelley-through-Dijkstra
POOL entries and metadata codecs.

## Summary by CodeRabbit

Measures URL limits in bytes and applies protocol-version-specific bounds
during pool registration.


ledger/shelley/genesis.go
```diff
@@ -279,6 +279,14 @@ func (g *ShelleyGenesis) effectivePools() (map[string]common.PoolRegistrationCer
...
 		}
 	}
+	for poolID, pool := range out {
+		if err := common.ValidatePoolMetadataForProtocolVersion(
+			pool.PoolMetadata,
+			g.ProtocolParameters.ProtocolVersion.Major,
+		); err != nil {
+			return nil, fmt.Errorf("invalid pool %s metadata: %w", poolID, err)
+		}
+	}
```
chrisguiney: Blocker: this loop only runs when `extraConfig.stakePools.data` is non-empty. Line 167 returns `g.Staking.Pools` unchanged before reaching it, so the same genesis pool is accepted or rejected depending on an unrelated field.

Reproduced on `9a5734e` with one `staking.pools` entry carrying a 65-byte metadata URL at `protocolVersion.major = 2`:

- `staking.pools` alone -> `InitialPools()` returns 1 pool, `err = nil`
- same `staking.pools` plus one unrelated `extraConfig.stakePools.data` entry -> `invalid pool 1e3105f2...005b metadata: pool metadata URL exceeds the protocol length limit: maximum 64 bytes, got 65`

Separately, the reference does not bound this at genesis at all. `Url` is `deriving newtype (... FromJSON ...)` from `Text` in `BaseTypes.hs`, `FromJSON PoolMetadata` reads `obj .: "url"` through that instance, and `validateGenesis` in `eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs` checks only epoch length, KES evolutions and quorum size. So the strict branch rejects a genesis `cardano-node` loads, and a node that cannot load its genesis does not start.

Either drop the genesis check, or hoist it above the early return at line 167 so both pool sources get the same treatment.

ledger/shelley/rules.go
```diff
@@ -819,6 +819,8 @@ func UtxoValidateWithdrawals(
 //   - WrongNetworkPOOL: a registration's reward account must be on the
 //     ledger's network. Gated on major protocol version > 4
 //     (hardforkAlonzoValidatePoolAccountAddressNetID).
+//   - PoolMetadataURLTooLong: a registration's metadata URL must fit the
+//     era-specific byte bound (64 before protocol version 9, 128 thereafter).
```
chrisguiney, line 890: `PoolMetadataURLTooLong` is not a predicate in that file. `ShelleyPoolPredFailure` in `eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs` has `StakePoolNotRegisteredOnKeyPOOL`, `StakePoolRetirementWrongEpochPOOL`, `StakePoolCostTooLowPOOL`, `WrongNetworkPOOL`, `PoolMedataHashTooBig` and `VRFKeyHashAlreadyRegistered`, and `poolTransition` raises no URL failure. The bound lives in `DecCBOR Url` in `libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs`.

The enclosing comment says "Predicates enforced here, all from that file", and the paragraph below it is careful about which reference predicate is *not* reimplemented and why, so this entry reads as provenance it does not have. Better to state that the reference enforces the bound in `DecCBOR Url` at decode, keyed on the block's era decoder version, and that this package applies it here because the shared `PoolMetadata` decoder has no era context. That is the fact a later reader needs.

ledger/shelley/rules.go
```diff
@@ -941,10 +944,18 @@ func validatePoolRegistration(
...
 	checkNetworkId bool,
 	checkVrfKeys bool,
 	inTxVrfKeys map[common.VrfKeyHash]common.PoolKeyHash,
 ) error {
+	if err := common.ValidatePoolMetadataForProtocolVersion(
+		cert.PoolMetadata,
+		protocolMajor,
+	); err != nil {
+		return err
+	}
```
chrisguiney, line 1023: Non-blocking, two notes on placing this first.

Every other failure in this function returns a typed error a consumer can match on (`WrongNetworkPoolError`, `StakePoolCostTooLowError`); this one returns a bare wrapped `ErrPoolMetadataURLTooLong` with no pool identity. A `PoolMetadataUrlTooLongError` carrying `PoolKeyHash`, the supplied length and the limit would match the surrounding style and make the failure attributable to a pool.

A certificate that is both on the wrong network and over the URL bound now reports the URL first. `poolTransition` evaluates `WrongNetworkPOOL` before anything else, so moving this below the network and cost checks keeps first-error order aligned with the reference.

ledger/common/certs.go
```diff
@@ -479,7 +479,113 @@ type PoolMetadata struct {
...
+	}{
+		Url:  p.Url,
+		Hash: p.Hash,
+	})
+}
+
 func (p *PoolMetadata) Utxorpc() (*utxorpc.PoolMetadata, error) {
+	if err := ValidatePoolMetadata(p); err != nil {
+		return nil, err
+	}
```
chrisguiney, line 655: Non-blocking, pre-existing. `ValidatePoolMetadata` returns `nil` for a nil `metadata`, so this guard passes for a nil receiver and line 590 then dereferences `p.Url`. `PoolRegistrationCertificate.Utxorpc()` calls `c.PoolMetadata.Utxorpc()` unconditionally at line 1318, and `pool_metadata` is optional in every era, so a registration without metadata panics.

Confirmed identical on `88145ad` and `9a5734e`: `PoolRegistrationCertificate{Margin: NewGenesisRat(1, 10)}.Utxorpc()` panics with a nil pointer dereference on both, so this PR does not introduce it. But the new guard sits directly above the dereference and reads as though it handled nil. `if p == nil { return nil, nil }` here would close it; otherwise it deserves its own issue.

ledger/common/certs.go
```diff
@@ -479,7 +479,113 @@ type PoolMetadata struct {
...
+	if err := ValidatePoolMetadata(&p); err != nil {
+		return nil, err
+	}
+	return json.Marshal(struct {
+		Url  string           `json:"url"`
+		Hash PoolMetadataHash `json:"hash"`
+	}{
+		Url:  p.Url,
+		Hash: p.Hash,
+	})
```
chrisguiney, line 649: Non-blocking. This changes `PoolMetadata`'s JSON output from `{"Url":...,"Hash":...}` to `{"url":...,"hash":...}` — default field-name marshaling before, explicit tags now.

The new keys match cardano-ledger's `ToJSON PoolMetadata` and the lowerCamel tags on the rest of `PoolRegistrationCertificate`, and unmarshaling stays compatible because Go's tag matching is case-insensitive (verified both key shapes decode on `88145ad` and `9a5734e`). It is still an output change for anything that serializes a pool registration certificate, and neither the commit body nor the PR description mentions it. Worth a line in the commit body.

chrisguiney (dismissed):
Requesting changes on one finding in `ledger/shelley/genesis.go`. Everything else below is non-blocking, and the core of the change checks out.

**Reference behaviour, verified**

- `DecCBOR Url` in `libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs` bounds the URL with `textSizeN`, which measures `lengthWord8` — UTF-8 **bytes**, `<= n` — and `ifDecoderVersionAtLeast (natVersion @9)` selects 128 from decoder version 9, 64 below it. CDDL agrees: `url = text .size (0 .. 64)` in `shelley.cddl`, `(0 .. 128)` in `conway.cddl`.
- It is a **decode** bound, not a POOL predicate. `poolTransition` in `eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs` raises only `StakePoolCostTooLowPOOL`, `WrongNetworkPOOL`, `PoolMedataHashTooBig` and `VRFKeyHashAlreadyRegistered`. Because blocks decode at the block's own era version (`decodeShelleyBlock = eraDecoder @era`, i.e. `eraProtVerLow`), the reference's 64/128 split falls on the same boundary as this PR's `protocolMajor < 9`, and both outcomes reject the block. Enforcing it in the rule is sound.
- Metadata **hash** length: the reference `pmHash` is an unbounded `ByteArray` and `pool_metadata = [url, bytes]` in both CDDLs. This PR does not touch hash handling.

**No sync risk from real chain data** (Koios `pool_list` + `pool_updates`): mainnet 6000 pools, 4 with URLs over 64 bytes (65/69/84/98) registered at epochs 634/648/655/564; preview 15 over 64 (max 121), earliest at epoch 828; preprod 4 over 64 (max 104), earliest at epoch 257. Protocol major is >= 9 at every one of those epochs. Nothing on any of the three networks exceeds 128, so no stored certificate is rejected by either the 128 codec bound or the era-aware rule.

**Era class is complete.** `allegra`, `mary`, `alonzo`, `babbage` and `conway` all delegate to `shelley.UtxoValidatePoolCertificates`, and `dijkstra` wires `conway.UtxoValidatePoolCertificates`. There is one implementation, and the new table exercises all seven entries.

**Wire compatibility.** `PoolMetadata` CBOR output is byte-identical to base for a normal certificate on both trees, so the new `MarshalCBOR` is not a wire change. Failures are returned errors, not panics.

**Fail-without-fix**, reverting each hunk in place at `9a5734e`:

- POOL-rule call removed -> `TestPoolMetadataURLBoundThroughPoolRule` fails on `pre-Conway over maximum` and `Conway over maximum` for all seven eras. The `maximum` cases pass either way; they are guards.
- `PoolMetadata` codec methods removed -> all four `TestPoolMetadataURLDecodeBound/over maximum *` subtests and `TestPoolMetadataURLLengthUsesBytes/over 128 UTF-8 bytes` fail.
- Genesis loop removed -> `TestGenesisExtraConfigPoolFieldValidation/legacy metadata URL too long` fails.

**Checks** on `9a5734e` and on a trial merge with `origin/main` (branch is 37 commits behind): `go build ./...` 0, `go vet ./...` 0, `go test ./ledger/...` 0, `go test -race ./ledger/...` 0, `gofmt -s -l .` clean, `golangci-lint run ./...` 0 issues (isolated cache), plus `go build`/`go vet` 0 in all 8 nested example modules. `go test ./...` on the merged tree hit `pipeline.TestValidateStage_Eta0ProviderError`; that assertion is `assert.Greater(item.ValidateDuration(), time.Duration(0))` and its sibling `TestValidateStage_NoProviderReturnsError` fails the same way on `origin/main` (1/15 whole-package runs), so it is a pre-existing wall-clock flake unrelated to this change.

**Bot standing on this head:** neither bot reviewed `9a5734e`. CodeRabbit's only review covers `88145ad..e138001` (commit 1 of 3) and its status comment is "Auto incremental reviews are disabled" — the green CodeRabbit check is that skip, not a review. Cubic's review is anchored to `f42a673` (commit 2), 80 minutes before `9a5734e` was pushed, so its "verified against the latest diff" line does not describe this head, and its open `certs.go` finding is only partly addressed by `9a5734e` (see the inline comment on `genesis.go`). CodeRabbit's merge-risk note about the codec still admitting 65-128 byte URLs for pre-version-9 data is accurate but acceptable: the shared `PoolMetadata` decoder has no era context, and erring permissive at decode is the safe direction.

**Class follow-up, not for this PR:** `DnsName` carries the identical `ifDecoderVersionAtLeast (natVersion @9)` 64/128 bound in the reference (`dns_name = text .size (0 .. 64)` / `(0 .. 128)`), and `PoolRelay.Hostname` here uses a flat `poolRelayMaxHostnameLen = 128` with no era branch. Pre-existing; this PR does not regress it.

ledger/common/certs.go
```diff
@@ -479,7 +479,103 @@ type PoolMetadata struct {
...
+	if p == nil {
+		return errors.New("nil PoolMetadata receiver")
+	}
+	type poolMetadata PoolMetadata
+	var tmp poolMetadata
+	if _, err := cbor.Decode(data, &tmp); err != nil {
+		return err
+	}
+	metadata := PoolMetadata(tmp)
+	if err := ValidatePoolMetadata(&metadata); err != nil {
```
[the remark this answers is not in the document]
chrisguiney: Partly valid, partly addressed. `9a5734e` added an era-aware check in `ShelleyGenesis.effectivePools`, so the premise (`UnmarshalJSON` always applies 128 and genesis pools never see the POOL rule) no longer holds for every genesis — but it still holds for most of them. The new loop sits after the early return at `ledger/shelley/genesis.go:167`, so it only runs when `extraConfig.stakePools.data` is non-empty; a genesis whose pools come from the standard `staking.pools` section is still accepted with a 65-byte URL at `protocolVersion.major = 2`. Reproduced on this head, with the identical pool rejected once an unrelated `extraConfig` stake pool is added.

The suggested remedy also needs qualifying: the reference does not bound this at genesis. `Url` derives `FromJSON` from `Text` in `BaseTypes.hs`, and `validateGenesis` in `Shelley/Genesis.hs` checks only epoch length, KES evolutions and quorum size. So validating genesis metadata is stricter than `cardano-node`, in the reject-a-valid-genesis direction.

Details and the reproduction are in the review comment on `ledger/shelley/genesis.go:282-289`. Note this thread was opened against `f42a673`; the commit it discusses is not this head.

ledger/shelley/rules.go
```diff
@@ -941,10 +942,18 @@ func validatePoolRegistration(
 	ls common.LedgerState,
 	networkId uint,
 	minPoolCost uint64,
+	protocolMajor uint,
 	checkNetworkId bool,
 	checkVrfKeys bool,
 	inTxVrfKeys map[common.VrfKeyHash]common.PoolKeyHash,
 ) error {
+	if err := common.ValidatePoolMetadataForProtocolVersion(
```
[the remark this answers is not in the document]
chrisguiney: Addressed by `9a5734e`, which added the bullet at `ledger/shelley/rules.go:822-823` — but the wording it introduced is wrong. `PoolMetadataURLTooLong` is not a predicate in `Shelley/Rules/Pool.hs`; `ShelleyPoolPredFailure` has no URL constructor and `poolTransition` raises no URL failure. The reference enforces the bound in `DecCBOR Url` in `BaseTypes.hs`, at decode. Since the enclosing comment says the predicate list is "all from that file", the new entry now claims provenance it does not have. Follow-up in the review comment on those lines.

arepala-uml (dismissed):

PR #2210, opened 2026-09-04
fix: support 32-bit genesis decoding

### LINKED ISSUE #2185: CI has no 32-bit architecture job
gouroboros CI runs only the host architecture, so a test that is vacuous on 32-bit passes unnoticed.

Concrete case from #2165: `saturating eMax leaves the bound vacuous` failed under `GOARCH=386` because `^uint(0)` does not overflow a uint64 sum on 32-bit, so the clamp under test was never reached. The test was green on amd64 while proving nothing there.

`GOARCH=386 go test ./ledger/` needs no emulation — it compiles and runs on an amd64 runner.

Acceptance criteria: a CI job runs `GOARCH=386 go test` over at least `./ledger/...`; the job is advisory or required per maintainer preference.

wolf31o2:
Fixes #2185

Use fixed-width integers for Byron genesis thresholds, Alonzo execution-unit budgets, and local-state-query system-start days so valid protocol values decode on 386. Keep CBOR array-header and fee-overflow coverage architecture-correct, and run the root test suite on amd64 and 386.

Validation:

- Full offline `GOARCH=386` suite passed with `GOMAXPROCS=1`.
- Native full tests and focused native race tests passed.
- Vet, formatting, diff, and workflow syntax checks passed.
- The default parallel 386 suite exposed one unrelated `protocol/blockfetch` `TestClientStartStopStart` timeout/goroutine-leak failure; the serialized rerun passed.


arepala-uml (dismissed):
Verified the fix directly: reproduced the pre-fix 386 compile failures (`GOOS=linux GOARCH=386 go vet ./...` on the base commit) across Byron genesis thresholds, Alonzo exec-unit budgets, `localstatequery.SystemStartResult.Day`, and the cbor/rules_test overflow cases — all match the PR description. Confirmed those failures are gone on the PR head, and the `SystemStartResult.Day` widening is backed by a real cardano-blueprint conformance vector (`getSystemStart/result.cbor`), not a speculative change. Full `go test ./...` passes natively; CI's 386 matrix legs are green.

`SystemStartResult.Day` moving from `int` to `int64` breaks `dingo/ledger/queries.go`'s `Day: utc.YearDay()` call once dingo bumps its gouroboros dependency (verified via a local module replace build). This is already tracked by a separate dingo PR, so no action needed here.

No behavioral or security defects. CodeRabbit and cubic both reported no actionable findings.

arepala-uml (approved):
Re-reviewed after the branch was updated to 4379801a (rebase resolving main conflicts, which dismissed the prior review). Confirmed the only content change since the previously verified commit (87e47a8b) is the new assertion in cbor/stream_decoder_test.go covering math.MaxInt on 32-bit vs 64-bit: expects ArrayHeaderSize == 9 on 64-bit (MaxInt >= 2^32) and == 5 on 32-bit (MaxInt < 2^32), which matches ArrayHeaderSize's documented 2^32 boundary. All other files in the PR diff are byte-identical to the previously reviewed commit. CI is green on amd64 and 386. CodeRabbit is rate-limited on this PR; cubic's completed review reported no issues, which is sufficient per project policy. No behavioral or security defects.

PR #2212, opened 2026-09-04
fix(ledger): measure a transaction the same way for maxTxSize and fees

chrisguiney:
## Why

Every era's `maxTxSize` rule measured a transaction by the raw length of its
reconstructed CBOR. The fee rules measured it one byte shorter, and the shorter
one is what the chain uses.

An Alonzo-and-later transaction is stored in a block as four parallel arrays —
bodies, witness sets, `IsValid` flags, auxiliary data. The `IsValid` flag is not
part of the transaction the chain sized. Reconstructing a standalone
transaction re-attaches it as the third element of a four-element array, one
byte the original never had. `common.TxSizeForFee` already subtracted it;
`UtxoValidateMaxTxSizeUtxo` did not.

A transaction sitting exactly on `maxTxSize` was therefore rejected as one byte
over — the node rejects a block the chain accepted and stops following the
chain.

## Evidence

A from-genesis Preview replay wedged at slot 27745494 (epoch 321, Conway),
reported as blinklabs-io/dingo#3928:

```
transaction size too large: size 16385, max 16384
```

Transaction `ddd3f6b502c45d7c9f60f75c79148b70c0bae6e238aafafa1d24e9b7ed228d85`,
CBOR fetched from Koios and hash-verified against the decoded bytes:

```
first byte      : 0x84  (four-element array)
raw CBOR length : 16385
length - 1      : 16384
maxTxSize       : 16384
```

Exactly on the limit, which is why it took 321 epochs to surface — it needs a
transaction deliberately built to fill `maxTxSize`.

## The change

`common.TxSize` is now the one measure, and `TxSizeForFee` delegates to it, so
the fee rule and the size rule cannot drift apart again.

**The rule is duplicated seven times.** Allegra delegates to Shelley; the other
six each computed the size themselves and none used the fee measure. All six
now call `common.TxSize`. This is the third time in this replay that fixing one
era left the others behaving as before, so I went through the whole set rather
than only Conway, where the wedge happened.

Babbage and Conway were additionally worse: they called `cbor.Encode(tx)`
unconditionally instead of reading the stored on-wire bytes, discarding the
original encoding before measuring it. They now use the same measure as the
rest.

## Tests

`ledger/max_tx_size_test.go`, in `package ledger_test` rather than in each
era's package, because the point of the test is that all seven eras agree.

- A transaction that is exactly `maxTxSize` on chain is accepted in each
  Alonzo-and-later era. Fails before the change in alonzo, babbage, conway and
  dijkstra; skipped for the three pre-Alonzo eras, which carry no `IsValid`
  byte.
- A genuinely oversize transaction is still rejected, and the size reported in
  the error is the on-chain measure. This is the discrimination that keeps the
  fix from being "move the limit by one" — it must move the boundary by exactly
  the `IsValid` byte, not disable the rule.
- Pre-Alonzo eras are unchanged: one byte over the limit is over the limit.

I also verified the fix against the real transaction directly: `common.TxSize`
returns 16384 for it and `conway.UtxoValidateMaxTxSizeUtxo` accepts it. That
check used a locally fetched CBOR file and is not committed — a 16KB fixture
did not seem worth carrying when the synthetic cases pin the same boundary.

## Validation

`go test ./...` passes. `make lint` reports 0 issues across the module and both
examples.

---
## Summary by cubic
Fixes maxTxSize validation rejecting valid transactions that sit exactly on the chain limit, which wedged a Preview replay at slot 27745494.

An Alonzo-and-later transaction is stored in a block as four parallel arrays, and the `IsValid` flag isn't part of the transaction the chain sized, but reconstructing it as a standalone transaction re-attaches that byte. The maxTxSize rule measured the reconstructed length one byte longer than the fee rule, so an exact-limit transaction was rejected as one byte over. `common.TxSize` is now the single measure for both rules, and `TxSizeForFee` delegates to it; all seven era copies of the maxTxSize rule use it. Babbage and Conway also stopped re-encoding the transaction and now read its stored on-wire bytes.

**Tests**
- Adds `ledger/max_tx_size_test.go` verifying exact-limit transactions are accepted in every Alonzo-and-later era, oversize transactions are rejected with the on-chain size, and pre-Alonzo behavior is unchanged.
- Fixes the Dijkstra row: block transactions there are three components, so nothing is re-attached and nothing may be subtracted — the adjustment follows the envelope shape, not the era.

<sup>Written for commit 512d55f494a74467018eee3d0d6945f393151231. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2212?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction size validation across all supported ledger eras.
  * Transactions at the exact maximum size are now accepted consistently with fee calculations.
  * Oversized transactions are rejected using the accurate on-chain size in validation errors.
  * Size calculation failures are now reported instead of being silently handled through alternate encoding behavior.
* **Compatibility**
  * Existing transaction-size functionality remains available for integrations.

### DIFF
--- a/ledger/common/rules.go
+++ b/ledger/common/rules.go
@@ -170,30 +170,27 @@ func VerifyTransaction(
 // when the transaction actually has the 4-component Alonzo-style envelope.
 const txTypeAlonzo = 4
 
-// TxSizeForFee returns the fee-relevant transaction size as defined by the
-// Cardano ledger spec. For Alonzo through Conway the on-wire CBOR contains a
-// 4-element array [body, witnesses, isValid, metadata]; the Haskell
-// toCBORForSizeComputation function encodes only 3 elements, so the fee-
-// relevant size is the full on-wire length minus 1 byte (the IsValid field).
-// Dijkstra block transactions use a 3-element envelope, so no adjustment is
-// applied unless an explicitly 4-component transaction is being sized.
+// TxSize returns the size of a transaction as the Cardano ledger spec defines
+// it — the measure used both for fees and for maxTxSize.
 //
-// When the transaction has no stored CBOR (e.g. programmatically constructed),
-// the function falls back to encoding the transaction to compute its size.
-// TxSize returns the size of a transaction as the chain measures it.
-//
-// A transaction stored in a block from Alonzo onwards is split across four
-// parallel arrays — bodies, witness sets, IsValid flags and auxiliary data —
-// and the IsValid flag is not part of the transaction the chain sized.
-// Reconstructing a standalone transaction re-attaches it as the third element
-// of a four-element array, one byte the original never had, so that byte is
-// subtracted here.
+// For Alonzo through Conway the on-wire CBOR is a 4-element array
+// [body, witnesses, isValid, metadata], while the Haskell
+// toCBORForSizeComputation encodes only 3 elements. The IsValid flag is not
+// part of the transaction the chain sized: a block stores bodies, witness
+// sets, IsValid flags and auxiliary data in four parallel arrays, and
+// reconstructing a standalone transaction re-attaches that byte. So the size
+// is the full on-wire length minus 1. Dijkstra block transactions use a
+// 3-element envelope, so no adjustment applies unless an explicitly
+// 4-component transaction is being sized.
 //
 // Every rule expressed against a transaction's size must use this, not
 // len(tx.Cbor()). The two differ by exactly one byte, which only matters for a
 // transaction sitting exactly on a limit — rare enough to survive a long way
 // into a replay before a single transaction built to fill maxTxSize exposes
 // it.
+//
+// When the transaction has no stored CBOR (e.g. programmatically constructed),
+// the function falls back to encoding the transaction to compute its size.
 func TxSize(tx Transaction) (int, error) {
 	cborData := tx.Cbor()
 	if len(cborData) == 0 {
--- a/ledger/common/rules.go
+++ b/ledger/common/rules.go
@@ -180,7 +180,21 @@ const txTypeAlonzo = 4
 //
 // When the transaction has no stored CBOR (e.g. programmatically constructed),
 // the function falls back to encoding the transaction to compute its size.
-func TxSizeForFee(tx Transaction) (int, error) {
+// TxSize returns the size of a transaction as the chain measures it.
+//
+// A transaction stored in a block from Alonzo onwards is split across four
+// parallel arrays — bodies, witness sets, IsValid flags and auxiliary data —
+// and the IsValid flag is not part of the transaction the chain sized.
+// Reconstructing a standalone transaction re-attaches it as the third element
+// of a four-element array, one byte the original never had, so that byte is
+// subtracted here.
+//
+// Every rule expressed against a transaction's size must use this, not
+// len(tx.Cbor()). The two differ by exactly one byte, which only matters for a
+// transaction sitting exactly on a limit — rare enough to survive a long way
+// into a replay before a single transaction built to fill maxTxSize exposes
+// it.
+func TxSize(tx Transaction) (int, error) {
 	cborData := tx.Cbor()
 	if len(cborData) == 0 {
 		// Fallback: encode the transaction to compute its size.
@@ -189,7 +203,7 @@ func TxSizeForFee(tx Transaction) (int, error) {
 		var err error
 		cborData, err = cbor.Encode(tx)
 		if err != nil {
-			return 0, fmt.Errorf("failed to encode transaction for fee size: %w", err)
+			return 0, fmt.Errorf("failed to encode transaction for size: %w", err)
 		}
 	}
 	fullSize := len(cborData)
@@ -197,16 +211,18 @@ func TxSizeForFee(tx Transaction) (int, error) {
 		dec, err := cbor.NewStreamDecoder(cborData)
 		if err == nil {
 			arrayLen, _, _, decodeErr := dec.DecodeArrayHeader()
-			if decodeErr == nil {
-				if arrayLen == 4 {
-					return fullSize - 1, nil
-				}
-				return fullSize, nil
+			if decodeErr == nil && arrayLen == 4 {
+				return fullSize - 1, nil
 			}
 		}
-		return fullSize, nil
 	}
 	return fullSize, nil
+}
+
+// TxSizeForFee returns the fee-relevant size of a transaction. It is the same
+// measure as TxSize; the name is kept for the fee rules that already call it.
+func TxSizeForFee(tx Transaction) (int, error) {
+	return TxSize(tx)
 }
 
 // CalculateMinFee computes the minimum fee for a transaction given its
--- a/ledger/max_tx_size_test.go
+++ b/ledger/max_tx_size_test.go
@@ -20,7 +20,6 @@
 package ledger_test
 
 import (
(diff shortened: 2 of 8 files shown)


ledger/common/rules.go
```diff
@@ -180,7 +180,21 @@ const txTypeAlonzo = 4
 //
 // When the transaction has no stored CBOR (e.g. programmatically constructed),
 // the function falls back to encoding the transaction to compute its size.
-func TxSizeForFee(tx Transaction) (int, error) {
+// TxSize returns the size of a transaction as the chain measures it.
```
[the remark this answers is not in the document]
chrisguiney: Valid, fixed in 3f414e2.

The rename replaced the function but not the doc block above it, so `TxSizeForFee`'s comment stayed put and godoc rendered the two as a single block opening with the wrong name.

Worth flagging that the merge was not only cosmetic: my replacement text had dropped real content the old block carried — the Haskell `toCBORForSizeComputation` reference and the Dijkstra 3-element envelope case. Both are back in `TxSize`'s doc, which is where they belong now that it is the shared measure. `TxSizeForFee` keeps a short comment saying it delegates.

This is the third time in this session I stranded a doc block this way, so I also swept the changed packages for comment blocks whose opening word names a different symbol than the declaration beneath them. The rest of this diff is clean; the other hits in `ledger/` are pre-existing and outside it.

ledger/max_tx_size_test.go
```diff
@@ -0,0 +1,244 @@
...
+				"the reported size must be the on-chain measure, not the "+
+					"reconstructed length")
+		})
+	}
+}
+
+// TestMaxTxSizePreAlonzoUnchanged pins that the adjustment is confined to the
+// eras that actually carry an IsValid flag. A pre-Alonzo transaction one byte
+// over the limit is over the limit.
+func TestMaxTxSizePreAlonzoUnchanged(t *testing.T) {
```
wolf31o2, line 285: The pre-Alonzo eras have no accept-case anywhere in this file. `TestMaxTxSizeUsesTheOnChainMeasure` skips them, and both this test and `TestMaxTxSizeStillRejectsOversize` only assert rejection -- yet shelley, allegra and mary all changed here too (inline computation replaced by `common.TxSize`). A validation change with only rejections is how a false rejection ships.

Missing case, exactly: an `unsegmentedTxCbor(16384)` transaction accepted by shelley, allegra and mary. That is the assertion that would catch `common.TxSize` ever subtracting for a 3-element envelope.

ledger/max_tx_size_test.go
```diff
@@ -0,0 +1,244 @@
...
+				tx := &dijkstra.DijkstraTransaction{}
+				tx.SetCbor(c)
+				return tx
+			},
+			pparams: func(m uint) common.ProtocolParameters {
+				pp := &dijkstra.DijkstraProtocolParameters{}
+				pp.MaxTxSize = m
+				return pp
+			},
+			validate:  dijkstra.UtxoValidateMaxTxSizeUtxo,
```
wolf31o2: The dijkstra row is `segmented: true`, so both tests feed it a `0x84` header and assert the `-1` adjustment. But `DijkstraTransaction.MarshalCBOR` emits `cbor.Encode([]any{t.Body, t.WitnessSet, aux})` -- three elements -- and the doc comment on `TxSize` says Dijkstra block transactions use a 3-element envelope.

Which shape does a Dijkstra transaction actually have once it comes out of a block? If it is 3-element, this row asserts a case that cannot occur and gives false confidence. If it is 4-element for a reason other than `IsValid`, then subtracting one byte is wrong and this row pins the wrong behaviour. Worth settling before merge -- it is the only era here whose envelope is not established by existing chain data.

ledger/max_tx_size_test.go
```diff
@@ -0,0 +1,244 @@
...
+			err := era.validate(tx, 0, ls, era.pparams(testMaxTxSize))
+			require.Error(t, err)
+			var sizeErr shelley.MaxTxSizeUtxoError
+			require.ErrorAs(t, err, &sizeErr)
+			assert.Equal(t, testMaxTxSize+1, sizeErr.TxSize)
+		})
+	}
+}
+
+var _ = bytes.Repeat
```
wolf31o2: Leftover: drop this and the `bytes` import rather than keeping the import alive with a blank assignment.

wolf31o2 (commented):
Confirmed against cardano-ledger: `validateMaxTxSizeUTxO` uses `tx ^. sizeTxF`, `alonzoMinFeeTx` uses the same `sizeTxF`, and `Shelley.Rules.Utxo` even carries the property `tx ^. sizeTxF == sizeTxForFeeCalculation tx`. `sizeAlonzoTxF` measures `toCBORForSizeComputation`, which is `encodeListLen 3 <> body <> wits <> auxData` -- so the on-wire 4-element envelope differs by exactly the list header plus the 1-byte `IsValid` bool, and `-1` is the right adjustment. The two quantities are the same measure. Nothing blocking.

One missing accept-case and one question about the dijkstra row, inline. Branch is on `fab2538`, four commits behind `main`.

ledger/max_tx_size_test.go
chrisguiney: You were right to ask, and the answer changes the test.

A Dijkstra block transaction is **three** components. `newDijkstraTransactionFromCborComponents` accepts 3 or 4, but rejects 4 outright when it comes from a block:

```go
if len(txArray) == 4 && !allowIsValid {
    return nil, errors.New("dijkstra transactions in blocks cannot include is_valid")
}
```

and the block sets `TxIsValid` from its own invalid-transaction index list, not from the transaction body. So nothing is re-attached and nothing may be subtracted. Four components are accepted only for a standalone transaction, which is not what maxTxSize validates.

The row is now `blockTxCarriesIsValid: false` for Dijkstra, and I renamed the field from `segmented` since "segmented" was the thing I had wrong.

Dijkstra also turns out to be the case that separates the two candidate rules — subtract because the era is Alonzo-or-later, versus subtract because the envelope has four components — so it now has a dedicated test asserting the envelope rule, with Dijkstra included.

ledger/max_tx_size_test.go
chrisguiney: Fixed. Accept-cases now run in every era, not just the four that carry an IsValid byte.

Being precise about what they prove, since it differs by era: with the adjustment suppressed in `common.TxSize`, the alonzo/babbage/conway accept-cases FAIL — they need it. The shelley/allegra/mary/dijkstra accept-cases PASS either way, because their block envelope is three components and the fix preserves their behaviour exactly. So for those four the accept-case is a regression guard rather than a proof of this change.

That is still the gap you named: those eras changed from an inline length computation to `common.TxSize` and had no accept-case at all, so nothing would have caught it if the shared helper had shortened them.

ledger/max_tx_size_test.go
chrisguiney: Dropped, along with the `bytes` import.

wolf31o2 (approved):

PR #2213, opened 2026-09-04
fix(alonzo): include datum hash in output ToPlutusData

wcatz:
## Summary
- `AlonzoTransactionOutput.ToPlutusData()` hardcoded an empty datum option (`NoOutputDatum`) regardless of `OutputDatumHash`, while `DatumHash()` on the same type correctly returns it.
- Any Plutus script that inspects a spent or reference input's datum option therefore sees "no datum" for legacy (pre-Babbage, array-encoded) outputs that actually carry a datum hash — causing scripts that branch on datum presence to take the wrong execution path and fail.
- Fixes the datum option to mirror what `BabbageTransactionOutput.ToPlutusData()` already does correctly for the hash-only case: `constructor 1` wrapping the hash bytes when `OutputDatumHash != nil`, `constructor 0` otherwise.

## Root cause context
Root-caused against a real, currently-rejected mainnet transaction (`2d295dbffc2898b14ccc9b359a342305006fdcf6ccd7907b444419520a98f351`, withdrawal script `f3c398920fc488cb995db26f09f7e7320b8538b123f300799bf101d5`, redeemer tag 3 index 0) being incorrectly rejected by [blinklabs-io/dingo#3860](https://github.com/blinklabs-io/dingo/issues/3860). The consuming script inspects one of its spent inputs' datum-option field; that input's underlying UTXO uses the legacy 3-tuple array output encoding, which `ledger.NewTransactionOutputFromCbor` correctly resolves to `*alonzo.AlonzoTransactionOutput`. `DatumHash()` on that value returns the correct hash, but `ToPlutusData()` silently dropped it, causing the script to see `NoOutputDatum` and fail.

Two existing tests in `alonzo_test.go` (`TestAlonzoTransactionOutputToPlutusDataCoinOnly`/`...CoinAssets`) carried a `// TODO: empty datum option` comment acknowledging this gap; this PR implements it and adds a dedicated regression test (`TestAlonzoTransactionOutputToPlutusDataWithDatumHash`) covering the hash-present case.

**Scope correction** (per @wolf31o2's review): this only affects the **PlutusV3** context. `TxInfoV1` renders outputs via `WithOptionDatum`, which builds the 3-field `PV1.TxOut` from `DatumHash()` directly, and `TxInfoV2` renders them via `WithZeroAdaAsset`, which already emits `Constr 1 [hash]` from `DatumHash()`. Only `TxInfoV3.ToPlutusData()` reaches `AlonzoTransactionOutput.ToPlutusData()` (via `toPlutusData`'s reflect slice path), so V1/V2 script contexts are unaffected and unchanged by this fix.

Also per that review, and explicitly out of scope for this PR: the value map in both `AlonzoTransactionOutput.ToPlutusData()` and `BabbageTransactionOutput.ToPlutusData()` omits the ada entry entirely when `OutputAmount.Amount == 0`, while cardano-ledger's `transValue`/`transCoinToValue` always include it. Filing that separately rather than folding it into this fix.

## Test plan
- [x] `go test ./ledger/alonzo/... -run TestAlonzo -v` — all pass, including new `TestAlonzoTransactionOutputToPlutusDataWithDatumHash`
- [x] `go test ./ledger/common/script/... -run TestTxInfoV3AlonzoResolvedInputDatumHash -v` — new V3-level integration test passes with the fix; confirmed it fails against the pre-fix code (reverted locally to check)
- [x] `go test -race ./ledger/...` — full ledger suite passes
- [x] `go vet ./ledger/...` — clean
- [x] `gofmt -s -l` on changed files — clean
- [ ] `golangci-lint` — could not run locally (linker toolchain issue in this environment); relying on CI

---
## Summary by cubic
Fixes `AlonzoTransactionOutput.ToPlutusData()` so it no longer drops the output datum hash. Previously it always emitted an empty datum option (`NoOutputDatum`), so Plutus scripts inspecting legacy (pre-Babbage, array-encoded) outputs with a datum hash saw "no datum" and could take the wrong execution path. Now it emits `constructor 1` wrapping the hash bytes when `OutputDatumHash` is set, and `constructor 0` otherwise, matching `BabbageTransactionOutput` behavior.

<sup>Written for commit 324f44d309fab309702adb6b7a7f6203f7e76dbe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2213?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Transaction outputs now correctly preserve datum hashes when converted to Plutus data.
  - Outputs without datum hashes continue to use an empty datum option.

- **Tests**
  - Added coverage for outputs containing datum hashes.
  - Clarified test descriptions for empty datum options and script references.

### DIFF
--- a/ledger/alonzo/alonzo.go
+++ b/ledger/alonzo/alonzo.go
@@ -570,13 +570,21 @@ func (o AlonzoTransactionOutput) ToPlutusData() data.PlutusData {
 			assetDataMap.Pairs...,
 		)
 	}
+	var datumOptionPd data.PlutusData
+	if o.OutputDatumHash != nil {
+		datumOptionPd = data.NewConstr(
+			1,
+			data.NewByteString(o.OutputDatumHash.Bytes()),
+		)
+	} else {
+		datumOptionPd = data.NewConstr(0)
+	}
 	tmpData := data.NewConstr(
 		0,
 		o.OutputAddress.ToPlutusData(),
 		data.NewMap(valueData),
-		// Empty datum option
-		data.NewConstr(0),
-		// Empty script ref
+		datumOptionPd,
+		// Empty script ref (no era support for script refs pre-Babbage)
 		data.NewConstr(1),
 	)
 	return tmpData
--- a/ledger/alonzo/alonzo_test.go
+++ b/ledger/alonzo/alonzo_test.go
@@ -146,9 +146,9 @@ func TestAlonzoTransactionOutputToPlutusDataCoinOnly(t *testing.T) {
 				},
 			},
 		),
-		// TODO: empty datum option
+		// Empty datum option (no OutputDatumHash set)
 		data.NewConstr(0),
-		// TODO: empty script ref
+		// Empty script ref
 		data.NewConstr(1),
 	)
 	tmpData := testTxOut.ToPlutusData()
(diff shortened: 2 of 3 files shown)


ledger/alonzo/alonzo.go
```diff
@@ -570,13 +570,21 @@ func (o AlonzoTransactionOutput) ToPlutusData() data.PlutusData {
 			assetDataMap.Pairs...,
 		)
 	}
+	var datumOptionPd data.PlutusData
+	if o.OutputDatumHash != nil {
+		datumOptionPd = data.NewConstr(
```
wolf31o2, line 575: Correct, but the description overstates the blast radius. `AlonzoTransactionOutput.ToPlutusData()` is only reachable from the **PlutusV3** context: `TxInfoV1` renders outputs through `WithOptionDatum`, which builds the 3-field `PV1.TxOut` from `v2.DatumHash()` directly, and `TxInfoV2` renders them through `WithZeroAdaAsset`, which already emits `Constr 1 [hash]` from `v.DatumHash()`. Only `TxInfoV3.ToPlutusData()` reaches this method (via `toPlutusData`'s reflect slice path). So no V1 or V2 context gains or changes a field here -- worth saying, since a context change is otherwise a hard fork in miniature.

Separately, and pre-existing: the value map above omits the ada entry entirely when `OutputAmount.Amount == 0`, while `WithZeroAdaAsset` and cardano-ledger's `transValue`/`transCoinToValue` always include it. Same in `BabbageTransactionOutput.ToPlutusData`. Not this PR's problem, but it is the other half of the same divergence between the two renderings.

wolf31o2 (commented):
Verified against cardano-ledger `Babbage.TxInfo.transTxOutV2`: `DatumHash dh -> PV2.OutputDatumHash (transDataHash dh)`, and `upgradeAlonzoTxOut` maps `TxOutCompactDH'` to a `DatumHash`. Constructor 1 wrapping the hash is right. Nothing blocking.

One scope correction and one missing test noted inline.

ledger/alonzo/alonzo_test.go
```diff
@@ -271,6 +271,72 @@ func TestAlonzoTransactionOutputToPlutusDataCoinAssets(t *testing.T) {
 	}
 }
 
+func TestAlonzoTransactionOutputToPlutusDataWithDatumHash(t *testing.T) {
```
wolf31o2, line 274: This unit-tests the changed function itself and fails without the fix, so it is real coverage. What is missing is the level the reported rejection actually lives at: a case that builds a `script.TxInfoV3` whose resolved input is an `*alonzo.AlonzoTransactionOutput` with `OutputDatumHash` set, and asserts the rendered context carries `Constr 1 [hash]` at the datum-option position. That is the only assertion that would catch a future change routing V3 outputs through a different renderer.

ledger/alonzo/alonzo.go
wcatz: Good catch, thanks. Corrected the PR description to scope this to PlutusV3 only (V1/V2 already render the hash correctly via WithOptionDatum/WithZeroAdaAsset), and separately noted the zero-ada value-map omission as out of scope for this fix rather than folding it in.

ledger/alonzo/alonzo_test.go
wcatz: Added TestTxInfoV3AlonzoResolvedInputDatumHash in ledger/common/script/context_test.go: builds a V3 TxInfo via NewTxInfoV3FromTransaction with a resolved *alonzo.AlonzoTransactionOutput input carrying OutputDatumHash, and asserts the rendered output PlutusData carries Constr 1 [hash] at the datum-option position. Confirmed it fails against the pre-fix alonzo.go and passes with it.

wolf31o2 (approved):

PR #2214, opened 2026-09-04
fix(ledger): validate Dijkstra batch withdrawals

### LINKED ISSUE #2181: Dijkstra missing unconditional batch withdrawal balance check
cardano-ledger validates Dijkstra withdrawals twice, and gouroboros implements only one of the two.

- ENTITIES `validateWithdrawals` (`eras/dijkstra/impl/src/Cardano/Ledger/Dijkstra/Rules/Entities.hs:329`, called `:288`) runs inside the `Phase2Valid` branch of `Ledger.hs`. `conway.UtxoValidateWithdrawals` covers this, and being phase-2 gated is correct.
- UTXO `validateBatchWithdrawals` (`.../Rules/Utxo.hs:265`, called `:400` from `dijkstraUtxoTransition`) runs unconditionally. It sums each account's withdrawals across the whole sub-transaction batch and compares against the original balance, reporting `WithdrawalsExceedAccountBalance` (`DijkstraUtxoPredFailure`, `Utxo.hs:172`).

gouroboros has no always-run counterpart for the second, so a phase-2-invalid Dijkstra transaction that over-withdraws is accepted. The fix is a new rule in Dijkstra's always-run group, not a change to `conway.UtxoValidateWithdrawals`.

Pre-existing: `ledger/conway/rules.go` already returned early on `!tx.IsValid()` before #2140.

wolf31o2:
Fixes #2181

- Aggregate top-level and subtransaction withdrawals per reward account.
- Reject totals above the original balance unconditionally, including phase-2-invalid transactions.
- Preserve existing per-level withdrawal semantics and add regression coverage.

## Summary by CodeRabbit

* **New Features**
  * Added validation for batch withdrawals across transaction and subtransaction bodies.
  * Transactions are rejected when combined withdrawals exceed an account’s balance.
  * Transactions withdrawing from unregistered reward accounts are rejected with a descriptive error.

* **Bug Fixes**
  * Improved detection and reporting of excessive aggregate withdrawals.


arepala-uml (dismissed):
Verified independently against head 266b185: `go build ./...`, `go test -race ./...` (full suite, all packages), `make lint` (0 issues), and `golines -l .` (no formatting violations) all pass clean. Targeted re-run of withdrawal/batch-withdrawal/validation-rule-mapping tests confirms the new `TestUtxoValidateBatchWithdrawals` and mapping tests pass.

Matches issue #2181: `UtxoValidateBatchWithdrawals` is a new, unconditional (not phase-2-gated) rule aggregating withdrawals across the top-level body and all sub-transactions, keyed by the address's raw encoded bytes rather than pointer identity, correctly rejecting totals above the original balance including for phase-2-invalid transactions — the actual gap the issue identified relative to the existing phase-2-gated `conway.UtxoValidateWithdrawals`. Rule registration is covered by the existing mapping-mutation test, so a missed wiring would have failed CI.

CodeRabbit's one actionable comment (on `rules.go:1090`, about `RewardAddress` being assigned from encoded bytes instead of the dereferenced `*common.Address`) is stale: it was posted against an earlier commit, and the current head already stores and assigns the dereferenced `common.Address` value, matching the field's declared type. Already addressed.

arepala-uml:
One more item before merge: this branch currently has a merge conflict with `main` (GitHub reports `mergeStateStatus: DIRTY`). @wolf31o2 please rebase/merge main and resolve before this is merged.

arepala-uml (approved):

PR #2216, opened 2026-09-04
fix(script): deduplicate sorted transaction inputs

### LINKED ISSUE #2177: SortInputs does not deduplicate, unlike cardano-ledger's Set TxIn
Redeemer index derivation uses `SortInputs` in `ledger/common/script/context.go:210,323,420`, `ledger/common/script/needed.go:310` and `ledger/conway/rules.go:2808`. It sorts by raw TxId bytes then index — matching Haskell `Ord TxIn` — but does not deduplicate, whereas cardano-ledger indexes a `Set TxIn`. A duplicate input therefore shifts every later spend index.

Currently masked: `UtxoValidateNoDuplicateInputs` is registered earlier in all three slices (babbage:43, conway:62, dijkstra:62) than every consumer. The helper remains a footgun for any future caller ordered before that rule.

Related divergence: pre-Conway cardano-ledger silently collapses duplicate inputs rather than rejecting them, so the masking rule is itself a difference from the reference implementation.

Found while reviewing #2163; pre-existing on main.

Acceptance criteria: `SortInputs` deduplicates, or its doc comment states the precondition and names the rule that establishes it; a test pins the index mapping for a transaction with a duplicated input.

wolf31o2:
Fixes #2177.

`SortInputs` now deduplicates transaction inputs after canonical raw `TxId`/index ordering, preserving the original input slice and transaction serialization. Added direct ordering coverage and a required-redeemer index regression test.

Validation:
- Focused tests and race tests for `ledger/common/script`, `ledger/conway`, and `ledger/dijkstra`
- `go vet`, `gofmt`, and `git diff --check`
- Offline conformance: 315/315 vectors

### DIFF
--- a/ledger/common/script/context.go
+++ b/ledger/common/script/context.go
@@ -577,10 +577,9 @@ func (t TimeRange) ToPlutusData() data.PlutusData {
 	)
 }
 
-// SortInputs returns a sorted copy of the given inputs, ordered by
-// (TxId, Index) in ascending byte order. This matches the canonical
-// ordering required by the Cardano ledger spec for redeemer index
-// mapping.
+// SortInputs returns a sorted, deduplicated copy of the given inputs, ordered
+// by (TxId, Index) in ascending byte order. This matches the canonical
+// ordering required by the Cardano ledger spec for redeemer index mapping.
 func SortInputs(inputs []lcommon.TransactionInput) []lcommon.TransactionInput {
 	ret := make([]lcommon.TransactionInput, len(inputs))
 	copy(ret, inputs)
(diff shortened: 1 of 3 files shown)


arepala-uml (approved):
Verified the fix and its test coverage independently.

- `SortInputs` is a single shared helper (5 files, 9 call sites across eras/versions); the dedup fixes both spend-redeemer index derivation and `TxInfo.Inputs` construction uniformly, matching cardano-ledger's `Set TxIn` semantics.
- Dedup correctness: duplicates always sort adjacent under the `(TxId, Index)` order regardless of `slices.SortFunc`'s lack of stability, so the adjacency the dedup pass relies on always holds.
- Regression-test discipline: reverted only the dedup lines (kept the new tests) and confirmed both new tests fail without the fix (`TestSortInputsDeduplicatesAndPreservesInputSlice`, `TestValidateRequiredRedeemersDeduplicatesInputs`), then restored.
- `go test -v -race -run 'TestSortInputs|TestValidateRequiredRedeemers' ./ledger/common/script/...`: all pass.
- CI green (build, lint, nilaway, CodeQL, DCO, both Go toolchains). Cubic: no issues. CodeRabbit rate-limited (documented, non-blocking per policy).

No behavioral, wire-format, or coverage gaps found.

PR #2217, opened 2026-09-04
fix(ledger): preserve invalid transaction script requirements

### LINKED ISSUE #2176: IsValid=false drops all reference-input script requirements
`common.ValidateScriptWitnesses` returns nil immediately when `!tx.IsValid()`, and `shelley.UtxoValidateBadInputsUtxo` iterates only `tx.Inputs()`. Nothing else covers reference inputs, so for an `IsValid=false` transaction:

- an unresolvable reference input returns nil in Babbage, Conway and Dijkstra (rules selected out of each era production slice, not called in isolation);
- a resolvable reference input carrying a needed-but-unsatisfied native script also returns nil.

The second shape re-opens #2149 for `IsValid=false`. Impact is bounded — `IsValid=false` declares phase-2 failure, so inputs are not consumed and only collateral is taken — but the documented invariant in `ledger/common/script/native.go:65-71` and the Babbage rule comment, that `UtxoValidateBadInputsUtxo` plus `UtxoValidateScriptWitnesses` reject an unresolvable input ahead of native script evaluation, is false for this case.

Found while reviewing #2162; pre-existing on main, not introduced there.

Acceptance criteria: reference inputs are resolved and their needed scripts checked for `IsValid=false` transactions, or the invariant comments are corrected to state the actual coverage; test covering an `IsValid=false` transaction with (a) an unresolvable reference input and (b) a resolvable reference input with an unsatisfied native script.

wolf31o2:
Fixes #2176.

- Preserve UTXOW script requirements when `IsValid=false`, including scripts supplied by reference inputs.
- Keep invalid-transaction state transitions and phase-2 validation behavior unchanged.
- Add Babbage, Conway, and Dijkstra regression coverage for matching and unrelated reference scripts.

Validation: `make test`, `make lint`, and 315/315 conformance vectors pass offline.

## Summary by CodeRabbit

- **Bug Fixes**
  - Script witness requirements are now enforced consistently for transactions marked invalid.
  - Transactions using reference scripts must provide the correct script; unrelated or missing scripts now produce an appropriate validation error.
  - Validation now checks script witnesses, redeemers, and extraneous scripts regardless of phase-2 validity status.
  - Behavior is consistent across Babbage, Conway, and Dijkstra eras.

### DIFF
--- a/ledger/common/reference_native_script_test.go
+++ b/ledger/common/reference_native_script_test.go
@@ -313,6 +313,87 @@ func TestReferenceProvidedNativeScriptsAreValidated(t *testing.T) {
 	}
 }
 
+// UTXOW witness requirements apply before phase-2 validity is interpreted.
+// An invalid transaction may therefore use a native script carried by a
+// reference input, but it may not omit the required script entirely.
+func TestInvalidTransactionsRetainReferenceScriptRequirements(t *testing.T) {
+	vkey := bytes.Repeat([]byte{0x64}, 32)
+	nativeScript := testPubkeyNativeScript(t, vkey)
+	unusedScript := testPubkeyNativeScript(t, bytes.Repeat([]byte{0x65}, 32))
+	scriptAddr := testScriptPaymentAddress(t, nativeScript.Hash())
+	keyAddr := testKeyPaymentAddress(t, vkey)
+
+	spendInput := shelley.NewShelleyTransactionInput(
+		"6666666666666666666666666666666666666666666666666666666666666666",
+		0,
+	)
+	refInput := shelley.NewShelleyTransactionInput(
+		"7777777777777777777777777777777777777777777777777777777777777777",
+		0,
+	)
+
+	eras := []struct {
+		name  string
+		rules []common.UtxoValidationRuleFunc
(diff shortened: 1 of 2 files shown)


arepala-uml (approved):
Verified the fix: reverted the `IsValid` early-return locally and confirmed the new test fails without it and passes with it. The change aligns UTXOW witness-completeness checks with the same IsValid-independent pattern already established in `ValidateRequiredRedeemers` (script/redeemers.go). `collectTransactionScriptRequirements` already returns `ReferenceInputResolutionError` for unresolvable reference inputs regardless of IsValid, so this fix also closes the unresolvable-reference-input half of #2176, though untested directly.

Bot review: CodeRabbit flagged one non-blocking gap — no explicit test for an `IsValid=false` tx with an unresolvable reference input. Confirmed the code path already handles it correctly; this is a coverage gap, not a defect. Cubic found no issues. All CI checks (lint, nilaway, go-test, conformance-adjacent) pass.

No merge blockers. Recommend a quick follow-up test for the unresolvable-reference-input case to fully close #2176's acceptance criteria, but not required for this PR.

PR #2219, opened 2026-09-04
perf(ledger): cache transaction UTxO lookups

### LINKED ISSUE #2167: Resolve the transaction input set once per transaction for script rules
`UtxoValidateScriptWitnesses` and `UtxoValidateNativeScripts` each call `common.collectTransactionScriptRequirements`, which resolves every input and every reference input through `LedgerState.UtxoById` and then sorts the purpose lists. Both rules are in the Allegra, Mary, Alonzo, Babbage, Conway, and Dijkstra `UtxoValidationRules`, so every transaction resolves its full input set twice.

A cheap pre-resolution skip is not sound: a reference script satisfying a native purpose can come from a consumed input as well as a reference input (`getReferenceScripts` in `eras/babbage/impl/src/Cardano/Ledger/Babbage/UTxO.hs` reads `inputs <> referenceInputs`), and a spending purpose depends on the resolved output's address. A guard that tried this was added and removed in blinklabs-io/gouroboros#2129 because it was true for any transaction with inputs.

The remaining option is to compute the requirements once per transaction and share them between the two rules. `common.UtxoValidationRuleFunc` is `(tx, slot, ls, pp)` with no per-transaction context, so this needs either a validation context passed through the rule lists or a cache keyed on the transaction.

Acceptance criteria:
- A transaction validated through a full era rule list resolves each input at most once for script requirement collection.
- No change to which transactions are accepted or rejected; the conformance corpus result is unchanged.
- Coverage that counts `UtxoById` calls for a transaction with both rules registered.

wolf31o2:
Fixes #2167.

`common.VerifyTransaction` resolved the same transaction inputs once per rule. It now wraps the caller's ledger state in a transaction-scoped `UtxoById` cache, created on entry and discarded on return, keyed on the input's transaction id and output index.

The wrapper embeds `LedgerState`, so it satisfies that interface but not the optional capability interfaces a concrete state may also implement. `common.UnwrapLedgerState` recovers the caller's state, and the rules that assert an optional capability call it first. `ledger/capability_unwrap_test.go` enforces that: it derives the capability set from the interfaces in `ledger/common/state.go` that `LedgerState` does not embed, and fails on any assertion to one of them outside `UnwrapLedgerState`.

The cache map is mutex-guarded, so substituting the wrapper does not weaken the concurrency guarantee of the state the caller passed in.

Also adds Babbage script-witness and native-script rule coverage.


ledger/common/rules.go
```diff
@@ -33,14 +33,57 @@ import (
...
+func (s *cachedLedgerState) UnderlyingLedgerState() LedgerState {
+	return s.LedgerState
+}
+
+// UnwrapLedgerState returns the caller's ledger state when validation is
+// running with the transaction-scoped UTxO lookup cache. Rules that inspect
+// optional LedgerState capabilities must use this before type assertions; the
+// cache wrapper preserves UTxO lookup behavior but cannot preserve assertions
+// against arbitrary provider types.
+func UnwrapLedgerState(ledgerState LedgerState) LedgerState {
```
chrisguiney, line 68: **Blocker (behavioral, appears on merge with `main`).** This opt-in unwrap makes every optional-capability assertion in the tree a maintenance obligation, and the class is already incomplete against current `main`.

`ledger/common/pool_deposit.go:58` on `main` is `epochState, ok := ls.(EpochState)`. It was added after this branch's merge base, so it is not in this diff and the merge has no textual conflict — but `PoolRegistrationDepositDue` is called from `UtxoValidateValueNotConservedUtxo` in shelley, mary, alonzo, babbage and conway, all of which run under `VerifyTransaction`. Behind `*cachedLedgerState` the assertion fails, the function returns `false`, and the produced value is one pool deposit short.

Reproduced on a clean trial merge of `cb44e318` and `origin/main`, invoking the same registered rule with the same ledger state, varying only the entry point:

```
--- PASS: TestReviewPoolDepositDirectRule
--- FAIL: TestReviewPoolDepositThroughVerifyTransaction
        value not conserved: consumed 3000000000, produced 2500000000
```

`UnwrapLedgerState(ls).(EpochState)` at that line makes it pass. Please rebase onto `main` and re-audit; the failure mode is silent, and `pool_deposit.go`'s own doc comment records it rejecting a valid block on Preview.

ledger/common/rules.go
```diff
@@ -33,14 +33,57 @@ import (
...
 
+type cachedUtxoLookup struct {
+	utxo Utxo
+	err  error
+}
+
+// cachedLedgerState keeps read-only UTxO lookups transaction-scoped. Several
+// validation rules need the same transaction view; sharing these results
+// avoids resolving each input again as the rule list advances.
+type cachedLedgerState struct {
```
chrisguiney, line 54: **Thread safety.** `lookups` is an unsynchronized map, and `VerifyTransaction` substitutes this wrapper for the state the caller passed in. Before this change a caller whose `LedgerState` was safe for concurrent `UtxoById` kept that property inside a rule; now it does not, and the substitution is invisible to the caller.

A rule that fans input resolution out across goroutines trips it:

```
WARNING: DATA RACE
  runtime.mapaccess1_faststr()
  ledger/common.(*cachedLedgerState).UtxoById()  rules.go:79
  ...
  ledger/common.VerifyTransaction()              rules.go:205
```

No rule in this repository spawns goroutines, so nothing in-tree reaches it today. Concurrent `VerifyTransaction` calls sharing one `LedgerState` are clean under `-race`, as expected. Either take a `sync.Mutex` in `UtxoById` or state the constraint on `UtxoValidationRuleFunc` alongside the unwrap note — right now the doc describes the wrapper without saying it is single-goroutine only, and the failure mode is a `concurrent map writes` fatal, not a degraded result.

ledger/common/rules.go
```diff
@@ -33,14 +33,57 @@ import (
...
+	if cached, ok := ledgerState.(interface {
+		UnderlyingLedgerState() LedgerState
+	}); ok {
+		return cached.UnderlyingLedgerState()
+	}
+	return ledgerState
+}
+
+func (s *cachedLedgerState) UtxoById(input TransactionInput) (Utxo, error) {
+	key := input.String()
```
chrisguiney, line 78: **Measured perf regression from the key.** `input.String()` is `fmt.Sprintf("%s#%d", ...)` over a hex-encoded 32-byte hash, allocated on every probe including hits. `TransactionInput` already exposes `Id() Blake2b256` and `Index() uint32`, so a comparable struct key is allocation-free.

Real mainnet fixture blocks, whole era rule list through `VerifyTransaction`, `-benchtime 200x -count 8 -cpu 1`, mean of 8:

| babbage | ns/op | B/op | allocs/op |
|---|---|---|---|
| base (no cache) | 678,879 | 167,314 | 1,973 |
| this PR | 691,205 | 190,020 | 2,353 |
| `struct{id Blake2b256; idx uint32}` key | 675,144 | 156,705 | 1,753 |

| conway | ns/op | B/op | allocs/op |
|---|---|---|---|
| base | 283,136 | 96,566 | 1,090 |
| this PR | 291,325 | 109,655 | 1,327 |
| struct key | 280,513 | 91,269 | 995 |

Against an in-memory ledger state the PR as written costs ~14% more bytes and ~20% more allocations than no cache at all; the struct key beats base on all three columns. A DB-backed state still wins either way, but the key change is free.

ledger/common/rules_test.go
```diff
@@ -1412,6 +1412,71 @@ func TestVerifyTransaction(t *testing.T) {
...
+		[]common.UtxoValidationRuleFunc{
+			babbage.UtxoValidateScriptWitnesses,
+			babbage.UtxoValidateNativeScripts,
+		},
+	)
+	require.NoError(t, err)
+	require.Equal(t, 1, lookupCount)
+}
+
+func TestVerifyTransactionUnwrapsOptionalLedgerStateCapabilities(t *testing.T) {
```
chrisguiney, line 1454: **Missing contract coverage.** This is the only test that exercises the unwrap, and it uses a locally declared rule. I replaced `UnwrapLedgerState`'s body with `return ledgerState` and ran `./ledger/...`: this test is the sole failure. None of the seven production call sites — `ValidateMIRGenesisQuorum`, `govPurposeRoots`, `UtxoValidateValueNotConservedUtxo`, `UtxoValidateWithdrawals`, `UtxoValidateCertificateDeposits`, `UtxoValidateCommitteeCertificates`, `UtxoValidateUnknownVoters`, `validatePoolRetirement` — is driven through `VerifyTransaction` with a capability-bearing state by any test.

That gap is why the `pool_deposit.go` regression merges green. Some of these degrade loudly, but `UtxoValidateValueNotConservedUtxo` and `govPurposeRoots` degrade silently and change accept/reject. A table over the registered rule lists asserting each optional capability survives `VerifyTransaction` would close the class.

ledger/common/rules.go
```diff
@@ -33,14 +33,57 @@ import (
...
+	return s.LedgerState
+}
+
+// UnwrapLedgerState returns the caller's ledger state when validation is
+// running with the transaction-scoped UTxO lookup cache. Rules that inspect
+// optional LedgerState capabilities must use this before type assertions; the
+// cache wrapper preserves UTxO lookup behavior but cannot preserve assertions
+// against arbitrary provider types.
+func UnwrapLedgerState(ledgerState LedgerState) LedgerState {
+	if cached, ok := ledgerState.(interface {
```
chrisguiney, line 69: Minor: the match is structural, so any caller-supplied state that happens to have an `UnderlyingLedgerState() LedgerState` method is unwrapped too, and only one level is peeled. An unexported marker method, or asserting `*cachedLedgerState` directly, would scope this to the wrapper this package actually creates.

chrisguiney (dismissed):
Reviewed at `cb44e318` against `origin/main` (`22746f6d`) in an isolated worktree. Requesting changes on one merge blocker.

## Bot standing on this head

- **CodeRabbit: did not review this diff.** The green `CodeRabbit` status is not a review — the only CodeRabbit comment, last updated at head, is "Review limit reached / You've used all free OSS reviews for now."
- **Cubic: reviewed the first commit only.** The review record's `original_commit_id` is `1b4f520`; the head run posted no new body. I verified its P2 marker "Addressed in `cb44e31`" — the second commit does add `UnwrapLedgerState` and update seven call sites, so the finding is genuinely fixed *for the tree at this branch's merge base*. It is not fixed against current `main` (blocker below).
- No human review yet.

## Blocker: `EpochState` is lost after merging `main`

The branch is 35 commits behind `main`. `ledger/common/pool_deposit.go` was added on `main` after this branch's merge base and does `ls.(EpochState)` with no `UnwrapLedgerState`. The merge is textually clean and semantically broken.

`PoolRegistrationDepositDue` is called from `UtxoValidateValueNotConservedUtxo` in shelley, mary, alonzo, babbage and conway — all registered rules reached through `VerifyTransaction`. Behind `*cachedLedgerState` the assertion fails, the function returns `false`, and the produced value comes out exactly one pool deposit short.

Trial merge (`cb44e318` + `origin/main`, no conflicts), same rule, same state, only the entry point differs:

```
--- PASS: TestReviewPoolDepositDirectRule
--- FAIL: TestReviewPoolDepositThroughVerifyTransaction
        transaction validation failed (value not conserved:
        consumed 3000000000, produced 2500000000)
```

That is the shape `pool_deposit.go` was written to fix (blinklabs-io/dingo#3908): a valid re-registration of a retired pool rejected, one deposit off. Changing line 58 to `UnwrapLedgerState(ls).(EpochState)` makes it pass. Please rebase and re-audit the class, since the whole `./ledger/...` suite stays green with the defect present.

## Verified correct

- The cache is created inside each `VerifyTransaction` call and never escapes it. No rollback, re-application or competing-fork staleness is reachable, and it is not process-global. Size is bounded by one transaction's distinct inputs and freed on return, so there is no mempool or chainsync memory growth.
- Negative results are cached, but only inside one transaction's fixed UTxO view.
- Pure memoization on real data: I ran all 51 transactions from the mainnet Shelley/Allegra/Mary/Alonzo/Babbage/Conway fixture blocks through `VerifyTransaction` with a deterministic lookup oracle that also returns not-found for some inputs. Per-transaction validation outcomes and the set of distinct inputs resolved are identical to base; the lookup count drops (e.g. 50 to 5, 26 to 4, 4 to 1).
- Whole-suite outcome diff base vs head: 4103 to 4105 test results, differing only by the two new tests. Conformance passes on head and on the trial merge.
- `TestVerifyTransactionCachesUtxoLookupsAcrossScriptRules` fails with the wrapping reverted in place (got 2, want 1). `TestVerifyTransactionUnwrapsOptionalLedgerStateCapabilities` passes on base, so it is a guard, but it does fail when `UnwrapLedgerState` is neutered.
- The `reflect` nil guard preserves the prior nil-interface and typed-nil-pointer behavior.

## Scope note

The cache lives only in `common.VerifyTransaction`. dingo (on gouroboros v0.193.3) iterates the era rule slices itself in `ledger/eras/*.go` and calls `ledger.VerifyBlock` with `SkipTransactionValidation: true`, so it never reaches this path. #2167's acceptance criterion is met for `VerifyTransaction` callers only.

## Evidence

| check | exit |
|---|---|
| `go build ./...` (head) | 0 |
| `go test -race ./...` (head) | 0 |
| `go test ./internal/test/conformance/...` (head) | 0 |
| `golangci-lint run ./...` (head, isolated cache) | 0 |
| `go build ./... && go test ./...` in all 8 `examples/*` modules | 0 |
| `go build ./...`, `go test -race ./...`, `golangci-lint run ./...`, conformance (trial merge with `main`) | 0 |
| pool-deposit repro through `VerifyTransaction` (trial merge) | 1 (fails) |

Not run: nilaway (needs `go run ...@latest`; CI green at head), Antithesis, live devnet.

ledger/common/rules.go
```diff
@@ -148,6 +187,14 @@ func VerifyTransaction(
 	protocolParams ProtocolParameters,
 	validationRules []UtxoValidationRuleFunc,
 ) error {
+	if ledgerState != nil &&
+		(reflect.ValueOf(ledgerState).Kind() != reflect.Pointer ||
+			!reflect.ValueOf(ledgerState).IsNil()) {
+		ledgerState = &cachedLedgerState{
```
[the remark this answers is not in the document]
chrisguiney: Verified this marker rather than taking it. The finding was real, and `cb44e31` does address it for the tree at this branch's merge base: `UnwrapLedgerState` plus seven updated call sites, and neutering that helper fails `TestVerifyTransactionUnwrapsOptionalLedgerStateCapabilities`.

It is not complete against current `origin/main`. `ledger/common/pool_deposit.go:58` landed after this branch's merge base and still does `ls.(EpochState)`, so `PoolRegistrationDepositDue` loses the capability behind `*cachedLedgerState` and `UtxoValidateValueNotConservedUtxo` comes out one pool deposit short. Reproduced on a clean trial merge; details in my review.

ledger/common/rules.go
arepala-uml, line 54: cachedLedgerState.lookups is an unsynchronized map, and UtxoById is part of the public LedgerState interface that this wrapper substitutes transparently. No current rule fans input resolution out across goroutines (grepped ledger/ for go func/errgroup: none), so this is not a live race today, but any future rule that parallelizes UtxoById calls -- here or in a downstream caller -- will race on this map with no signal that the cache introduced it. A sync.Mutex around lookups is a two-line fix with negligible cost given VerifyTransaction already runs rules sequentially per call.

arepala-uml (changes requested):
Verified independently against merge-base 71f40b8: all 8 optional-capability assertions (`pool_deposit.go`, `common/rules.go` ValidateMIRGenesisQuorum, 6 in conway/shelley rules.go) now go through `UnwrapLedgerState`, confirmed with build + `go test ./ledger/... -race`. Requesting changes on one item below before approval.

ledger/common/rules.go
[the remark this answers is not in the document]
wolf31o2: Both points are resolved at `15ed707`.

The `cb44e31` fix stands: `UnwrapLedgerState` plus the updated call sites. `pool_deposit.go` was corrected at `e8fa9e6`, and `origin/main` is merged at `957c3ee` — `git diff origin/main..HEAD` after the merge is exactly this PR's six files, so nothing from `main` was dropped.

The per-site obligation is now enforced rather than remembered. `ledger/capability_unwrap_test.go` derives the optional-capability set from `ledger/common/state.go` — every interface declared there that `LedgerState` does not transitively embed — and parses every non-test file under `ledger/` for a type assertion to one of them. An assertion not applied to `common.UnwrapLedgerState(...)` fails the test. Restoring `pool_deposit.go:58` to `ls.(EpochState)` in place reproduces the original defect as a failure:

```
--- FAIL: TestOptionalCapabilityAssertionsUnwrapCachedState
    Should be empty, but was [common/pool_deposit.go:58:20: assertion to
    EpochState is not applied to common.UnwrapLedgerState(...)]
```

Because the capability set is derived and not listed, a capability added to `state.go` later extends the guard without an edit.

ledger/common/rules.go
wolf31o2: Both points are resolved at `15ed707`.

The `cb44e31` fix stands: `UnwrapLedgerState` plus the updated call sites. `pool_deposit.go` was corrected at `e8fa9e6`, and `origin/main` is merged at `957c3ee` — `git diff origin/main..HEAD` after the merge is exactly this PR's six files, so nothing from `main` was dropped.

The per-site obligation is now enforced rather than remembered. `ledger/capability_unwrap_test.go` derives the optional-capability set from `ledger/common/state.go` — every interface declared there that `LedgerState` does not transitively embed — and parses every non-test file under `ledger/` for a type assertion to one of them. An assertion not applied to `common.UnwrapLedgerState(...)` fails the test. Restoring `pool_deposit.go:58` to `ls.(EpochState)` in place reproduces the original defect as a failure:

```
--- FAIL: TestOptionalCapabilityAssertionsUnwrapCachedState
    Should be empty, but was [common/pool_deposit.go:58:20: assertion to
    EpochState is not applied to common.UnwrapLedgerState(...)]
```

Because the capability set is derived and not listed, a capability added to `state.go` later extends the guard without an edit.

ledger/common/rules.go
wolf31o2: Fixed at `15ed707`. `cachedLedgerState.mu` guards the map; the wrapped `UtxoById` call runs with the lock released, so a concurrent miss on the same input can reach the wrapped state twice rather than serializing a slow lookup. The UTxO view is fixed for the duration of validation, so both calls observe the same result. The struct comment now states this.

`TestVerifyTransactionCacheIsSafeForConcurrentRules` in `ledger/common/cached_state_test.go` drives 64 goroutines over 32 inputs, each resolved twice so the miss and hit paths overlap. Removing the two lock pairs in place reproduces the original failure under `-race`:

```
WARNING: DATA RACE
  runtime.mapaccess2()
fatal error: concurrent map read and map write
```

With the lock, `-race -count=4` passes.

ledger/common/rules.go
wolf31o2: Fixed at `15ed707`. The key is now `utxoCacheKey{id Blake2b256; index uint32}` built from `input.Id()` and `input.Index()`, so a probe allocates nothing.

Measured on the probe path alone — one rule resolving 16 inputs 8 times through `VerifyTransaction`, `-benchtime 2000x -count 6 -cpu 1`:

| key | ns/op | B/op | allocs/op |
|---|---|---|---|
| `input.String()` | 72,776-74,910 | 32,857 | 520 |
| `utxoCacheKey` | 11,340-12,519 | 5,528 | 8 |

That isolates the formatting cost rather than amortizing it over a full rule list, which is why the ratio is larger than the whole-era figures above; the direction and the allocation counts agree.

ledger/common/rules_test.go
wolf31o2: Addressed at `15ed707` by a source-level guard rather than a table, because a table only covers the rules someone remembered to add to it and the regression this thread describes arrived through a merge.

`ledger/capability_unwrap_test.go` derives the optional-capability set structurally from `ledger/common/state.go` — every interface declared there that `LedgerState` does not transitively embed, currently `CommitteeCredentialState`, `DRepDelegationState`, `EpochState`, `GenesisDelegationState`, `GovPurposeRootsState`, `StakeCredentialDepositState` and `TipState` — then parses every non-test file under `ledger/` and fails on any type assertion to one of them that is not applied to `common.UnwrapLedgerState(...)`. It covers rules that no test drives through `VerifyTransaction`, including future ones, and a capability added to `state.go` later is picked up without an edit.

Reverting `pool_deposit.go:58` to `ls.(EpochState)` in place fails it, naming the file and line.

`TestVerifyTransactionUnwrapRestoresEveryOptionalCapability` in `ledger/common/cached_state_test.go` carries the behavioral half: a state implementing all seven capabilities is passed through `VerifyTransaction`, and inside the rule each capability is asserted twice. Every direct assertion must fail and every unwrapped one must succeed. The direct half keeps the test from going vacuous — if the wrapper ever satisfied a capability by accident, that assertion would pass and the test would fail.

ledger/common/rules.go
wolf31o2: Fixed at `15ed707`. `UnwrapLedgerState` now asserts `*cachedLedgerState` directly and the exported `UnderlyingLedgerState` method is gone, so the match is to the one type this package wraps with and the structural-collision and multi-level questions do not arise.

`TestVerifyTransactionUnwrapLeavesForeignStateAlone` passes a caller-supplied state that does carry an `UnderlyingLedgerState() LedgerState` method and asserts it is returned unchanged.

ledger/common/rules.go
wolf31o2: Fixed at `15ed707`. `cachedLedgerState` now holds a `sync.Mutex` around the lookup map. The lock is released across the wrapped `UtxoById` call, so a concurrent miss can reach the underlying state twice rather than serializing a slow lookup behind one goroutine; the UTxO view is fixed for the duration of validation, so both callers observe the same result, and the struct comment records that.

`TestVerifyTransactionCacheIsSafeForConcurrentRules` covers it with a rule that fans 64 goroutines over 32 inputs. Removing the lock in place gives `fatal error: concurrent map read and map write` under `-race`.

ledger/common/rules.go
```diff
@@ -28,19 +28,79 @@ import (
...
+	key := utxoCacheKey{id: input.Id(), index: input.Index()}
+	s.mu.Lock()
+	result, ok := s.lookups[key]
+	s.mu.Unlock()
+	if ok {
+		return result.utxo, result.err
+	}
+	utxo, err := s.LedgerState.UtxoById(input)
+	s.mu.Lock()
+	s.lookups[key] = cachedUtxoLookup{utxo: utxo, err: err}
```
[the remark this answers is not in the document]
wolf31o2: Not a defect against the stated contract, so no in-flight coalescing added.

#2167's acceptance criterion is that a transaction validated through a full era rule list resolves each input at most once. `VerifyTransaction` runs its rules sequentially in a single loop, so that holds, and `TestVerifyTransactionCachesUtxoLookupsAcrossScriptRules` is the call-counting coverage the issue asked for: it registers both Babbage script rules and asserts the wrapped state saw exactly one `UtxoById` call.

The duplicate-miss window needs two goroutines inside one rule. Nothing in the tree creates one: `ledger/` contains no `go func` outside `_test.go`, and every `UtxoById` call site is a sequential loop. The concurrent test is the only thing that reaches the window, and it fans out deliberately; its purpose is to show the map is safe for a caller that resolves from several goroutines, not to count calls.

In that window both callers read the same transaction-scoped view and record the same result, so coalescing would remove a repeated lookup rather than change an outcome. The cache is created in `VerifyTransaction`, never escapes it, and is discarded on return, so the view cannot move underneath the two callers. That trade-off is documented on `cachedLedgerState`.

arepala-uml (approved):

PR #2222, opened 2026-09-05
fix(conway)!: bound committee terms at ratification and widen expiry epochs to uint64

### LINKED ISSUE #2144: ledger: enforce constitutional committee maximum term length
## Summary

Conway committee certificate validation does not enforce the protocol parameter limiting a constitutional committee member term. A term that exceeds `CommitteeMaxTermLength` must be rejected.

## Acceptance criteria

- Reject committee authorization or registration state that exceeds the configured maximum term length.
- Preserve valid terms at the boundary and below it.
- Cover absent, zero, boundary, and over-limit parameter values without panics.
- Add regression tests through the production Conway rule path.
- Run full normal/race, conformance, lint, vet, and diff checks.

## Scope

Keep the change within committee-term validation and its protocol-parameter contract.

wolf31o2:
`UpdateCommitteeGovAction` expiry epochs are bounded by the constitutional
committee maximum term length, mirroring `validCommitteeTerm` in
`Cardano.Ledger.Conway.Rules.Ratify`.

- `conway.ValidateCommitteeTerm(action, pp, currentEpoch)` exposes that bound
  for an epoch-boundary ratification caller. It is deliberately not registered
  as a UTxO validation rule: the reference evaluates the predicate in
  `ratifyTransition` to gate enactment, and `actionWellFormed` in `Gov.hs`,
  the GOV analogue, does not bound committee terms. A transaction proposing an
  over-limit expiry is therefore valid and is simply not ratified, and the
  bound loosens each epoch so such a proposal becomes ratifiable later.
  `TestValidateCommitteeTermIsNotATransactionPredicate` pins that the bound is
  absent from the transaction path.
- The bound is `currentEpoch + maxTermLength` in `uint64`, matching
  `addEpochInterval`, which is `EpochNo (n + fromIntegral m)` on `Word64` and
  wraps. An expiry at the boundary is within the term; one past it is not.
- `common.CommitteeMaxTermLengthProvider` is a new optional
  protocol-parameter capability, implemented nil-receiver safe by Conway and
  Dijkstra parameters. No ledger-state capability is added, so a consumer that
  has not implemented a new method keeps proposing committee updates
  unaffected.
- `CommitteeTermTooLongError` names a committee member chosen in sorted order,
  so the error does not vary with map iteration order across runs.

Breaking: `UpdateCommitteeGovAction.CredEpochs` and the matching
`NewUpdateCommitteeGovAction` parameter widen their map value from `uint` to
`uint64`. The CDDL encodes that value as `epoch = uint .size 8`, so `uint`
could not carry a conforming expiry at or above 2^32 on a 32-bit build.
`CommitteeMember.ExpiryEpoch`, the enacted form of the same quantity, is
already `uint64`.

Related to #2144. That issue asks for the bound on committee certificates;
`AuthCommitteeHotCertificate` and `ResignCommitteeColdCertificate` carry no
expiry epoch, and the reference applies the limit only to the
`UpdateCommittee` governance action, which is what this bounds. It does not
close #2144.


ledger/conway/rules.go
```diff
@@ -995,6 +995,9 @@ func UtxoValidateGovActionWellFormedness(
 				)
 				return ConflictingCommitteeUpdateError{Credentials: conflicting}
 			}
+			if err := validateCommitteeTerms(a, ls, pp); err != nil {
```
chrisguiney, line 998: **Blocker — behavioral / consensus.** This enforces a RATIFY-time predicate as a transaction-validity predicate.

`UtxoValidateGovActionWellFormedness` is registered as a UTxO validation rule (`rules.go:50`, `rules.go:297`, reused by `dijkstra/rules.go:57`). Returning an error here makes the transaction — and therefore any block containing it — invalid.

In cardano-ledger, `validCommitteeTerm` is not a transaction-validity check. It sits in `ratifyTransition` (`Ratify.hs:338`) and gates *enactment*, not admission. `Gov.hs:393-399` `actionWellFormed`, the direct analogue of this function, does not mention committee terms at all. A transaction proposing `UpdateCommittee` with an expiry beyond `currentEpoch + committeeMaxTermLength` is accepted onto mainnet today; it simply is not ratified while the predicate is false.

Two consequences:

1. **Chain split.** gouroboros/dingo would reject a block that the Haskell node accepts.
2. **The rejection is permanent where the reference is temporary.** `validCommitteeTerm` is re-evaluated every epoch against the then-current epoch, so the bound *loosens* over time and an over-limit proposal becomes ratifiable later. Rejecting at submission with the submission epoch is strictly stricter, and irrecoverably so.

There is also a determinism concern that follows from the placement: the verdict depends on the validating node's notion of "current epoch" rather than on anything in the transaction, so mempool admission and block-body re-validation can disagree, and replaying history gives a different answer than live validation unless `ls` is epoch-accurate for the block under validation.

gouroboros has no RATIFY implementation (`grep -rni ratif` over non-test `.go` finds only this comment and the localstatequery query type), so there is nowhere correct to wire this today. Suggested shape: keep `validateCommitteeTerms` but export it as a ratification helper for the consumer's epoch-boundary path and drop this call site, rather than making it a transaction predicate. That also matches the function's own doc comment, which already names `Rules.Ratify` as the source.

ledger/conway/rules.go
```diff
@@ -3919,6 +3922,71 @@ func UtxoValidateCommitteeCertificates(
...
+	}
+	termParams, ok := pp.(common.CommitteeMaxTermLengthProvider)
+	if !ok {
+		return CommitteeTermLimitUnavailableError{}
+	}
+	maxTermLength, ok := termParams.CommitteeMaxTermLength()
+	if !ok {
+		return CommitteeTermLimitUnavailableError{}
+	}
+	epochState, ok := ls.(common.CurrentEpochState)
```
chrisguiney, line 4111: **Blocker — API / compatibility.** This fails closed on an optional capability, which is the opposite of the convention this repository documents for exactly this situation.

`ledger/common/state.go` already carries a load-bearing note on `EpochState`:

> It is deliberately optional and degrading: a ledger state that does not implement it keeps every other POOL predicate and does not get the retirement-epoch bound enforced, so that adopting a gouroboros release containing this rule cannot reject otherwise-valid pool retirements in a consumer that has not implemented the method yet.

Here, a `common.LedgerState` that does not implement `CurrentEpoch()` gets `CurrentEpochStateUnavailableError` and every `UpdateCommittee` proposal is rejected. The same applies to `CommitteeTermLimitUnavailableError` at line 3937.

This is not hypothetical. `ouroboros-mock`'s `MockLedgerState` does not implement `CurrentEpoch()` in v0.17.0 (this PR's pin) or in v0.19.0 (main's pin) — I grepped both module directories. That is why this PR had to wrap the conformance state manager and rewrite `rules_gov_transition_test.go:295` to stop passing `nil`. Two in-repo consumers already broke; downstream implementors of `common.LedgerState` will break silently at runtime on upgrade, with the failure mode being "rejects a valid governance action".

Separately, the capability is a near-duplicate. `common.EpochState.EpochForSlot(slot) (uint64, error)` already exists, and this rule receives `slot` in its signature. Deriving the current epoch from the existing capability and *skipping* the bound when it is absent would both follow the documented convention and avoid adding a second epoch interface. Note also that `CurrentEpoch() uint64` returns no error, unlike every other state capability in that file.

ledger/common/state.go
```diff
@@ -149,6 +149,12 @@ type SlotState interface {
 	TimeToSlot(time.Time) (uint64, error)
 }
 
+// CurrentEpochState is the optional ledger-state capability used by rules
+// whose protocol bounds are expressed relative to the current epoch.
+type CurrentEpochState interface {
+	CurrentEpoch() uint64
```
chrisguiney, line 167: See the blocker on `ledger/conway/rules.go:3945`. This interface is added ~85 lines below `EpochState`, whose doc comment in this same file spells out why a current-epoch capability must degrade rather than fail closed, and which already exposes epoch information via `EpochForSlot(slot)`.

If a distinct capability is genuinely wanted, please state in the doc comment why it does not follow the `EpochState` contract, and match the file's `(value, error)` shape. Otherwise reuse `EpochState`.

internal/test/conformance/conformance_test.go
```diff
@@ -156,6 +156,66 @@ var (
...
+	newEpoch uint64,
+) error {
+	if err := m.committeeStateManager.ProcessEpochBoundary(newEpoch); err != nil {
+		return err
+	}
+	m.currentEpoch = newEpoch
+	return nil
+}
+
+func (m *currentEpochStateManager) GetStateProvider() conformance.StateProvider {
```
chrisguiney, line 66: **Blocker — the branch cannot merge, and this file is why.** `git merge-tree HEAD origin/main` reports a content conflict here.

It is not a textual nuisance. `91f816b` (`build(deps): adopt ouroboros-mock v0.19.0`, #2223) landed on main and **deleted** both types this new wrapper embeds — `committeeStateManager` and `legacyCommitteeStateProvider` are gone, replaced by `TestStateProviderExposesCommitteeCredentials`, because v0.19.0's provider implements `common.CommitteeCredentialState` directly. Main also moved the pin from v0.17.0 to v0.19.0; `go.mod:9` here is still v0.17.0.

So after a rebase this whole block does not compile, and the CodeRabbit and cubic findings on line 201 (panic on a non-legacy provider) become moot — the type they are about no longer exists. Please rebase before addressing them.

Worth flagging for the rebase: v0.19.0's provider still has no `CurrentEpoch()` method (I grepped `ledger/` and `conformance/` in the v0.19.0 module). With the fail-closed rule at `rules.go:3945`, any conformance vector containing an `UpdateCommittee` proposal will be rejected unless a replacement wrapper supplies the epoch. Conformance is green on this head only because v0.17.0 plus this file's wrapper happen to supply it.

ledger/conway/rules.go
```diff
@@ -3919,6 +3922,71 @@ func UtxoValidateCommitteeCertificates(
...
+			a.credential.Credential.Bytes(),
+			b.credential.Credential.Bytes(),
+		)
+	})
+	for _, term := range terms {
+		credential := term.credential
+		// This is equivalent to expiryEpoch > currentEpoch+maxTermLength,
+		// without overflowing when the sum exceeds uint64.
+		expiry := term.expiry
+		if expiry > currentEpoch && expiry-currentEpoch > maxTermLength {
```
chrisguiney, line 4144: Non-blocking, and partly a confirmation: the boundary is right. `Ratify.hs:381` is `all (<= addEpochInterval currentEpoch committeeMaxTermLength)`, so at-boundary is accepted and one past it is rejected, which is what this predicate and the `below/at/over boundary` cases encode.

Two small notes on the overflow handling the PR body advertises:

1. `addEpochInterval` in the reference is `EpochNo (n + fromIntegral m)` on `Word64`, i.e. it **wraps**. The `uint64 addition would overflow` case here (`currentEpoch = MaxUint64-2`, `maxTerm = 5`, `expiry = MaxUint64`) is accepted by this code, whereas the reference computes a wrapped bound of `1` and rejects. Unreachable in practice, but the test asserts a divergence from the reference rather than agreement with it — either match the wrap or say in the test that the case is unreachable and deliberately diverges.
2. `ConwayProtocolParameters.CommitteeTermLimit` is `uint64` while the reference `EpochInterval` is `Word32`. Not introduced here, just noting it as the reason the overflow case is even expressible.

chrisguiney (dismissed):
Blocking. The boundary arithmetic is right; the placement is not.

Verified against IntersectMBO/cardano-ledger master:

- `Ratify.hs:375-381` — `validCommitteeTerm` is `all (<= addEpochInterval currentEpoch committeeMaxTermLength)`, evaluated inside `ratifyTransition` (line 338) against `reCurrentEpoch`. So the inclusive bound and the "measured from the current epoch" reading in this PR are both correct.
- `Gov.hs:393-399` — `actionWellFormed`, the proposal-time analogue of `UtxoValidateGovActionWellFormedness`, checks only `ParameterChange`. Grepping `Gov.hs` for `committeeMaxTermLength|validCommitteeTerm` returns nothing.

That is the problem: this is a RATIFY predicate wired into a per-transaction UTxO rule. Details inline.

**Era class audit — complete.** `UpdateCommittee` reaches only `conway.UtxoValidateGovActionWellFormedness`, registered at `conway/rules.go:50` and `:297` and reused verbatim by `dijkstra/rules.go:57`. Both pparams types implement `CommitteeMaxTermLengthProvider`, and the Dijkstra override is nil-safe. No other era carries the rule, and no sibling era file needs the same fix.

**Fail-before check** (my worktree, head `2dc2424`): reverting only the `validateCommitteeTerms` call at `rules.go:998`, keeping the helper so it still compiles, fails 4 subtests — `TestCommitteeTermLimitProductionRule/over_boundary`, `/zero_limit_with_positive_remaining_term`, and both of `TestCommitteeTermLimitRequiresStateAndParameters`. The new tests do exercise the production path.

**Local checks on `2dc2424`** (worktree off `rev-2222`, isolated lint cache): `go build ./...` 0; `go vet ./...` 0; `gofmt -l .` clean; `golangci-lint run ./ledger/...` 0 issues; `go test -race ./ledger/conway/... ./ledger/common/... ./ledger/dijkstra/...` ok; `go test ./internal/test/conformance/...` ok.

**Merge state:** `CONFLICTING` / `DIRTY` at `2dc2424`, confirmed with `git merge-tree` — content conflict in `internal/test/conformance/conformance_test.go`. Not cosmetic; see inline.

**Bot state on this head:** cubic reviewed `2dc2424` (it marked three of its own threads addressed there). CodeRabbit did **not** — its status check is green with body `Review skipped: incremental reviews are disabled`, and its review covers `8ac7a06` only.

**Bot finding dispositions:**

| Finding | Disposition |
|---|---|
| cubic P2 `rules.go:3929` removal-only proposals fail closed | Already addressed in `2dc2424` (`len(a.CredEpochs) == 0` early return) |
| cubic P2 + CodeRabbit `dijkstra/pparams.go:37` typed-nil receiver | Already addressed in `2dc2424`; the CodeRabbit thread is unresolved but stale |
| cubic P3 `rules.go:3947` map-iteration nondeterminism | Already addressed in `2dc2424` by the `slices.SortFunc` on credentials |
| CodeRabbit + cubic `conformance_test.go:201` panic on non-legacy provider | Real, but moot after rebase — see inline |

Also note: issue #2144 is CLOSED (alongside PR #2148, closed unmerged 2026-08-30), and the PR body carries no issue reference. Worth re-opening #2144 or filing a fresh one so the scope is tracked, especially since this change deliberately reinterprets #2144's acceptance criteria (which name committee *certificates*) as the `UpdateCommittee` gov-action bound — the reinterpretation is correct ledger semantics, but it should be recorded somewhere durable rather than only in a code comment.

No human review exists on this PR yet.

ledger/conway/rules.go
wolf31o2: Fixed. The call site is removed from `UtxoValidateGovActionWellFormedness`. `validCommitteeTerm` is reached only from `ratifyTransition` in the reference, and `actionWellFormed` does not bound committee terms, so the predicate is exported as `conway.ValidateCommitteeTerm(action, pp, currentEpoch)` for an epoch-boundary caller instead. `TestValidateCommitteeTermIsNotATransactionPredicate` asserts that an over-limit proposal passes transaction validation and fails if the rule is rewired; `TestValidateCommitteeTermLoosensWithEpoch` pins that the bound is temporary rather than permanent.

ledger/conway/rules.go
wolf31o2: Fixed by removing the requirement rather than degrading it. `ValidateCommitteeTerm` takes `currentEpoch` as a parameter, so no ledger-state capability is consulted at all: `common.CurrentEpochState` and `CurrentEpochStateUnavailableError` are deleted, and no `LedgerState` implementation has to change to keep proposing committee updates. `EpochForSlot` is not needed either, since an epoch-boundary caller already holds the epoch.

ledger/common/state.go
wolf31o2: Fixed. The interface is removed from `ledger/common/state.go`. No new ledger-state capability is added in its place: the epoch is an explicit parameter of `conway.ValidateCommitteeTerm`, so neither `EpochState` nor a sibling of it is consulted.

internal/test/conformance/conformance_test.go
wolf31o2: Resolved. `origin/main` is merged into the branch and the `ouroboros-mock` pin is v0.19.0. `internal/test/conformance/conformance_test.go` is restored to main's content, since the rule no longer reads the epoch from state and the vectors need no wrapper to supply one.

ledger/conway/rules.go
wolf31o2: Matched the wrap. The bound is computed as `currentEpoch + maxTermLength` in uint64, which wraps exactly as `addEpochInterval` does on `Word64`, so the `MaxUint64-2` case is now rejected here as the reference rejects it. The saturating form is gone. `TestValidateCommitteeTermBound` covers the wrapped bound in both directions and fails when the comparison is rearranged to saturate.

internal/test/conformance/conformance_test.go
```diff
@@ -156,6 +156,66 @@ var (
...
+	state, ok := m.committeeStateManager.GetStateProvider().(*legacyCommitteeStateProvider)
```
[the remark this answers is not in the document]
wolf31o2: Resolved by removal. The committee-term bound is no longer a transaction-validity rule, so the harness needs no current-epoch capability. `internal/test/conformance/conformance_test.go` is back to its `main` content: the wrapper and its type assertion are gone, so the pass-through and unknown-provider paths this names no longer exist.

ledger/dijkstra/pparams.go
```diff
@@ -34,6 +34,8 @@ type DijkstraProtocolParameters struct {
 	QuorumStakeThreshold     *cbor.Rat
 }
 
+var _ common.CommitteeMaxTermLengthProvider = (*DijkstraProtocolParameters)(nil)
```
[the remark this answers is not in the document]
wolf31o2: Confirmed and fixed. `*DijkstraProtocolParameters` declares `CommitteeMaxTermLength` explicitly and returns `(0, false)` for a nil receiver rather than promoting the embedded Conway method through a nil outer pointer. `ledger/dijkstra/pparams_committee_term_test.go` asserts it through the `common.CommitteeMaxTermLengthProvider` interface; with the override removed that test panics.

internal/test/conformance/conformance_test.go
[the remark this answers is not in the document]
wolf31o2: Resolved by removal. The committee-term bound is no longer a transaction-validity rule, so the harness needs no current-epoch capability. `internal/test/conformance/conformance_test.go` is back to its `main` content: the wrapper and its type assertion are gone, so the pass-through and unknown-provider paths this names no longer exist.

ledger/conway/rules.go
```diff
@@ -3919,6 +3922,47 @@ func UtxoValidateCommitteeCertificates(
 	return nil
 }
 
+// validateCommitteeTerms mirrors validCommitteeTerm in
+// Cardano.Ledger.Conway.Rules.Ratify. The bound applies to the expiry epochs
+// proposed by an UpdateCommittee action, not to certificates operating on a
+// committee member that was already proposed by an earlier action.
+func validateCommitteeTerms(
```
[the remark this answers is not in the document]
wolf31o2: Addressed. `ValidateCommitteeTerm` returns nil before any capability lookup when the action proposes no expiry epochs, so a removal-only update is unaffected by an absent parameter. The bound is also no longer reachable from transaction validation, so no fail-closed path can reject a transaction. Covered by `TestValidateCommitteeTermWithoutTermLimit/removal_only_without_parameters`.

ledger/dijkstra/pparams.go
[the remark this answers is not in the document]
wolf31o2: Confirmed and fixed. `*DijkstraProtocolParameters` declares `CommitteeMaxTermLength` explicitly and returns `(0, false)` for a nil receiver rather than promoting the embedded Conway method through a nil outer pointer. `ledger/dijkstra/pparams_committee_term_test.go` asserts it through the `common.CommitteeMaxTermLengthProvider` interface; with the override removed that test panics.

ledger/conway/rules.go
```diff
@@ -3919,6 +3922,47 @@ func UtxoValidateCommitteeCertificates(
...
+	maxTermLength, ok := termParams.CommitteeMaxTermLength()
+	if !ok {
+		return CommitteeTermLimitUnavailableError{}
+	}
+	epochState, ok := ls.(common.CurrentEpochState)
+	if !ok {
+		return CurrentEpochStateUnavailableError{}
+	}
+	currentEpoch := epochState.CurrentEpoch()
+	for credential, expiryEpoch := range a.CredEpochs {
```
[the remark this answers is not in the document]
wolf31o2: Addressed. The proposed terms are collected and sorted by type then hash before the bound is applied, so `CommitteeTermTooLongError` names the same committee member on every run. `TestValidateCommitteeTermReportsDeterministically` asserts that over repeated evaluations and fails when the sort is removed.

arepala-uml (approved):

wolf31o2:
Beep bop boop 🤖

PR #2230, opened 2026-09-05
fix(script): deduplicate txInfoSignatories

wolf31o2:
`signatoriesInfo` sorted required signer hashes without deduplicating.

This PR fixes `txInfoSignatories` for duplicate required signers. Input
deduplication remains tracked by #2216.

Tests: `TestTxInfoSignatoriesDeduplicated` (V1 and V2).

Refs blinklabs-io/dingo#3935

### DIFF
--- a/ledger/common/script/context.go
+++ b/ledger/common/script/context.go
@@ -577,31 +577,32 @@ func (t TimeRange) ToPlutusData() data.PlutusData {
 	)
 }
 
-// SortInputs returns a sorted copy of the given inputs, ordered by
-// (TxId, Index) in ascending byte order. This matches the canonical
-// ordering required by the Cardano ledger spec for redeemer index
-// mapping.
+func compareInputs(a, b lcommon.TransactionInput) int {
+	// Compare TX ID
+	x := bytes.Compare(a.Id().Bytes(), b.Id().Bytes())
+	if x != 0 {
+		return x
+	}
+	// Compare index
+	return cmp.Compare(a.Index(), b.Index())
+}
+
+// SortInputs returns a sorted, deduplicated copy of the given inputs, ordered
+// by (TxId, Index) in ascending byte order. cardano-ledger holds both
+// inputsTxBodyL and referenceInputsTxBodyL as Set TxIn and renders them with
+// Set.toList, so a repeated input contributes one entry to txInfoInputs and
+// txInfoReferenceInputs and does not shift the redeemer index of any later
+// spend.
 func SortInputs(inputs []lcommon.TransactionInput) []lcommon.TransactionInput {
 	ret := make([]lcommon.TransactionInput, len(inputs))
 	copy(ret, inputs)
-	slices.SortFunc(
+	slices.SortFunc(ret, compareInputs)
+	return slices.CompactFunc(
 		ret,
-		func(a, b lcommon.TransactionInput) int {
-			// Compare TX ID
-			x := bytes.Compare(a.Id().Bytes(), b.Id().Bytes())
-			if x != 0 {
-				return x
-			}
-			// Compare index
-			if a.Index() < b.Index() {
-				return -1
-			} else if a.Index() > b.Index() {
-				return 1
-			}
-			return 0
+		func(a, b lcommon.TransactionInput) bool {
+			return compareInputs(a, b) == 0
 		},
 	)
-	return ret
 }
 
 func expandInputs(
@@ -803,6 +804,15 @@ func redeemersInfo(
 	return ret, nil
 }
 
+// signatoriesInfo renders txInfoSignatories. cardano-ledger holds
+// reqSignerHashesTxBodyG as Set (KeyHash 'Witness) and translates it with
+// transTxBodyReqSignerHashes = transKeyHash <$> Set.toList, which both sorts
+// and deduplicates. Every era and every Plutus language version routes through
+// that one function, so a transaction body that repeats a required signer
+// still yields a single signatory. Emitting the repeat makes a validator that
+// walks txInfoSignatories execute more reductions than the block producer's
+// evaluator charged for, so its execution units exceed the declared budget and
+// a canonical block is rejected.
 func signatoriesInfo(
 	requiredSigners []lcommon.Blake2b224,
 ) []lcommon.Blake2b224 {
@@ -814,7 +824,12 @@ func signatoriesInfo(
 			return bytes.Compare(a.Bytes(), b.Bytes())
 		},
 	)
-	return tmp
+	return slices.CompactFunc(
+		tmp,
+		func(a, b lcommon.Blake2b224) bool {
+			return bytes.Equal(a.Bytes(), b.Bytes())
+		},
+	)
 }
 
 func votingInfo(
--- a/ledger/common/script/context.go
+++ b/ledger/common/script/context.go
@@ -577,32 +577,31 @@ func (t TimeRange) ToPlutusData() data.PlutusData {
 	)
 }
 
-func compareInputs(a, b lcommon.TransactionInput) int {
-	// Compare TX ID
-	x := bytes.Compare(a.Id().Bytes(), b.Id().Bytes())
-	if x != 0 {
-		return x
-	}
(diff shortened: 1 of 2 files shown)


ledger/common/script/context_set_semantics_test.go
```diff
@@ -0,0 +1,122 @@
...
+	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
+	"github.com/blinklabs-io/gouroboros/ledger/common/script"
+)
+
+// duplicateRequiredSignerTxHex is a transaction body whose required signer set
+// (key 14) lists the same key hash twice, spending one input and producing no
+// outputs. Preview transaction
+// eab27325c569121613728db87a4d8333ce74cc857bbe4991875ceab8787d0213 at slot
+// 41098839 does the same and was accepted by the network.
+const duplicateRequiredSignerTxHex = "84a4008182582020f9a5a89ed5da223f992427733c6fbe6e44cf4a35f48ead39b1e8366cd92d94010180021a00030d400e82581c910da2e8649cb8b76ad829a233d08e83cade13c551c0c04294fe175a581c910da2e8649cb8b76ad829a233d08e83cade13c551c0c04294fe175aa0f5f6"
```
chrisguiney, line 32: Non-blocking: the fixture pins only the deduplication half of `Set.toList`. Both entries in key 14 are the same hash, so the assertions below hold for any ordering. A third, lexicographically smaller hash would also pin the ascending byte order that `Ord (KeyHash 'Witness)` requires (derived from `Hash ADDRHASH` over `ShortByteString`).

This is the only test in `ledger/common/script` that references `TxInfo.Signatories`, so the sort that `signatoriesInfo` has always performed is otherwise unpinned.

ledger/common/script/context_set_semantics_test.go
```diff
@@ -0,0 +1,122 @@
...
+// `transKeyHash <$> Set.toList (txBody ^. reqSignerHashesTxBodyG)`. Set.toList
+// deduplicates, so a body that repeats a required signer yields one signatory.
+// Every era and every Plutus language version routes through that one function,
+// so one deduplication covers every version. Emitting the repeat makes a
+// validator that walks the list execute extra reductions, which pushes its
+// execution units past the producer-declared budget.
+//
+// Only V1 and V2 are exercised from the wire: Conway encodes reqSignerHashes
+// as a tagged set and its decoder rejects a duplicate member outright, so a
+// V3 TxInfo cannot be reached with a repeated required signer.
```
chrisguiney, line 47: Non-blocking: the conclusion holds, but the reason covers only Conway. `NewTxInfoV3FromTransaction` is also reached from Dijkstra, where required signers come from `dijkstraRequiredSigners`, which concatenates `guards.KeyHashes` with the addr-key-hash members of `guards.Credentials` rather than reading one decoded set. What keeps V3 duplicate-free there is `DijkstraGuards.UnmarshalCBOR`: it populates exactly one of the two fields and calls `CheckForDuplicatesAlways` on it.

Worth naming, because the reference folds that same guard `OSet` into a `Set` (`reqSignerHashesTxBodyG = guardsTxBodyL . to (foldr' insertKeyHash mempty . OSet.toSet)`), and the deduplication added here is what would keep the Go side matching if the guards shape ever carried both fields.

chrisguiney (approved):
**APPROVE.**

### Reference behaviour (cardano-ledger `master`)

- `eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Plutus/TxInfo.hs:327`: `transTxBodyReqSignerHashes txBody = transKeyHash <$> Set.toList (txBody ^. reqSignerHashesTxBodyG)`.
- Babbage `TxInfo.hs:357,398`, Conway `TxInfo.hs:446,488,533` and Dijkstra `TxInfo.hs:348,413,460` all assign `txInfoSignatories` from that one function. PV1/PV2/PV3 in every era share it, so there is no version split and one deduplication covers all three Go call sites.
- Order: `Set.toList` is ascending under `Ord (KeyHash r)`, derived from `Hash ADDRHASH` over `ShortByteString`, i.e. plain byte order. The pre-existing `bytes.Compare` sort already matches, and `slices.CompactFunc` after the sort preserves it.

### Reachability (real chain data)

Preview tx `eab27325...d0213` (block 1764478, epoch 475, `valid_contract: true`) encodes key 14 as an untagged two-element array holding the same hash twice: `0e 82 581c910da2... 581c910da2...`. Alonzo/Babbage `CheckForDuplicates` only rejects tag-258 encodings, so this decodes; Conway's `CheckForDuplicatesAlways` and Dijkstra's guard decoder reject duplicates regardless of tag. The network accepted the block, so the reference decoder collapsed the repeat rather than rejecting it. The divergence was gouroboros-side only.

I ran the fix against that full transaction with its four real resolved UTxOs rather than only the synthetic fixture: the body carries 2 required signers, `NewTxInfoV2FromTransaction` yields 1 signatory, and the V2 script context builds.

### Effect on replay and budget

No hash changes: the script context is not committed to by the transaction (`scriptIntegrityHash` covers redeemers, datums and langviews). It does change the `ScriptContext` for Alonzo/Babbage transactions with a repeated required signer, in the direction of the reference. A shorter list can only reduce the work of a validator that walks `txInfoSignatories`, so this cannot push a transaction the network accepted over its declared budget; it removes a false rejection.

Not verified: the CEK numbers in the first commit message (621 steps, 16 builtin calls, 16359467 CPU, 62240 memory). Reproducing them needs the preview PlutusV2 cost model as of epoch 475; the source I could reach returns a 332-entry array for both PlutusV1 and PlutusV2, which matches no known cost-model length, so a failed reproduction would not have distinguished a wrong harness from a wrong claim. Not load-bearing for the code or the test.

### Class audit: set-valued `TxInfo` fields

`txInfoInputs` and `txInfoReferenceInputs` go through `SortInputs`, which sorts then deduplicates on `main` after #2216; `txInfoData` deduplicates by datum hash in `dataInfo`; `ValidateRequiredVKeyWitnesses` already collects required signers into a map. `txInfoSignatories` was the last member. The class is complete.

### Validation

Isolated worktree off current `origin/main`. The branch is based on `89ca16e`, which predates #2216; `git merge origin/main` is clean and every check below also passes on the merged tree.

| check | exit |
| --- | --- |
| `go build ./...` | 0 |
| `go test ./...` (39 packages) | 0 |
| `go test ./internal/test/conformance/...` | 0 |
| `golangci-lint run ./...` (v2.13.2, isolated cache) | 0, 0 issues |
| `go vet ./ledger/...` | 0 |
| `gofmt -s -l` on both changed files | 0, no output |
| `nilaway ./ledger/common/script/...` | 0, no diagnostics |
| `go build ./...` in each of the 8 `examples/*` modules | 0 |

Fail-before: reverting only the `slices.CompactFunc` call in place fails `TestTxInfoSignatoriesDeduplicated/v1` and `/v2` with `expected 1 signatory, got 2`, and nothing else under `./ledger/...` fails. It is a genuine regression test, not a compile-time guard.

Bot standing on this head: CodeRabbit did not review this diff. Its check is green with the title "Review rate limited" and its only comment is the quota warning. Cubic did review; its one P3 is answered in the thread.

Two non-blocking notes on test comments are inline. Neither blocks merge.

ledger/common/script/context.go
```diff
@@ -803,6 +803,15 @@ func redeemersInfo(
 	return ret, nil
```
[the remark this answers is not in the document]
chrisguiney: Already addressed, and the residual-exposure part is stale.

Scope: the title was narrowed to `fix(script): deduplicate txInfoSignatories` at 13:13:35Z, before this review posted at 13:17:19Z, and the body states that input deduplication is tracked by #2216.

Residual exposure: #2216 merged as `68e3bc1` on `main`. `SortInputs` there sorts and then deduplicates, so a repeated input already collapses in `txInfoInputs` and `txInfoReferenceInputs` and no longer shifts the redeemer index of a later spend. This branch is based on `89ca16e`, which predates that commit; merging current `origin/main` into it is clean and `go test ./...` passes on the merged tree.

No change requested.

PR #2232, opened 2026-09-05
fix(ledger): validate transaction network identifiers

### LINKED ISSUE #2084: ledger: validate transaction network identifiers
Alonzo and Babbage transaction bodies decode the optional network identifier but
do not compare it with the active network during ledger validation.

Acceptance criteria:

- Validate a present transaction network identifier against the active network.
- Preserve the era-defined behavior when the field is absent.
- Cover matching, mismatched, absent, and invalid identifiers.

wolf31o2:
Alonzo and Babbage decoded the optional transaction network identifier but did not validate it against the active ledger network.

- Track key 15 presence, including an explicitly encoded testnet identifier.
- Reject mismatched identifiers in Alonzo and Babbage validation.
- Add matching, absent, mismatched, invalid, and regression coverage.

Fixes #2084


ledger/alonzo/alonzo.go
```diff
@@ -327,7 +327,7 @@ func (b *AlonzoTransactionBody) UnmarshalCBOR(cborData []byte) error {
 		return fmt.Errorf("mint: %w", err)
 	}
 	*b = AlonzoTransactionBody(tmp)
-	if err := b.DecodeValidityIntervalUpperBoundPresence(cborData, b.Ttl); err != nil {
+	if err := b.DecodeTransactionBodyFieldPresence(cborData, b.Ttl, false); err != nil {
```
chrisguiney, line 330: Non-blocking, measured. Switching from `DecodeValidityIntervalUpperBoundPresence` to `DecodeTransactionBodyFieldPresence(..., false)` makes the raw `map[uint]cbor.RawMessage` body scan unconditional. The old helper short-circuited when `Ttl != 0`; the new one hardcodes `currentTreasuryValueNonzero` to `false`, so `if b.validityIntervalUpperBoundPresent && b.currentTreasuryValuePresent { return nil }` can never fire for Alonzo or Babbage.

`BenchmarkCBORDecodeTx -benchtime=2000x -count=5`, this branch trial-merged with `origin/main` against an `origin/main` baseline, on the real mainnet fixtures (both have key 3 set and no key 15):

- Alonzo: 42319 B/op, 324 allocs/op -> 46202 B/op, 337 allocs/op (~23.1us -> ~24.8us)
- Babbage: 17873 B/op, 100 allocs/op -> 19115 B/op, 112 allocs/op (~5.65us -> ~6.3us)
- Mary and Conway are byte-identical controls, so this is the scan and not noise.

Part of the cost is inherent: absent-vs-explicit-zero for key 15 can only be answered from the raw bytes, and these fixtures carry no key 15, so no typed shortcut helps. But materializing every field into a `RawMessage` is heavier than a key-presence probe needs to be, and this is genesis-sync path for two eras. Same line in `ledger/babbage/babbage.go:440`.

ledger/common/tx.go
```diff
@@ -351,6 +368,7 @@ func (b *TransactionBodyBase) DecodeTransactionBodyFieldPresence(
 	if !b.currentTreasuryValuePresent {
 		b.currentTreasuryValuePresent = presence.currentTreasuryValue
 	}
+	b.networkIdPresent = presence.networkId
```
chrisguiney, line 371: Non-blocking. This assignment sits after the early return at the top of the function, so key-15 presence is silently dropped for any body decoded with `currentTreasuryValueNonzero=true` and a nonzero TTL. Measured on a Conway body: `{3:10, 15:0, 21:5}` yields `NetworkIdPresent()==false`, while `{3:10, 15:0}` yields `true`.

Inert today. Conway and Dijkstra carry their own `TxNetworkId *uint8`, neither body satisfies the `NetworkIdPresent()` + `TransactionNetworkId()` pair `EncodeTransactionBodyWithValidityIntervalUpperBound` asserts on, and Alonzo/Babbage pass `false` so their early return never fires. But the bit now lives on the shared base, so the next body type that adopts it inherits a presence flag that is wrong exactly when two unrelated fields happen to be set. Folding the network-id bit into the early-return condition would keep the three fields consistent.

Unlike keys 3 and 21 this one also cannot use the `if !b.x` guard the other two use, since there is no typed pre-set to preserve — worth a comment either way.

ledger/common/tx.go
```diff
@@ -370,7 +388,16 @@ func EncodeTransactionBodyWithValidityIntervalUpperBound(
 	treasuryValue := body.CurrentTreasuryValue()
 	preserveTreasuryZero := TransactionCurrentTreasuryValuePresent(body) &&
 		treasuryValue != nil && treasuryValue.Sign() == 0
-	if !preserveUpperBoundZero && !preserveTreasuryZero {
+	networkIdValue, networkIdPresent := body.(interface {
+		NetworkIdPresent() bool
```
chrisguiney, line 392: Nit: `networkIdPresent` here is the type-assertion ok value, not key-15 presence. Sitting next to `preserveNetworkIdZero` and `networkIdValue.NetworkIdPresent()` it reads as the latter. `hasNetworkId` or plain `ok` would be clearer.

ledger/babbage/rules_test.go
```diff
@@ -135,6 +135,72 @@ func TestUtxoValidateOutsideValidityIntervalUtxo(t *testing.T) {
...
+		assert.IsType(t, common.WrongTransactionNetworkIdError{}, err)
+	})
+	t.Run("invalid identifier is rejected", func(t *testing.T) {
+		err := validate(&babbage.BabbageTransaction{Body: babbage.BabbageTransactionBody{NetworkId: 2}})
+		require.Error(t, err)
+		assert.IsType(t, common.WrongTransactionNetworkIdError{}, err)
+	})
+}
+
+func TestTransactionNetworkIdPresencePreservesEncodingAndHash(t *testing.T) {
```
chrisguiney, line 175: Non-blocking coverage gap: there is no Alonzo counterpart to this test, and the key-15 branch in `EncodeTransactionBodyWithValidityIntervalUpperBound` is shared. I confirmed Alonzo does preserve `network_id = 0` through the constructed-encode path, but nothing pins it — removing that branch fails only this Babbage test.

The hash half is legitimately Babbage-only: `BabbageTransactionBody` is the only transaction body with a shadow `hash` field, so the `SetNetworkIdPresence` override at `ledger/babbage/babbage.go:528` is the complete fix for that class, not a one-off.

ledger/alonzo/rules_test.go
```diff
@@ -134,6 +134,43 @@ func TestUtxoValidateOutsideValidityIntervalUtxo(t *testing.T) {
 	)
 }
 
+func TestUtxoValidateTransactionNetworkId(t *testing.T) {
+	ledgerState := mockledger.NewLedgerStateBuilder().
+		WithNetworkId(common.AddressNetworkMainnet).
```
chrisguiney, line 139: Non-blocking: both era tables only build a mainnet ledger state, so the testnet direction is never exercised — explicit `network_id = 0` accepted and explicit `1` rejected when `ls.NetworkId()` is `common.AddressNetworkTestnet`. That is the direction every testnet node runs, and it is the case where the presence bit does not matter, so a second ledger state would pin the discriminator itself rather than the presence plumbing that the current cases already cover.

chrisguiney (approved):
APPROVE. Non-blocking notes inline; nothing found that blocks merge.

**Reference and era gating.** `network_id` (transaction-body key 15) appears first in the Alonzo CDDL (`eras/alonzo/impl/cddl/data/alonzo.cddl`) and is absent from `mary.cddl`, and `validateWrongNetworkInTxBody` in `Cardano.Ledger.Alonzo.Rules.Utxo` is unconditional with no `ProtVer` gate. Adding the rule to Alonzo and Babbage matches the reference. Conway (`ledger/conway/rules.go:2634`) and Dijkstra (`ledger/dijkstra/rules.go:2751`) already had it, and Shelley/Allegra/Mary correctly have no copy because the field does not exist in those eras, so the per-era class is complete.

**Split risk.**
- Absent key 15 is accepted: `TransactionNetworkId()` returns nil, confirmed by decoding a body with no key 15.
- Rejection is a deterministic phase-1 error from a `UtxoValidationRuleFunc`, registered in `AlwaysUtxoValidationRules` at the same position Conway uses. Nothing on the decode path rejects, and there is no panic or connection kill.
- Comparison is against `ls.NetworkId()`, the same 0/1 discriminator `shelley.UtxoValidateWrongNetwork` already uses; not hardcoded to one network.
- All 18 real mainnet Alonzo and Babbage transactions in `internal/testdata` pass the new rule against a mainnet ledger state, and 315 conformance vectors pass including `Conway/Imp/AlonzoImpSpec/UTXO/Wrong_network_ID`.
- Making the raw presence scan unconditional does not widen the decode-error surface: the typed decoder already rejects non-uint body keys identically on `main` and on this branch.

**Fail-before evidence**, each produced by reverting the change in place on this branch:
- Neutering both validators to `return nil`: 3 of 5 subtests fail per era. The two "accepted" subtests are guards and pass either way.
- Dropping the presence check from `TransactionNetworkId()`: `mismatched identifier is rejected` and `explicit testnet identifier decoded from CBOR is present` fail in both eras.
- Dropping `b.networkIdPresent = presence.networkId`: the CBOR-decoded subtest fails in both eras.
- Dropping the key-15 branch in `EncodeTransactionBodyWithValidityIntervalUpperBound`: `TestTransactionNetworkIdPresencePreservesEncodingAndHash` fails at the key-15 assertion.
- Dropping the Babbage `SetNetworkIdPresence` override: the same test's final hash assertion fails, so the fix for the open cubic thread is covered by a test.
- Dropping the Alonzo descriptor entry: `TestUtxoValidationRuleDescriptors/Alonzo` fails.

**Checks.** The branch is 30 commits behind `main`; the trial merge is clean and everything below ran on the merged tree. `go build ./...` 0. `go test ./...` 1 — `pipeline.TestValidateStage_NoProviderReturnsError`, which reproduces identically on an `origin/main` baseline at `-count=100`, so it is pre-existing and unrelated. `go test -race` on `ledger/{alonzo,babbage,common,conway,dijkstra}` 0. `go vet ./ledger/...` 0. `golangci-lint run ./...` with an isolated cache: 0 issues. `gofmt -l ledger/`: clean. Build and test for all 8 nested `examples/*/go.mod` modules: 0.

**Bot standing on head `3e206de`.** CodeRabbit did not review this diff — its check is green, but the comment body is "Review limit reached ... You've used all free OSS reviews for now". cubic's head run was incremental over 2 files ("0 issues found across 2 files (changes from recent commits)"); its full 11-file run is anchored to the first commit `a4f6cdb`. All four cubic findings across both runs are fixed at head, verified against the checkout; replies posted on each thread.

ledger/common/tx.go
```diff
@@ -307,6 +311,19 @@ func (b *TransactionBodyBase) CurrentTreasuryValuePresent() bool {
 	return b.currentTreasuryValuePresent
 }
 
+// SetNetworkIdPresence records whether transaction-body key 15 is present.
+// Calling it for a programmatically constructed body invalidates stored CBOR.
+func (b *TransactionBodyBase) SetNetworkIdPresence(present bool) {
+	b.networkIdPresent = present
+	b.hash = nil
```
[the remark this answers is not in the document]
chrisguiney: Already addressed at head `3e206de`. `ledger/babbage/babbage.go:528` adds a Babbage-specific `SetNetworkIdPresence` that clears the shadow `hash` before delegating to the base. Verified by reverting that override in place: `TestTransactionNetworkIdPresencePreservesEncodingAndHash` then fails on its final `assert.NotEqual(firstHash, body.Id())`.

Class audited: `BabbageTransactionBody` is the only transaction body in the repo with a shadow `hash` field and the only one overriding `Id()`. Alonzo, Conway, Dijkstra, Mary, Allegra and Shelley bodies all use `TransactionBodyBase.Id()`, so no other era needs the same override. Thread can be resolved.

ledger/common/tx.go
```diff
@@ -307,6 +311,19 @@ func (b *TransactionBodyBase) CurrentTreasuryValuePresent() bool {
 	return b.currentTreasuryValuePresent
 }
 
+// SetNetworkIdPresence records whether transaction-body key 15 is present.
+// Calling it for a programmatically constructed body invalidates stored CBOR.
+func (b *TransactionBodyBase) SetNetworkIdPresence(present bool) {
```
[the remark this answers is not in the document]
chrisguiney: Already addressed at head `3e206de`. `EncodeTransactionBodyWithValidityIntervalUpperBound` now patches key 15 alongside keys 3 and 21. Verified by removing the key-15 branch in place: `TestTransactionNetworkIdPresencePreservesEncodingAndHash` fails at the `assert.True(t, present)` check on key 15.

I also confirmed the Alonzo side works through the same shared branch, though only Babbage has a test pinning it — noted separately in the review.

ledger/common/tx.go
```diff
@@ -370,7 +388,11 @@ func EncodeTransactionBodyWithValidityIntervalUpperBound(
 	treasuryValue := body.CurrentTreasuryValue()
 	preserveTreasuryZero := TransactionCurrentTreasuryValuePresent(body) &&
 		treasuryValue != nil && treasuryValue.Sign() == 0
-	if !preserveUpperBoundZero && !preserveTreasuryZero {
+	networkIdValue, networkIdPresent := body.(interface {
+		NetworkIdPresent() bool
+	})
+	preserveNetworkIdZero := networkIdPresent && networkIdValue.NetworkIdPresent()
```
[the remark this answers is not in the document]
chrisguiney: Already addressed at head `3e206de`, which applies the suggested gate: `preserveNetworkIdZero` is now set only when `TransactionNetworkId()` is non-nil and zero, so a nonzero key 15 no longer forces the map round trip.

ledger/babbage/rules_test.go
```diff
@@ -135,6 +135,64 @@ func TestUtxoValidateOutsideValidityIntervalUtxo(t *testing.T) {
...
+	encoded, err := body.MarshalCBOR()
+	require.NoError(t, err)
+	var fields map[uint]cbor.RawMessage
+	_, err = cbor.Decode(encoded, &fields)
+	require.NoError(t, err)
+	_, present := fields[15]
+	assert.True(t, present)
+
+	body.SetNetworkIdPresence(false)
+	assert.NotEqual(t, firstHash, body.Id())
```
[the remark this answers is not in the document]
chrisguiney: Already addressed at head `3e206de`. The test now re-encodes after `SetNetworkIdPresence(false)`, asserts key 15 is dropped, and calls `body.SetCbor(encoded)` before comparing hashes, so the final assertion is no longer hashing nil. Confirmed it is load-bearing: reverting the Babbage `SetNetworkIdPresence` override makes exactly that assertion fail.

PR #2236, opened 2026-09-05
fix(ledger): decode pool metadata hashes of any length

### LINKED ISSUE #1707: Dijkstra GenTx decoding to match ouroboros-mock upstream fixtures
The `ouroboros-consensus` golden `GenTx_Dijkstra` fixtures (synced into `ouroboros-mock/fixtures` via blinklabs-io/ouroboros-mock#164) currently validate only through payload/body-hash semantics because the imported fixture shape is ahead of full Dijkstra transaction decoding support here.

Bring Dijkstra transaction decoding up to parity so the fixture harness can exercise a full `GenTx` decode path instead of the payload/body-hash fallback.

wolf31o2:
`GenTx_Dijkstra` and the other `CardanoNodeToNodeVersion2` GenTx goldens did not decode. The blocker was not the Dijkstra transaction shape, which `ledger/dijkstra` already accepts, but `pool_metadata_hash`.

## Wire format

A GenTx is `[era_id, #6.24(bytes .cbor transaction)]` and the paired GenTxId is `[era_id, bytes .size 32]`, where the identifier is the Blake2b-256 hash of the transaction body's own CBOR. The era identifier is the ledger transaction type.

| Era | Era id | Transaction array |
|-----|--------|-------------------|
| Shelley | 1 | `[body, witness_set, auxiliary_data]` |
| Allegra | 2 | `[body, witness_set, auxiliary_data]` |
| Mary | 3 | `[body, witness_set, auxiliary_data]` |
| Alonzo | 4 | `[body, witness_set, is_valid, auxiliary_data]` |
| Babbage | 5 | `[body, witness_set, is_valid, auxiliary_data]` |
| Conway | 6 | `[body, witness_set, is_valid, auxiliary_data]` |
| Dijkstra | 7 | `[body, witness_set, auxiliary_data]` |

`pool_metadata_hash` is an unbounded byte string. `pmHash` in `libs/cardano-ledger-core/src/Cardano/Ledger/State/StakePool.hs` is a `ByteArray`, and `PoolMetadata`'s `DecCBOR` instance applies no length check. The 32-byte bound is a POOL rule predicate, `PoolMedataHashTooBig` in `eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs`, gated on `SoftForks.restrictPoolMetadataHash` (`pv > ProtVer 4 0`).

Every GenTx golden from Shelley to Dijkstra carries a pool registration whose metadata hash is the two bytes `7b7d`, so all seven failed with `invalid blake2b-256 hash length: expected 32 bytes, got 2`.

## Changes

- `ledger/common/certs.go`: `PoolMetadataHash` is a byte-string type instead of an alias for `Blake2b256`. It keeps the `Bytes`, `String` and JSON behavior of the previous type; text decoding now accepts any even-length hex payload. It encodes a byte string for a nil value and rejects a payload that is not a CBOR byte string.
- `ledger/common/protocol_version.go`: `PoolMetadataHashRestricted` gates the bound on major protocol version > 4.
- `ledger/shelley/rules.go`, `ledger/shelley/errors.go`: `UtxoValidatePoolCertificates` rejects a metadata hash longer than 32 bytes with `PoolMetadataHashTooBigError`, between `WrongNetworkPOOL` and `StakePoolCostTooLowPOOL` as in the reference. The rule doc comment no longer records the predicate as unreachable.

## Tests

- `internal/test/conformance/gentx_conformance_test.go`: `TestConsensusGenTxFixtures` reads the `GenTx_*` and `GenTxId_*` goldens from the `ouroboros-mock` v0.17.0 embedded upstream fixtures and, for each of the seven eras, checks the envelope era identifier, the transaction array width, that `ledger.NewTransactionFromCbor` decodes it, that the transaction CBOR is preserved and re-encodes to the original bytes, and that `tx.Hash()` equals the paired GenTxId payload.
- `ledger/pool_rules_test.go`: `TestUtxoValidatePoolCertificatesMetadataHashTooBig` covers a 33-byte hash rejected at major version 5, the same hash accepted at major version 4, and 2-byte and 32-byte hashes accepted.
- `ledger/common/pool_reward_account_test.go`: `TestPoolMetadataHashLengthIsFixed` is replaced by `TestPoolMetadataHashDecodesAnyLength`, which decodes 2, 31, 32 and 33-byte hashes, re-encodes them unchanged, and rejects a text string.
- `ledger/shelley/genesis_test.go`: the genesis `extraConfig` metadata-hash case now rejects a non-hex value rather than a short one.

## Downstream

`blinklabs-io/dingo` currently converts a `Blake2b256` to `PoolMetadataHash` in `database/plugin/metadata/sqlstore/pool.go` and `ledger/view.go`. Those call sites need to pass the stored bytes directly before Dingo updates to a gouroboros release containing this change; tracked in dingo#4058.

Closes #1707

---
## Summary by cubic
Fixes #1707 by making pool metadata hash decoding accept any byte-string length, so all `CardanoNodeToNodeVersion2` GenTx goldens from Shelley through Dijkstra decode. `PoolMetadataHash` is now a distinct byte-string type instead of a `Blake2b256` alias, and the 32-byte limit is enforced by the POOL rule only from protocol version 5.

**Tests**
- Adds `TestConsensusGenTxFixtures` for upstream GenTx/GenTxId goldens across all seven eras.
- Adds `TestUtxoValidatePoolCertificatesMetadataHashTooBig` for the version-gated 32-byte bound.

**Migration**
- Replace `common.NewBlake2b256(...)` with `common.PoolMetadataHash(...)` when constructing pool metadata hashes.
- Downstream converters should pass stored bytes directly instead of converting through `Blake2b256`.

<sup>Written for commit a6a33f8240c7fd8ce41e74e2c2a24c29e5856a5b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2236?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Enhancements**
  * Pool metadata hashes can now be decoded and preserved at varying lengths.
  * Protocol version 4 accepts metadata hashes longer than 32 bytes; protocol version 5 and later reject them according to pool-registration rules.
  * Added clearer validation errors when pool metadata hashes exceed the permitted size or use invalid encoding.

* **Documentation**
  * Updated conformance test documentation with GenTx coverage details and run instructions.

* **Tests**
  * Added coverage for transaction decoding across supported eras and pool metadata hash validation.


ledger/pool_rules_test.go
```diff
@@ -1095,3 +1095,107 @@ func TestPoolRulePparamsWiring(t *testing.T) {
 		})
 	}
 }
+
+// poolRegCertWireWithMetadata builds a pool registration certificate carrying a
+// pool_metadata entry whose hash is hashLen bytes long.
+func poolRegCertWireWithMetadata(
```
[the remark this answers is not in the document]
wolf31o2: Fixed in 7ba963c. `decodePoolRegCertWire` now builds the registration array, taking the `pool_metadata` entry as its last argument; `poolRegCertWire` passes nil and `poolRegCertWireWithMetadata` passes the entry.

internal/test/conformance/README.md
```diff
@@ -440,6 +440,33 @@ Tests Byron-era block parsing and validation using real mainnet and testnet bloc
 
 ---
 
+## GenTx Conformance Tests
```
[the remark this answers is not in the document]
wolf31o2: Fixed in 7ba963c. The summary table gains a GenTx Goldens row with 7 cases and the total is 421.

arepala-uml (dismissed):
Verified independently against head 7ba963c: `go build ./...`, `go test -race ./...` (all packages, including new `TestConsensusGenTxFixtures` across all 7 eras and `TestUtxoValidatePoolCertificatesMetadataHashTooBig`), and `make lint` all pass clean. Confirmed `byteStringHeader` rejects non-byte-string CBOR before decode, and confirmed the `PoolMetadataHashRestricted` gating and error-ordering placement in `validatePoolRegistration` match the PR description.

Root cause is accurately diagnosed (fixed 32-byte `PoolMetadataHash` alias vs. the reference ledger's unbounded `pmHash`, with the 32-byte bound actually enforced by a version-gated POOL rule) and fixed at the right layer.

Non-blocking follow-up: `PoolMetadataHash` changes from a `Blake2b256` alias to a distinct `[]byte` type. `blinklabs-io/dingo` currently does `PoolMetadataHash(NewBlake2b256(...))` in two places (`database/plugin/metadata/sqlstore/pool.go`, `ledger/view.go`), which is an illegal array-to-slice conversion in Go and will fail to build once dingo bumps its gouroboros dependency past this change. Not yet tracked in a dingo issue/PR. Recommend opening one for the two-line fix (pass the stored bytes directly, dropping the `NewBlake2b256` wrap) so a routine dependency bump doesn't break dingo's build.

Please make sure to resolve merge conflcts.

ledger/common/protocol_version.go
```diff
@@ -59,3 +59,18 @@ func PoolAccountNetworkIdValidated(major uint) bool {
...
+//
+// Reference: eras/shelley/impl/src/Cardano/Ledger/Shelley/SoftForks.hs,
+// restrictPoolMetadataHash (pv > ProtVer (natVersion @4) 0), consumed by
+// poolTransition in eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs.
+//
+// The reference compares the whole protocol version, so a 4.x version with a
+// non-zero minor would also restrict. Only the major version reaches the POOL
+// rule here, and no 4.x protocol version with a non-zero minor exists, so the
+// comparison is on the major version alone.
+func PoolMetadataHashRestricted(major uint) bool {
```
chrisguiney, line 74: **Blocker (mechanical): this is the merge conflict.**

`main` at `f2b0fcc` appends `MirTransferAllowed` here with the identical `\treturn major > 4\n}` tail, so git aligns the two additions and reports a content conflict. Keeping both functions resolves it.

I resolved it that way in a local worktree and the merged tree is clean: `go build ./...`, `go vet ./...`, `gofmt -l .`, `go test ./...` and `golangci-lint run ./...` (0 issues) all pass.

ledger/common/certs.go
```diff
@@ -410,11 +410,78 @@ func (c *StakeDelegationCertificate) Type() uint {
...
+// libs/cardano-ledger-core/src/Cardano/Ledger/State/StakePool.hs is a
+// ByteArray, and the DecCBOR instance for PoolMetadata applies no length
+// check. The length bound belongs to the POOL rule instead
+// (PoolMedataHashTooBig in
+// eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs, gated on
+// SoftForks.restrictPoolMetadataHash), which rejects only hashes longer than
+// 32 bytes and only from protocol version 5 onwards. A fixed 32-byte type
+// would reject registrations the node accepts, so the bound is applied in
+// shelley.UtxoValidatePoolCertificates.
+type PoolMetadataHash []byte
```
chrisguiney, line 429: Non-blocking, downstream. Dropping the `PoolMetadataHash = Blake2b256` alias is a breaking API change for consumers.

`blinklabs-io/dingo` does `lcommon.PoolMetadataHash(lcommon.NewBlake2b256(...))` at `database/plugin/metadata/sqlstore/pool.go:1566` and `ledger/view.go:262`. Array-to-slice conversion is illegal in Go; I confirmed with a minimal repro:

```
cannot convert NewBlake2b256(...) (value of array type Blake2b256) to type PoolMetadataHash
```

dingo pins gouroboros `v0.193.3`, so a routine dependency bump breaks its build. I searched dingo issues and PRs for `PoolMetadataHash` and found none, so this is still untracked — same point @arepala-uml raised. Worth opening the dingo issue before this merges.

(`api/utxorpc/submit_test.go:900` uses `PoolMetadataHash{}`, which stays valid for a slice type.)

ledger/common/certs.go
```diff
@@ -410,11 +410,78 @@ func (c *StakeDelegationCertificate) Type() uint {
...
+
+func (h PoolMetadataHash) MarshalJSON() ([]byte, error) {
+	return json.Marshal(h.String())
+}
+
+func (h PoolMetadataHash) MarshalText() ([]byte, error) {
+	return []byte(h.String()), nil
+}
+
+func (h *PoolMetadataHash) UnmarshalText(text []byte) error {
```
chrisguiney, line 449: Non-blocking, description accuracy. The PR body says the new type "keeps the `Bytes`, `String`, JSON and text behavior of the previous type". The text behavior does change: `Blake2b256.UnmarshalText` rejects anything that is not 32 bytes (`ledger/common/common.go:76`), whereas this accepts any even-length hex including the empty string.

That is consistent with the decode change and the `genesis_test.go` update reflects it, but the description overstates the preservation. `ToPlutusData()` is also no longer available on the type; nothing in-tree uses it.

chrisguiney (commented):
Reviewed at head `77b494a`. **Not approving yet**: the branch is `CONFLICTING` against `main` (`f2b0fcc`), and neither bot actually reviewed this head. The change itself is correct and genuinely tested. I'll re-review once the conflict is resolved.

## Checks I ran (worktree on `77b494a`)

| Check | Result |
|---|---|
| `go build ./...` / `go vet ./...` / `gofmt -l .` | clean |
| `go test ./ledger/... ./internal/test/conformance/...` | pass |
| `golangci-lint run ./...` (isolated cache) | 0 issues |
| `git merge-tree rev-2236 origin/main` | 1 conflict, `ledger/common/protocol_version.go` |
| merged tree, conflict resolved | build, vet, gofmt, `go test ./...`, lint all clean |

## Fail-before

Both halves of the fix reverted in place:

- Re-imposing the 32-byte bound in `PoolMetadataHash.UnmarshalCBOR` fails 6 of 7 `TestConsensusGenTxFixtures` era cases, 3 subtests of `TestPoolMetadataHashDecodesAnyLength`, and `TestUtxoValidatePoolCertificatesMetadataHashTooBig`.
- Disabling the `checkMetadataHash` predicate in `validatePoolRegistration` fails `TestUtxoValidatePoolCertificatesMetadataHashTooBig/oversized_hash_is_rejected_from_major_version_5`.

Neither half is untested.

## Class audit — complete

- `PoolMetadataHash` has exactly one field of that type, `PoolMetadata.Hash` (`ledger/common/certs.go:546`). Certificates live in `ledger/common` and are not duplicated per era, so the decode change is era-wide by construction.
- The POOL rule has one implementation, `shelley.validatePoolRegistration`. Shelley, Allegra, Mary, Alonzo, Babbage and Conway route to `shelley.UtxoValidatePoolCertificates` directly; Dijkstra goes via `conway.UtxoValidatePoolCertificates`. All six register it in their rule tables, so the new predicate is reachable from every era. No sibling era left unfixed.

## Wire and semantics

Acceptance is not widened beyond the reference. At major version >= 5 the 32-byte bound is restored as a rule predicate, so the accepted set is unchanged. At <= 4 it widens, which is the correct direction: the reference accepts oversized hashes before `restrictPoolMetadataHash`, so the previous fixed-32-byte type was *stricter* than the node.

Raw bytes are preserved. `TestConsensusGenTxFixtures` asserts `tx.Cbor()` equals the golden bytes, `cbor.Encode(tx)` re-encodes to them, and `tx.Hash()` matches the paired `GenTxId` payload, for all seven eras. `Closes #1707` matches what the change does.

Overlap with #2209 (pool metadata URL bounds): not a semantic contradiction — the reference bounds the URL in the decoder and the hash in the POOL rule, which is what the two PRs do respectively. They will conflict textually in `ledger/common/certs.go`, and #2209 adds a `PoolMetadata.MarshalCBOR` that re-encodes `[]any{p.Url, p.Hash}`, so whichever lands second should re-run the GenTx round-trip assertions.

## Bot and human review status on this head

Neither bot reviewed `77b494a`.

- **CodeRabbit**: status reads `pass`, but the comment body is `Review limit reached — next included review available in 38 minutes`. That is not a review.
- **cubic**: check is `success` with `output.summary` = "AI review not needed for merge commit ... cubic will review your next non-merge commit". Its actual review was of `c65f53b`. Both P3 findings (README GenTx row + total 421; shared `decodePoolRegCertWire` builder) are addressed in `7ba963c` — verified in the diff.
- **@arepala-uml**'s approval was dismissed by the `77b494a` push, so there is no standing human approval.

All 13 check runs on `77b494a` are `success` (read from the check-runs API, not the summary line). No failing job.

## Next action

Resolve the `protocol_version.go` conflict (keep both functions) and push. That will also re-trigger cubic on a non-merge commit. Owner: @wolf31o2.

wolf31o2:
### Review follow-up

- The two downstream Dingo conversions are tracked in [Dingo #4058](https://github.com/blinklabs-io/dingo/issues/4058). They must pass stored bytes directly before adopting this API change.
- The PR description now accurately documents the changed text-decoding behavior.
- The requested test-helper and conformance README updates were included in `7ba963c`. The remaining code blocker, the `protocol_version.go` conflict, was resolved in `a6a33f8`.

Please re-review the current head.

arepala-uml (approved):
Re-reviewed at head a6a33f8. Merge conflict in ledger/common/protocol_version.go is resolved correctly (both PoolMetadataHashRestricted and MirTransferAllowed present).

Verified independently in an isolated worktree:
- go build ./..., gofmt -l ., make lint (0 issues) all clean
- go test -race ./ledger/... ./internal/test/conformance/... passes, including TestConsensusGenTxFixtures (7 eras) and TestUtxoValidatePoolCertificatesMetadataHashTooBig
- Fail-before: reverting the checkMetadataHash gate in ledger/shelley/rules.go fails TestUtxoValidatePoolCertificatesMetadataHashTooBig/oversized_hash_is_rejected_from_major_version_5, confirming the test is not vacuous
- Class audit: all seven eras (Shelley through Dijkstra, via Conway) route through shelley.UtxoValidatePoolCertificates
- Error ordering in validatePoolRegistration matches the PR description
- dingo#4058 is open and correctly names both breaking call sites (sqlstore/pool.go:1566, ledger/view.go:262)

All 15 checks green at this head, no conflicts. No further blockers.

PR #2238, opened 2026-09-05
fix(ledger): reject null block headers

### LINKED ISSUE #2083: ledger: reject null block headers during decoding
Shelley and Byron main blocks can decode a CBOR null header successfully on
paths that defer body-hash validation. Later header accessors then dereference a
nil value instead of returning a decode error.

Acceptance criteria:

- Reject null headers during block decoding for every affected era.
- Keep valid block and Byron EBB decoding unchanged.
- Cover null, truncated, and valid headers through the public decode path.

wolf31o2:
Fixes #2083

- Reject null block headers in the public block decoder, including when body-hash validation is skipped.
- Preserve Dijkstra header null handling and blockfetch fallback safety.
- Add public-path regression coverage across Byron through Dijkstra.

### DIFF
--- a/ledger/block.go
+++ b/ledger/block.go
@@ -66,9 +66,11 @@ func NewBlockFromCbor(
 	if err != nil {
 		return nil, err
 	}
-	header := block.Header()
-	if header == nil || reflect.ValueOf(header).IsNil() {
-		return nil, errors.New("block header is nil")
+	if !cfg.SkipHeaderValidation {
+		header := block.Header()
+		if header == nil || reflect.ValueOf(header).IsNil() {
+			return nil, errors.New("block header is nil")
+		}
 	}
 	return block, nil
 }
--- a/ledger/block.go
+++ b/ledger/block.go
@@ -15,7 +15,9 @@
 package ledger
 
 import (
+	"errors"
 	"fmt"
+	"reflect"
 
 	"github.com/blinklabs-io/gouroboros/ledger/common"
 )
@@ -37,27 +39,38 @@ func NewBlockFromCbor(
 	}
 	// Default: validation enabled (SkipBodyHashValidation = false)
 
+	var block Block
+	var err error
 	switch blockType {
 	case BlockTypeByronEbb:
-		return NewByronEpochBoundaryBlockFromCbor(data, cfg)
+		block, err = NewByronEpochBoundaryBlockFromCbor(data, cfg)
 	case BlockTypeByronMain:
-		return NewByronMainBlockFromCbor(data, cfg)
+		block, err = NewByronMainBlockFromCbor(data, cfg)
 	case BlockTypeShelley:
-		return NewShelleyBlockFromCbor(data, cfg)
+		block, err = NewShelleyBlockFromCbor(data, cfg)
 	case BlockTypeAllegra:
-		return NewAllegraBlockFromCbor(data, cfg)
+		block, err = NewAllegraBlockFromCbor(data, cfg)
 	case BlockTypeMary:
-		return NewMaryBlockFromCbor(data, cfg)
+		block, err = NewMaryBlockFromCbor(data, cfg)
 	case BlockTypeAlonzo:
-		return NewAlonzoBlockFromCbor(data, cfg)
+		block, err = NewAlonzoBlockFromCbor(data, cfg)
 	case BlockTypeBabbage:
-		return NewBabbageBlockFromCbor(data, cfg)
+		block, err = NewBabbageBlockFromCbor(data, cfg)
 	case BlockTypeConway:
-		return NewConwayBlockFromCbor(data, cfg)
+		block, err = NewConwayBlockFromCbor(data, cfg)
 	case BlockTypeDijkstra:
-		return NewDijkstraBlockFromCbor(data, cfg)
+		block, err = NewDijkstraBlockFromCbor(data, cfg)
+	default:
+		return nil, fmt.Errorf("unknown node-to-client block type: %d", blockType)
 	}
-	return nil, fmt.Errorf("unknown node-to-client block type: %d", blockType)
+	if err != nil {
+		return nil, err
+	}
+	header := block.Header()
+	if header == nil || reflect.ValueOf(header).IsNil() {
+		return nil, errors.New("block header is nil")
+	}
+	return block, nil
 }
 
 func NewBlockHeaderFromCbor(blockType uint, data []byte) (BlockHeader, error) {
--- a/ledger/common/verify_config.go
+++ b/ledger/common/verify_config.go
@@ -101,6 +101,10 @@ func NewValidationError(
 // VerifyConfig holds runtime verification toggles.
 // Default values favor safety; tests or specific flows can opt out.
 type VerifyConfig struct {
+	// SkipHeaderValidation disables the public decoder's nil-header check.
+	// This is for callers that need to distinguish a representable block from
+	// a validation failure before applying their own validation.
+	SkipHeaderValidation bool
 	// SkipBodyHashValidation disables body hash verification in VerifyBlock().
 	// When false (default), full block CBOR must be available for validation.
 	// Useful for scenarios where full block CBOR is unavailable.
--- a/ledger/verify_block_edge_test.go
+++ b/ledger/verify_block_edge_test.go
@@ -376,27 +376,29 @@ func TestNewBlockFromCbor_RejectsNullHeaders(t *testing.T) {
 		name      string
 		blockType uint
 		data      []byte
+		wantError string
 	}{
-		{"byron ebb", ledger.BlockTypeByronEbb, []byte{0x83, 0xf6, 0x80, 0x80}},
+		{"byron ebb", ledger.BlockTypeByronEbb, []byte{0x83, 0xf6, 0x80, 0x80}, "decode Byron EBB block error: byron EBB block missing header"},
 		{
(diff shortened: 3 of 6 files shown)


protocol/blockfetch/client_raw_test.go
```diff
@@ -517,15 +494,8 @@ func TestGetBlockRangeRawFuncRejectsNilHeaderBlock(t *testing.T) {
...
-	// The fallback must have been declined outright, not entered and then
-	// tripped over the null header while extracting the point. Both end in a
-	// failed request, so without this the test would pass either way.
-	require.NotContains(
-		t,
-		connErr.Error(),
-		"raw block header",
-		"the fallback was entered for a block rejected on validation",
-	)
+	// Header extraction may still report the malformed header while correlating
```
chrisguiney, line 497: Non-blocking: the negative control dropped here is valid again on this head and worth restoring.

`ea28a2b` made `blockLayoutRepresentable` skip the nil-header check, so a nil-header block is representable again and the fallback is declined outright rather than entered and then tripping over correlation. I re-added `require.NotContains(t, connErr.Error(), "raw block header")` on this head and it passes; with only `640625b`'s `rawblock.go` it fails with

```
"protocol error: conway block header is nil\nraw block header has no body" should not contain "raw block header"
```

so it is precisely the assertion that catches the routing regression Cubic flagged. Without it, dropping `SkipHeaderValidation` from `blockLayoutRepresentable` again would re-open the hole with this test still green, because `rawCalled` stays empty on both sides.

ledger/block.go
```diff
@@ -37,27 +39,40 @@ func NewBlockFromCbor(
...
+		block, err = NewDijkstraBlockFromCbor(data, cfg)
+	default:
+		return nil, fmt.Errorf("unknown node-to-client block type: %d", blockType)
+	}
+	if err != nil {
+		return nil, err
+	}
+	if !cfg.SkipHeaderValidation {
+		header := block.Header()
+		if header == nil || reflect.ValueOf(header).IsNil() {
```
chrisguiney, line 71: Non-blocking: `reflect.Value.IsNil` panics on a non-nilable kind (`reflect: call of reflect.Value.IsNil on struct Value`), so this line would itself panic on a future `Block` whose `Header()` returns a value type rather than a pointer.

The class is complete today — all nine `Header()` implementations (byron main, byron EBB, shelley, allegra, mary, alonzo, babbage, conway, dijkstra) return a `*T` field, and `NewBlockFromCbor` only ever constructs those nine. A `Kind()` guard would keep the nil-safety check from becoming its own panic source as eras are added.

ledger/common/verify_config.go
```diff
@@ -101,6 +101,10 @@ func NewValidationError(
 // VerifyConfig holds runtime verification toggles.
 // Default values favor safety; tests or specific flows can opt out.
 type VerifyConfig struct {
+	// SkipHeaderValidation disables the public decoder's nil-header check.
+	// This is for callers that need to distinguish a representable block from
+	// a validation failure before applying their own validation.
+	SkipHeaderValidation bool
```
chrisguiney, line 107: Minor: inserting the field at the top of the struct changes the meaning of any positional `VerifyConfig{...}` literal in downstream code. Nothing here or in dingo uses one — all call sites are keyed — and `go vet`'s composites check flags unkeyed literals of imported structs, so this is a note rather than a defect. Appending would avoid it entirely.

chrisguiney (approved):
Verified against `ea28a2b` in an isolated worktree.

**Is a null header actually invalid?** Yes, in every era. cardano-ledger CDDL has `block = [header, ...]` and `header = [header_body, body_signature]` for Shelley, Allegra, Mary, Alonzo, Babbage, Conway and Dijkstra, and Byron's `mainblock`/`ebblock` require a `blockhead`/`ebbhead` array. `nil` is admitted only for `prev_hash` and `auxiliary_data`, both of which this change leaves alone. The tightening cannot reject a header a real node accepts.

**The crash is real and this fixes it.** Reverting the `ledger/block.go` check in place and replaying a null-header Conway block over a mock BlockFetch conversation with `SkipBlockValidation: true` panics in the peer-driven goroutine:

```
conway.(*ConwayBlock).SlotNumber(...)   ledger/conway/conway.go:204
blockfetch.(*Client).handleBlock(...)   protocol/blockfetch/client.go:1160
protocol.(*Protocol).recvLoop(...)      protocol/protocol.go:955
```

Unrecovered, so it takes the process down. With the fix the peer gets `protocol error: block header is nil` and the connection terminates, not the node. `SkipBlockValidation` is off by default and dingo never sets it, so the crash was reachable only for callers that opt in.

**Entry points covered.** `NewBlockFromCbor` is the only in-repo constructor of a `ledger.Block`, and every caller reaches it without `SkipHeaderValidation`: `blockfetch.handleBlock` (network), `chainsync/client.go:1110` (NtC), `pipeline.DecodeStage`, `NewBlockFromCborWithOffsets`. `blockLayoutRepresentable` is the sole opt-out and discards the block. Database reads through this path (dingo's `mithril/sync_gap.go`, `internal/node/backfill.go`, both `SkipBodyHashValidation: true`) newly get an error instead of a nil-header block; no stored valid block can be affected, since the CDDL forbids the shape.

**Era class is complete.** All nine `Header()` implementations return a `*T` field, and I confirmed both `f6` and `f7` in header position are rejected for all nine block types. `NewBlockHeaderFromCbor` (NtN chainsync headers) still accepts a bare `f6` and yields a zero header for every era, but that is pre-existing, causes no nil dereference, and is a separate change.

**Fail-before, by reverting each hunk in place:** byron main, shelley, allegra, mary, alonzo, babbage and conway fail without the `ledger/block.go` check. dijkstra fails without the `dijkstra.go` change and only that one — the old `var header DijkstraBlockHeader` left `BlockHeader` non-nil, so the block.go check could not see it. byron ebb passes either way; it is a guard on the pre-existing `byron EBB block missing header`, not a fail-before case.

**Bot standing.** CodeRabbit did not review this PR — its only comment is "Review limit reached". Cubic's review is anchored to `640625b`; its head run was incremental over recent commits and did not re-evaluate that review. Both Cubic findings are addressed by `ea28a2b`; replied on both threads.

**Checks on `ea28a2b`:** `go build ./...` 0 · `go test ./...` 0 (39 pkgs) · `go test -race ./ledger/... ./protocol/blockfetch/... ./protocol/chainsync/... ./pipeline/...` 0 · `go vet ./...` 0 · `golangci-lint 2.13.2 run ./...` 0 · build + lint of all 8 example modules 0 · merge with `origin/main` then build + test 0 (branch is 28 commits behind, no conflict). nilaway is not installed locally; the CI nilaway check is green on this head.

Two non-blocking notes inline plus one minor. Approving.

ledger/block.go
```diff
@@ -37,27 +39,38 @@ func NewBlockFromCbor(
...
+		block, err = NewDijkstraBlockFromCbor(data, cfg)
+	default:
+		return nil, fmt.Errorf("unknown node-to-client block type: %d", blockType)
+	}
+	if err != nil {
+		return nil, err
+	}
+	header := block.Header()
+	if header == nil || reflect.ValueOf(header).IsNil() {
+		return nil, errors.New("block header is nil")
```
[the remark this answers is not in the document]
chrisguiney: Valid at `640625b`, and addressed at the current head `ea28a2b`.

`blockLayoutRepresentable` now passes `SkipHeaderValidation: true`, so a nil-header block is representable again, `rawFallback` is false, and the request fails outright before any raw routing. Verified on this head: `connErr` is `protocol error: conway block header is nil` with no `raw block header` correlation error, meaning the fallback was declined rather than entered. With `640625b`'s `rawblock.go` the same probe yields `protocol error: conway block header is nil\nraw block header has no body`, which is the state you described.

One part still stands: the `require.NotContains(connErr, "raw block header")` assertion the first commit removed was not restored, so the guarantee is again enforced only by behavior and not by a test. Raised as a non-blocking comment on `protocol/blockfetch/client_raw_test.go`.

ledger/verify_block_edge_test.go
```diff
@@ -366,3 +366,37 @@ func TestVerifyBlock_BlockTypeEdgeCases(t *testing.T) {
...
+		{"babbage", ledger.BlockTypeBabbage, []byte{0x85, 0xf6, 0x80, 0x80, 0xa0, 0x80}},
+		{"conway", ledger.BlockTypeConway, []byte{0x85, 0xf6, 0x80, 0x80, 0xa0, 0x80}},
+		{"dijkstra", ledger.BlockTypeDijkstra, []byte{0x82, 0xf6, 0x84, 0xf6, 0x80, 0xf6, 0xf6}},
+	}
+	config := skipAllValidationConfig()
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := ledger.NewBlockFromCbor(tt.blockType, tt.data, config)
+			require.Error(t, err)
+			assert.Contains(t, strings.ToLower(err.Error()), "header")
```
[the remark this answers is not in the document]
chrisguiney: Addressed at `ea28a2b`. The table now carries a `wantError` per case and asserts `require.EqualError`, pinning `block header is nil` for byron main through conway, `decode Dijkstra block error: dijkstra block header is nil`, and `decode Byron EBB block error: byron EBB block missing header`. Confirmed the byron ebb case is a guard on pre-existing behavior rather than a fail-before case: it still passes with the `ledger/block.go` check reverted.

PR #2206, opened 2026-09-04
fix(kes): enforce strict signature verification

wolf31o2:
Closes #2204.

- require canonical, non-small-order Ed25519 points at SumKES leaves
- reject non-canonical scalars and malformed leaf inputs
- cover accepted standard-library edge cases through the 448-byte consensus path
- preserve reference vectors and add wrong-message, wrong-root, and wrong-period controls

---
## Summary by cubic
Enforces strict Ed25519 signature verification at SumKES leaves so malformed signatures no longer verify, matching Cardano's KES implementation.

**Bug Fixes**
- Replaces `ed25519.Verify` with `verifyEd25519Strict`, which validates canonical encodings and rejects small-order points before the signature equation check.
- Adds regression tests for malformed public keys, R points, and scalars in `kes`, plus a consensus-boundary test in `ledger`.

<sup>Written for commit df7c16b18fbbc7ce659ad0ea5447764b8c300227. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2206?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #2239, opened 2026-09-05
fix(crypto): apply strict Ed25519 rules outside Byron

### LINKED ISSUE #2234: Strict Ed25519 verification is applied only in kes; nine other verifiers accept small-order keys
#2206 added `kes/kes.go:169` `verifyEd25519Strict`, which rejects non-canonical and small-order public and R points and non-canonical S before calling `ed25519.Verify`. It is unexported and confined to `kes`. Every other Ed25519 verifier in this repository calls `crypto/ed25519.Verify` directly, which applies neither check.

`crypto/ed25519.Verify` accepts the edwards25519 identity public key `0100…00` with an all-zero signature for an arbitrary message, because `[S]B == R + [h]A` reduces to `identity == identity` for `A = R = identity, S = 0`:

```
crypto/ed25519.Verify(identity, "")                                = true
crypto/ed25519.Verify(identity, "a")                               = true
crypto/ed25519.Verify(identity, "arbitrary transaction body hash")  = true
pub = 0100000000000000000000000000000000000000000000000000000000000000
sig = 0100000000000000000000000000000000000000000000000000000000000000
      0000000000000000000000000000000000000000000000000000000000000000
```

**The call sites do not all share one reference primitive, so they must not all get the same treatment.**

### Verified against libsodium `crypto_sign_ed25519_verify_detached` — these need the strict rules

| call site | reference path |
|---|---|
| `ledger/common/verify.go:32` (`VerifyVKeySignature`, via `ValidateVKeyWitnesses` / `UtxoValidateSignatures`) | `Cardano.Ledger.Keys.Internal.verifySignedDSIGN` -> `DSIGN.Ed25519DSIGN` -> `cardano-crypto-class` -> libsodium |
| `ledger/common/verify.go:220` (`ValidateBootstrapWitnesses`) | `Cardano.Ledger.Keys.Bootstrap.verifyBootstrapWit` calls the same `verifySignedDSIGN`, **not** the Byron path, despite the witness being Byron-shaped |
| `ledger/verify_opcert.go:96` (`VerifyOpCertSignature`) | opcert cold signature, same `Ed25519DSIGN` |
| `consensus/validate.go:577` | the same opcert cold signature over `OpCertSignableBytes`, on the header path |
| `protocol/common/authentication.go:275` | the same opcert cold signature, on the handshake path |

libsodium rejects the identity proof for all three messages tested (`-1`), with honest and altered-message controls returning `0` / `-1`. `verifyEd25519Strict` rejects it too: the identity point is small-order, so `isSmallOrder` (`kes/kes.go:196`) is true. These five boundaries currently diverge from the node.

### Byron is a different primitive and must not be tightened

`consensus/byron/validate.go:340,364,641` and `ledger/byron/sign.go:98` verify Byron-era signatures. The reference is `Cardano.Crypto.Signing.Signature.verifySignatureRaw`, which calls `CC.verify` from `cardano-crypto-wallet`, whose bundled ed25519-donna `ed25519_sign_open` is:

```c
if ((RS[63] & 224) || !ge25519_unpack_negative_vartime(&A, pk))
        return -1;
...
return ed25519_verify(RS, checkR, 32) ? 0 : -1;
```

It checks only the high bits of S, does no small-order test on `A` or `R`, and compares packed bytes. It accepts proofs libsodium rejects, including the identity pair above. Byron blocks are immutable history, so applying the strict rules there would reject blocks the node accepts and break sync from genesis.

### Acceptance criteria

- A single strict verifier, exported from a location `ledger`, `consensus` and `protocol` can all reach, replaces `crypto/ed25519.Verify` at the five boundaries in the first table. `kes` uses the same one rather than keeping a private copy.
- The Byron call sites keep permissive verification, with a comment naming ed25519-donna as the reason so a later sweep does not "fix" them.
- Tests assert the identity key/signature pair is rejected at each strict boundary and **accepted** at the Byron ones, alongside a seeded-key control that verifies its own message and rejects an altered one.

wolf31o2:
Fixes #2234

## Defect

#2206 added `verifyEd25519Strict` (`kes/kes.go`), which rejects non-canonical and small-order public and R points and non-canonical S. It was unexported and confined to `kes`, so every other Ed25519 verifier still called `crypto/ed25519.Verify`, which applies neither check.

`crypto/ed25519.Verify` accepts the edwards25519 identity public key `0100…00` with an all-zero signature for an arbitrary message: `[S]B == R + [h]A` reduces to `identity == identity` when `A = R = identity` and `S = 0`. libsodium's `crypto_sign_ed25519_verify_detached` rejects the same pair.

## The boundaries are not uniform

Five verifiers reach libsodium in the reference and are changed here:

| call site | reference |
|---|---|
| `ledger/common/verify.go` `VerifyVKeySignature` | `Cardano.Ledger.Keys.Internal.verifySignedDSIGN` -> `Ed25519DSIGN` -> `cardano-crypto-class` |
| `ledger/common/verify.go` `ValidateBootstrapWitnesses` | `Cardano.Ledger.Keys.Bootstrap.verifyBootstrapWit` calls that same `verifySignedDSIGN`, not the Byron primitive, despite the witness being Byron-shaped |
| `ledger/verify_opcert.go` `VerifyOpCertSignature` | opcert cold signature, same DSIGN |
| `consensus/validate.go` | the same cold signature over `OpCertSignableBytes`, on the header path |
| `protocol/common/authentication.go` | the same cold signature, on the handshake path |

Four Byron verifiers are deliberately left permissive: `consensus/byron/validate.go` (three) and `ledger/byron/sign.go`. Their reference is `Cardano.Crypto.Signing.Signature.verifySignatureRaw` -> `CC.verify` in `cardano-crypto-wallet`, whose bundled ed25519-donna `ed25519_sign_open` is

```c
if ((RS[63] & 224) || !ge25519_unpack_negative_vartime(&A, pk))
        return -1;
```

— only the high bits of S, no small-order test on `A` or `R`. It accepts the identity pair. Byron blocks are immutable, so tightening these would reject chain the node accepts. Both files carry a comment saying so.

## Change

The criteria move to `internal/ed25519strict`, with the reasoning and the Byron exclusion in the package comment. `kes.verifyEd25519Strict` now delegates there rather than keeping a private copy.

## Tests

- `internal/ed25519strict` — the identity pair is rejected for three messages while `crypto/ed25519.Verify` accepts it, an honest signature verifies and fails against an altered message, and malformed sizes are rejected.
- `ledger/common` — `VerifyVKeySignature` rejects the identity pair and accepts an honest witness.
- `ledger` — `VerifyOpCertSignature` rejects an identity cold key and accepts an honest certificate, failing once the signed body changes.
- `ledger/byron` — `verifyEd25519` **accepts** the identity pair, guarding the Byron boundary against a later sweep, and still rejects a signature over a different payload.

The three rejection tests fail without the criteria. `kes`'s own `TestVerifySignedKESRejectsNonStrictEd25519Leaves` also fails, which is what pins the delegation.

---
## Summary by cubic
Fixes #2234 by applying strict Ed25519 verification to every signature boundary outside Byron; Byron keeps its permissive checks deliberately. Previously only the KES verifier was strict, while the others used `crypto/ed25519.Verify`, which accepts an identity public key with an all-zero signature for any message.

**Changes**
- Moves the criteria into `internal/ed25519strict`; `kes` now delegates to it.
- Routes vkey and bootstrap witness checks plus the three opcert checks (ledger, consensus header, handshake) through the strict verifier.

**Byron boundary**
- Byron's reference, `cardano-crypto-wallet`'s ed25519-donna, accepts the identity pair, and Byron blocks are immutable history, so tightening would reject accepted chain.
- Every permissive call site carries a warning comment and a pinning test against a future sweep.

<sup>Written for commit fd3d3b965274f514916a21a4120eec782b91d5ef. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2239?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Security Improvements**
  * Strengthened Ed25519 signature validation for operational certificates, transaction witnesses, bootstrap witnesses, and KES verification.
  * Strict verification now rejects non-canonical keys, signatures, and small-order points while continuing to accept valid signatures.
  * Byron signature verification remains permissive for compatibility with the Cardano reference implementation.

* **Tests**
  * Added coverage for valid signatures, altered messages, malformed inputs, and identity-key edge cases.

### DIFF
--- a/consensus/byron/validate.go
+++ b/consensus/byron/validate.go
@@ -337,6 +337,13 @@ func (v *HeaderValidator) validateSimpleSignature(
 		}
 		// Permissive fallback (opt-in only): verify directly against HeaderCbor
 		// WARNING: This is less secure and should only be used for testing
+		// Byron signatures are verified permissively on purpose. The reference is
+		// Cardano.Crypto.Signing.Signature.verifySignatureRaw, which calls CC.verify in
+		// cardano-crypto-wallet, whose bundled ed25519-donna ed25519_sign_open checks
+		// only the high bits of S and performs no small-order test on A or R. It
+		// accepts proofs libsodium rejects, and Byron blocks are immutable history, so
+		// routing these through internal/ed25519strict would reject chain the node
+		// accepts. Do not "fix" these to match the non-Byron boundaries.
 		valid := ed25519.Verify(
 			input.IssuerPubKey,
 			input.HeaderCbor,
--- a/consensus/byron/validate.go
+++ b/consensus/byron/validate.go
@@ -367,7 +367,10 @@ func (v *HeaderValidator) validateSimpleSignature(
 		return fmt.Errorf("failed to domain-separate ToSign: %w", err)
 	}
 
-	// Verify Ed25519 signature on the domain-separated ToSign.
+	// Byron verification stays permissive at this primary path. Its
+	// ed25519-donna reference accepts small-order public and R points, and
+	// tightening immutable Byron history to the non-Byron criteria would break
+	// sync from genesis. Do not route this through internal/ed25519strict.
 	valid := ed25519.Verify(
 		input.IssuerPubKey,
 		signed,
@@ -645,6 +648,10 @@ func (v *HeaderValidator) validateProxySignature(
 	// Use the Ed25519 portion of the delegate's extended key (first 32 bytes)
 	delegatePubKey := delegateVK[:32]
 
+	// Byron verification stays permissive at this delegated block-signature
+	// path. Its ed25519-donna reference accepts small-order public and R points,
+	// and tightening immutable Byron history to the non-Byron criteria would
+	// break sync from genesis. Do not route this through internal/ed25519strict.
 	valid := ed25519.Verify(delegatePubKey, signedBuf, blockSig)
 	if !valid {
 		return fmt.Errorf(
--- a/consensus/byron/validate_test.go
+++ b/consensus/byron/validate_test.go
@@ -53,6 +53,60 @@ func testByronConfig() ByronConfig {
 	}
 }
 
+func testByronIdentityPair() (ed25519.PublicKey, []byte) {
+	publicKey := make(ed25519.PublicKey, ed25519.PublicKeySize)
+	publicKey[0] = 0x01
+	signature := make([]byte, ed25519.SignatureSize)
+	signature[0] = 0x01
+	return publicKey, signature
+}
+
+func testByronHeaderCbor(t *testing.T) []byte {
+	t.Helper()
+	header := &byron.ByronMainBlockHeader{
+		ProtocolMagic: testByronProtocolMagicMainnet,
+	}
+	headerCbor, err := cbor.Encode(header)
+	require.NoError(t, err)
+	return headerCbor
+}
+
+func testByronProxyInput(
+	t *testing.T,
+	delegatePublicKey, blockSignature []byte,
+) *ValidateHeaderInput {
+	t.Helper()
+	issuerVK := make([]byte, byron.VerificationKeySize)
+	issuerVK[0] = 0x02
+	delegateVK := make([]byte, byron.VerificationKeySize)
+	copy(delegateVK, delegatePublicKey)
+	certSignature := make([]byte, ed25519.SignatureSize)
+
+	header := &byron.ByronMainBlockHeader{
+		ProtocolMagic: testByronProtocolMagicMainnet,
+	}
+	header.ConsensusData.BlockSig = []any{
+		uint64(byronSigTypeHeavy),
+		[]any{
+			[]any{uint64(7), issuerVK, delegateVK, certSignature},
+			blockSignature,
+		},
+	}
+	headerCbor, err := cbor.Encode(header)
+	require.NoError(t, err)
+
+	// Real callers populate BlockSig and HeaderCbor from the same decoded
+	// header. Round-trip here because proxyCertEpochCbor walks the raw bytes.
+	var decoded byron.ByronMainBlockHeader
+	_, err = cbor.Decode(headerCbor, &decoded)
+	require.NoError(t, err)
+
+	return &ValidateHeaderInput{
+		HeaderCbor: headerCbor,
+		BlockSig:   decoded.ConsensusData.BlockSig,
+	}
+}
+
 func TestNewHeaderValidator(t *testing.T) {
 	config := testByronConfig()
 	validator := NewHeaderValidator(config)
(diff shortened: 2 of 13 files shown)


consensus/byron/validate.go
```diff
@@ -337,6 +337,13 @@ func (v *HeaderValidator) validateSimpleSignature(
...
 		// Permissive fallback (opt-in only): verify directly against HeaderCbor
 		// WARNING: This is less secure and should only be used for testing
+		// Byron signatures are verified permissively on purpose. The reference is
+		// Cardano.Crypto.Signing.Signature.verifySignatureRaw, which calls CC.verify in
+		// cardano-crypto-wallet, whose bundled ed25519-donna ed25519_sign_open checks
+		// only the high bits of S and performs no small-order test on A or R. It
+		// accepts proofs libsodium rejects, and Byron blocks are immutable history, so
+		// routing these through internal/ed25519strict would reject chain the node
+		// accepts. Do not "fix" these to match the non-Byron boundaries.
 		valid := ed25519.Verify(
```
arepala-uml, line 347: GitHub only allows anchoring inline comments to lines this diff actually touches, so both findings below are attached here instead of at their real locations (`validateSimpleSignature`'s primary path at line 371, and `validateProxySignature`'s block-signature check at line 648).

### 1. Line 371 — `validateSimpleSignature`, primary (non-fallback) path

This is the primary, non-fallback path every ordinary (non-delegated) Byron block header verifies against. Only the `AllowSignatureFallback` branch above (line 329) carries the "Byron signatures are verified permissively on purpose ... Do not fix these" comment. This path — the one actually exercised on the hot path — and `validateProxySignature`'s block-signature check (line 648) have neither that comment nor a test pinning the permissive behavior, unlike `ledger/byron/sign.go`'s `verifyEd25519` (comment + `TestVerifyEd25519StaysPermissive`).

Ran this against the current head (`117a49a`) — it passes today, confirming the gap:

```go
// TestValidateSimpleSignatureStaysPermissive pins validateSimpleSignature's
// primary, non-fallback path: the one every ordinary (non-delegated) Byron
// block header is checked against. It currently accepts the identity
// public key with an all-zero signature for any header -- the same proof
// libsodium rejects and the same proof internal/ed25519strict was built to
// reject at the non-Byron boundaries.
func TestValidateSimpleSignatureStaysPermissive(t *testing.T) {
	header := &ledgerbyron.ByronMainBlockHeader{ProtocolMagic: 764824073}
	headerCbor, err := cbor.Encode(header)
	if err != nil {
		t.Fatalf("encode header: %v", err)
	}

	validator := NewHeaderValidator(testByronConfig())
	// AllowSignatureFallback stays false (the default): this must exercise
	// the primary domain-separated path, not the opt-in fallback branch that
	// already carries this comment.
	input := &ValidateHeaderInput{
		IssuerPubKey:   identityPubKey,
		BlockSignature: identitySig,
		HeaderCbor:     headerCbor,
	}

	if err := validator.validateSimpleSignature(input); err != nil {
		t.Errorf(
			"validateSimpleSignature rejected the identity pair; ed25519-donna "+
				"accepts it, so this boundary must stay permissive: %v",
			err,
		)
	}
}
```

Paired with a `TestValidateSimpleSignatureRejectsGarbage` control (random key/signature, expect rejection) to prove this isn't a vacuous pass. Suggest adding the same guard comment as the fallback branch here, plus these two tests.


---

### 2. Line 648 — `validateProxySignature`, delegate block-signature check

Same gap as `validateSimpleSignature`'s primary path (line 371): this is the delegate block-signature check for every PBFT heavy/light-delegation Byron header, and it's uncommented and untested for permissiveness, unlike `ledger/byron/sign.go`'s `verifyEd25519`.

Ran this against the current head (`117a49a`) — passes today. `SkipDelegationCertVerification` isolates this check from the separate, already-covered delegation-certificate check so the identity pair reaches this line:

```go
// TestValidateProxySignatureStaysPermissive pins validateProxySignature's
// block-signature check: the delegate-signed half of every PBFT
// heavy/light-delegation Byron header. It currently accepts an identity
// delegate key with an all-zero block signature, for the same
// ed25519-donna reason as the simple-signature path.
func TestValidateProxySignatureStaysPermissive(t *testing.T) {
	issuerVK := make([]byte, 64)
	issuerVK[0] = 0x02 // arbitrary; only used for the (skipped) cert check
	delegateVK := make([]byte, 64)
	copy(delegateVK, identityPubKey) // block-signature check reads delegateVK[:32]
	certSig := make([]byte, 64)      // size-checked only; cert check is skipped

	header := &ledgerbyron.ByronMainBlockHeader{ProtocolMagic: 764824073}
	header.ConsensusData.BlockSig = []any{
		uint64(byronSigTypeHeavy),
		[]any{
			[]any{uint64(7), issuerVK, delegateVK, certSig},
			identitySig,
		},
	}
	headerCbor, err := cbor.Encode(header)
	if err != nil {
		t.Fatalf("encode header: %v", err)
	}
	// Real callers populate BlockSig from the same decoded header as
	// HeaderCbor (see consensus/byron/pbft.go); round-trip here so the two
	// stay consistent, since proxyCertEpochCbor re-walks the raw HeaderCbor.
	var decoded ledgerbyron.ByronMainBlockHeader
	if _, err := cbor.Decode(headerCbor, &decoded); err != nil {
		t.Fatalf("decode header: %v", err)
	}

	validator := NewHeaderValidator(testByronConfig())
	validator.SkipDelegationCertVerification = true
	input := &ValidateHeaderInput{
		HeaderCbor: headerCbor,
		BlockSig:   decoded.ConsensusData.BlockSig,
	}

	if err := validator.validateProxySignature(input, byronSigTypeHeavy); err != nil {
		t.Errorf(
			"validateProxySignature rejected an identity delegate key and an "+
				"all-zero block signature; ed25519-donna accepts this pair: %v",
			err,
		)
	}
}
```

Paired with a `TestValidateProxySignatureRejectsGarbage` control (random delegate key/signature, expect rejection). Suggest adding the same guard comment used at the other Byron call sites here, plus these two tests.

wolf31o2: Addressed in fd3d3b9: both production verification paths now document the Byron compatibility boundary, and each path has identity-pair acceptance plus unrelated-signature rejection coverage. `go test -race ./consensus/... -count=1` and scoped lint pass.

arepala-uml (approved):
Verified the strict-verification logic in `internal/ed25519strict` matches the already-tested code it was extracted from in `kes/kes.go` (PR #2206) byte-for-byte. Grepped all remaining `ed25519.Verify(` call sites in the PR head: only the 4 Byron files (each with a comment explaining the deliberate exclusion) and the strict package's own delegation call remain — no boundary missed. No new dependency (`filippo.io/edwards25519` was already in use). Identity-key rejection tests assert `crypto/ed25519.Verify` still accepts the pair first, so they can't silently stop discriminating. CI is green and CodeRabbit/Cubic have no open findings.

Issue #2080, opened 2026-08-25
address: reject overflowing pointer varints


wolf31o2:
Pointer-address varint decoding does not report overflow. Values beyond the
supported range are silently narrowed and no longer round-trip to the original
address.

Acceptance criteria:

- Detect overflow while decoding each pointer component.
- Return a typed error instead of a narrowed value.
- Cover maximum valid, overflowing, truncated, and round-trip inputs.

PR #2253, opened 2026-09-06
fix(ledger): accept trailing address bytes before Babbage

wolf31o2:
`Address` decoding now mirrors Cardano's era-specific trailing-byte rules. Pre-Babbage transaction outputs accept and crop trailing Shelley address bytes, while Babbage-and-later outputs, reward accounts, proposal return addresses, general address constructors, and Byron addresses remain strict.

The change also resets reused Alonzo output shape state and adds regression coverage for strict public decoders, proposal addresses, and output reuse.

Validated with `go test ./...`.

### DIFF
--- a/ledger/address_trailing_bytes_test.go
+++ b/ledger/address_trailing_bytes_test.go
@@ -1,0 +1,254 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ledger_test
+
+import (
+	"bytes"
+	"encoding/hex"
+	"strings"
+	"testing"
+
+	"github.com/blinklabs-io/gouroboros/cbor"
+	"github.com/blinklabs-io/gouroboros/ledger/alonzo"
+	"github.com/blinklabs-io/gouroboros/ledger/babbage"
+	"github.com/blinklabs-io/gouroboros/ledger/common"
+	"github.com/blinklabs-io/gouroboros/ledger/dijkstra"
+	"github.com/blinklabs-io/gouroboros/ledger/mary"
+	"github.com/blinklabs-io/gouroboros/ledger/shelley"
+)
+
+// mainnetTrailerAddress is a type 6 (payment key, no staking) mainnet address
+// carrying one byte past its payload. It is one of the addresses that exist on
+// Cardano mainnet with unconsumed trailing bytes
+// (IntersectMBO/cardano-ledger#2729).
+const mainnetTrailerAddress = "61549b5a20e449a3e394b762705f64b9a26b99013003a2bfdba239967c00"
+
+// testnetTrailerAddress is the same shape on a testnet, which the previous
+// mainnet-only trailer whitelist never admitted. fromCborBackwardsBothAddr is
+// unconditional on network (cardano-ledger
+// libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs).
+var testnetTrailerAddress = hex.EncodeToString(
+	append(
+		append(
+			[]byte{byte(common.AddressTypeKeyNone << 4)},
+			make([]byte, common.AddressHashSize)...,
+		),
+		0xFF,
+	),
+)
+
+func trailerAddressCbor(t *testing.T, addrHex string) []byte {
+	t.Helper()
+	raw, err := hex.DecodeString(addrHex)
+	if err != nil {
+		t.Fatalf("bad test address hex: %v", err)
+	}
+	encoded, err := cbor.Encode(raw)
+	if err != nil {
+		t.Fatalf("encode address: %v", err)
+	}
+	return encoded
+}
+
+// arrayOutput builds [address, coin], the Shelley through Alonzo output shape.
+func arrayOutput(addrCbor []byte) []byte {
+	out := []byte{0x82}
+	out = append(out, addrCbor...)
+	// 1000000 lovelace
+	return append(out, 0x1a, 0x00, 0x0f, 0x42, 0x40)
+}
+
+// mapOutput builds {0: address, 1: coin}, the Babbage output shape.
+func mapOutput(addrCbor []byte) []byte {
+	out := []byte{0xa2, 0x00}
+	out = append(out, addrCbor...)
+	return append(out, 0x01, 0x1a, 0x00, 0x0f, 0x42, 0x40)
+}
+
+func decodedAddress(t *testing.T, out common.TransactionOutput) common.Address {
+	t.Helper()
+	return out.Address()
+}
+
+// TestPreBabbageOutputsAcceptTrailingAddressBytes covers decoder versions below
+// 7, where fromCborBothAddr selects fromCborBackwardsBothAddr and the trailing
+// bytes are cropped rather than rejected.
+func TestPreBabbageOutputsAcceptTrailingAddressBytes(t *testing.T) {
+	for network, addrHex := range map[string]string{
+		"mainnet": mainnetTrailerAddress,
+		"testnet": testnetTrailerAddress,
+	} {
+		addrCbor := trailerAddressCbor(t, addrHex)
+		raw, err := hex.DecodeString(addrHex)
+		if err != nil {
+			t.Fatalf("bad test address hex: %v", err)
+		}
+		consumed := raw[:len(raw)-1]
+		trailer := raw[len(raw)-1:]
+
+		outputBytes := arrayOutput(addrCbor)
+		for _, tc := range []struct {
+			name   string
+			output common.TransactionOutput
+		}{
+			{"shelley", &shelley.ShelleyTransactionOutput{}},
+			{"mary", &mary.MaryTransactionOutput{}},
+			{"alonzo", &alonzo.AlonzoTransactionOutput{}},
+		} {
+			t.Run(tc.name+"/"+network, func(t *testing.T) {
+				if _, err := cbor.Decode(outputBytes, tc.output); err != nil {
+					t.Fatalf("decode output: %v", err)
+				}
+				addr := decodedAddress(t, tc.output)
+				if !bytes.Equal(addr.TrailingBytes(), trailer) {
+					t.Fatalf(
+						"trailing bytes: got %x, want %x",
+						addr.TrailingBytes(),
+						trailer,
+					)
+				}
+				addrBytes, err := addr.Bytes()
+				if err != nil {
+					t.Fatalf("address bytes: %v", err)
+				}
+				if !bytes.Equal(addrBytes, consumed) {
+					t.Fatalf(
+						"address bytes: got %x, want the cropped %x",
+						addrBytes,
+						consumed,
+					)
+				}
+			})
+		}
+	}
+}
+
+// TestBabbageOnwardOutputsRejectTrailingAddressBytes covers decoder version 7
+// and later, where fromCborRigorousBothAddr reaches ensureBufIsConsumed.
+func TestBabbageOnwardOutputsRejectTrailingAddressBytes(t *testing.T) {
+	addrCbor := trailerAddressCbor(t, mainnetTrailerAddress)
+	for _, tc := range []struct {
+		name   string
+		bytes  []byte
+		output common.TransactionOutput
+	}{
+		{
+			"babbage map form",
+			mapOutput(addrCbor),
+			&babbage.BabbageTransactionOutput{},
+		},
+		{
+			"babbage legacy array form",
+			arrayOutput(addrCbor),
+			&babbage.BabbageTransactionOutput{},
+		},
+		{
+			"dijkstra map form",
+			mapOutput(addrCbor),
+			&dijkstra.DijkstraTransactionOutput{},
+		},
+		{
+			"dijkstra array form",
+			arrayOutput(addrCbor),
+			&dijkstra.DijkstraTransactionOutput{},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			_, err := cbor.Decode(tc.bytes, tc.output)
+			if err == nil {
+				t.Fatal("expected trailing address bytes to be rejected")
+			}
+			if !strings.Contains(err.Error(), "unexpected trailing byte") {
+				t.Fatalf("unexpected error: %v", err)
+			}
+		})
+	}
+}
+
+// TestBabbageOnwardOutputsAcceptWellFormedAddresses is the negative control
+// for the gate above: the same output shapes without a trailer still decode.
+func TestBabbageOnwardOutputsAcceptWellFormedAddresses(t *testing.T) {
+	addrHex := mainnetTrailerAddress[:len(mainnetTrailerAddress)-2]
+	addrCbor := trailerAddressCbor(t, addrHex)
+	for _, tc := range []struct {
+		name   string
+		bytes  []byte
+		output common.TransactionOutput
+	}{
+		{
+			"babbage map form",
+			mapOutput(addrCbor),
+			&babbage.BabbageTransactionOutput{},
+		},
+		{
+			"babbage legacy array form",
+			arrayOutput(addrCbor),
+			&babbage.BabbageTransactionOutput{},
+		},
+		{
+			"dijkstra map form",
+			mapOutput(addrCbor),
+			&dijkstra.DijkstraTransactionOutput{},
+		},
+		{
+			"dijkstra array form",
+			arrayOutput(addrCbor),
+			&dijkstra.DijkstraTransactionOutput{},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			if _, err := cbor.Decode(tc.bytes, tc.output); err != nil {
+				t.Fatalf("decode output: %v", err)
+			}
+			addr := decodedAddress(t, tc.output)
+			if len(addr.TrailingBytes()) != 0 {
+				t.Fatalf(
+					"unexpected trailing bytes: %x",
+					addr.TrailingBytes(),
+				)
+			}
+		})
+	}
+}
+
+// TestRewardAccountRejectsTrailingBytes covers decodeAccountAddressT, which
+// calls ensureBufIsConsumed with no version gate, so a reward account with
+// unconsumed bytes is invalid in every era.
+func TestRewardAccountRejectsTrailingBytes(t *testing.T) {
+	// Type 14 (no payment, staking key) on mainnet, with one byte past the
+	// 28 byte credential
+	raw := append(
+		append(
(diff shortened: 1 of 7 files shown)


ledger/common/address.go
```diff
@@ -419,78 +425,47 @@ func (a *Address) populateFromBytes(data []byte) error {
...
 	if len(payload) > 0 {
-		if a.networkId != AddressNetworkMainnet ||
-			!isKnownMalformedAddressTrailer(payload) {
-			return fmt.Errorf(
-				"invalid address data: %d unexpected trailing byte(s)",
-				len(payload),
-			)
-		}
-		a.extraData = payload[:]
+		a.trailingBytes = slices.Clone(payload)
```
chrisguiney, line 441: Blocker: this makes every `Address` decode lenient, but only outputs and withdrawals are re-tightened. The Conway and Dijkstra proposal procedure return address is not, and the reference keeps it strict in every era.

`ProposalProcedure.pProcReturnAddr :: !AccountAddress` (`eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs:537`) decodes through `instance DecCBOR AccountAddress where decCBOR = fromCborAccountAddress` (`Address.hs:387-389`), which reaches `decodeAccountAddressT`'s ungated `ensureBufIsConsumed` (`Address.hs:942-954`). gouroboros models it as `common.Address` (`ledger/conway/gov.go:32`, `ledger/dijkstra/gov.go:31`) and nothing calls `RewardAccountCredential` or `CheckAddressFullyConsumed` on it: `ledger/conway/rules.go:604` and `:1362` only read `NetworkId()` and `StakeCredential()`.

Reachable from `ConwayTransactionBody` key 20 (`ledger/conway/conway.go:647`), `DijkstraTransactionBody` (`ledger/dijkstra/dijkstra.go:757`) and `DijkstraSubTransactionBody` (`:1073`). Decoding this body, a return address of `e1ab..ab` plus one `ff` byte:

```
a40080018002001481841a000f4240581ee1abababababababababababababababababababababababababab
ababff8106826058200000000000000000000000000000000000000000000000000000000000000000
```

- on e6c5679 (base): `invalid address data: 1 unexpected trailing byte(s)`
- on a3bb5a6: decodes, `TrailingBytes()` is `ff`, return address `stake1ux46h2at4w46h2at4w46h2at4w46h2at4w46h2at4w46h2cfk870n`

so a node built on this accepts a transaction that cardano-node rejects at decode.

ledger/common/address.go
```diff
@@ -419,78 +425,47 @@ func (a *Address) populateFromBytes(data []byte) error {
...
-	},
-	{44},
+// CheckAddressFullyConsumed rejects an address whose encoding carried bytes
+// past its payload. From decoder version 7 (Babbage) fromCborBothAddr selects
+// fromCborRigorousBothAddr, whose decodeAddrStateLenientT reaches
+// "unless isLenient $ ensureBufIsConsumed" (cardano-ledger
+// libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs), so an output
+// address from Babbage onward must be rejected where Shelley through Alonzo
+// crop it.
+func CheckAddressFullyConsumed(a Address) error {
```
chrisguiney, line 453: The reference's general parser stays strict: `decodeAddr = decodeAddrStateT = decodeAddrStateLenientT False False` (`Address.hs:597-602`). Leniency exists only inside `fromCborBackwardsBothAddr`, i.e. the pre-Babbage CBOR output decode. Here the default is inverted: leniency lives in `populateFromBytes`, and strictness becomes opt-in per call site through this function, so every consumer of the general API is now lenient.

Two consequences outside the ledger decode path, both measured against the base commit:

- `NewAddress` on a bech32 string encoding a 29-byte type 6 payload plus two garbage bytes returned `invalid address data: 2 unexpected trailing byte(s)` on the base; on a3bb5a6 it returns an `Address` whose `String()` is a *different* address than the input (`addr1v...7745y4leut` in, `addr1v...c4u8cnp` out). A consumer that validates a user-supplied address string now gets "valid" plus a silently rewritten address.
- The `cbor.RegisterAddressFormatter` hook (`ledger/common/address.go:65`) now matches 32-byte blobs whose first byte is a plausible header: `61` followed by 31 bytes renders as `addr1vyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zygatvcjl`. Headers `0x60/0x61/0x70/0x71/0xe0/0xe1/0xf0/0xf1` and the pointer types all qualify, so a hash-sized byte string is annotated as an address in diagnostic output where it previously was not.

Confining the crop to the pre-Babbage output decode removes both, and removes the need for an allowlist of call sites to re-tighten.

ledger/common/address.go
```diff
@@ -341,6 +344,9 @@ func (a *Address) populateFromBytes(data []byte) error {
 		a.paymentPayload = AddressPayloadKeyHash{
 			Hash: AddrKeyHash(NewBlake2b224(byronAddr.Hash)),
 		}
+		if byronLen < len(data) {
```
chrisguiney, line 347: The description's rationale for this branch does not match the reference. The bootstrap branch is not lenient: `decodeBootstrapAddress` runs `decodeFull' byronProtVer` over the whole buffer (`Address.hs:694-702`), which fails on leftover bytes regardless of the `isLenient` flag, so the reference rejects a Byron address with trailing bytes in every era rather than accepting it below version 7.

The behaviour change itself is in the right direction and is untested. With `base58("Ae2tdPwUPEYwFx4dmJheyNPPYXtvHbJLeCaA96o6Y2iiUL18cAt7AizN2zG") + 0xFF` as the output address:

- Babbage map form: accepted on the base commit, rejected on a3bb5a6 — this now matches the reference.
- Shelley array form: accepted on both. That divergence from the reference is pre-existing and not introduced here, since the Byron branch returned before the old trailing-byte check.

ledger/address_trailing_bytes_test.go
```diff
@@ -0,0 +1,254 @@
...
+				)
+			}
+		})
+	}
+}
+
+// TestRewardAccountRejectsTrailingBytes covers decodeAccountAddressT, which
+// calls ensureBufIsConsumed with no version gate, so a reward account with
+// unconsumed bytes is invalid in every era.
+func TestRewardAccountRejectsTrailingBytes(t *testing.T) {
```
chrisguiney, line 262: `TestRewardAccountRejectsTrailingBytes` calls `RewardAccountCredential` directly. It does fail without the fix, but at its setup line — `address_trailing_bytes_test.go:241: decode reward account: invalid address data: 1 unexpected trailing byte(s)` — not at the assertion, so it does not pin the guard it describes.

No test decodes a transaction body carrying a trailing-byte withdrawal key, which is the path `ValidateWithdrawalAddresses` protects in each era, and none covers a proposal procedure return address.

chrisguiney (changes requested):
Reviewed a3bb5a6 in an isolated worktree against `origin/main`.

**Reference check** (cardano-ledger master, `libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs`):

- The era split is real. `fromCborBothAddr` (:459-476) selects `fromCborBackwardsBothAddr` below decoder version 7 and `fromCborRigorousBothAddr` from 7 on, and `type ProtVerLow BabbageEra = 7` (`libs/cardano-ledger-core/internal/Cardano/Ledger/Internal/Definition/Era.hs:124`). The backwards decoder crops to `bytesConsumed` (:495-503), so returning the consumed prefix from `Bytes()` matches what the reference keeps as `CompactAddr`.
- The Babbage wire decoder is `decCBOR = decodeBabbageTxOut fromCborBothAddr` (`eras/babbage/impl/src/Cardano/Ledger/Babbage/TxOut.hs:522`), which routes both the map form and the legacy array form through the version gate, so applying the check to the legacy branch is correct. The backwards decoder appears only in `DecShareCBOR` (:536-539), which is ledger-state deserialisation rather than the wire.
- `decodeAccountAddressT` calls `ensureBufIsConsumed "AccountAddress"` with no version gate (:938-955), as the description says.

**Disposition: request changes.** One blocker. The relaxation is placed in the shared parser rather than in the pre-Babbage output decode, and one decode site the reference keeps strict in every era is not re-tightened: the Conway and Dijkstra proposal procedure return address. Verified as a regression against the base commit at full `ConwayTransactionBody` level. Two non-blocking findings are inline.

The class is otherwise complete. Every path that decodes a `common.Address` from block or transaction CBOR was checked: outputs (Shelley/Allegra/Mary/Alonzo lenient, Babbage/Conway/Dijkstra checked, including the Dijkstra map form that reuses the Babbage output), withdrawals (`ValidateWithdrawalAddresses` at each era's body decode plus `TreasuryWithdrawalGovAction`), pool registration reward accounts (`rewardAccountCBOR`, untouched and still strict), and Byron-era outputs (exact field bytes, no trailer possible). Proposal return addresses are the only gap. The `common.Address` fields in `ledger/*/errors.go` are constructed in-process, not decoded from the wire.

**Validation on a3bb5a6**

| check | result |
| --- | --- |
| `go build ./...` | pass |
| `go test -race ./...` | pass, 39 packages |
| `golangci-lint run ./...` (isolated cache, v2.13.2) | 0 issues |
| 8 nested example modules (`replace ../..`), `go build ./...` | pass |

Fail-before, by reverting the accept branch in place and keeping the new API so the tests still compile: `TestPreBabbageOutputsAcceptTrailingAddressBytes` (6 subtests), `TestPopulateFromBytesCropsTrailingBytes` (8), `TestKnownMalformedMainnetAddressesCrop`, `TestPopulateFromBytesClearsAndCopiesTrailingBytes`, `TestAddressFromBytes` and `TestCIP0019_MaximumLengthAddresses` all fail. `TestBabbageOnwardOutputsRejectTrailingAddressBytes` passes under that revert, but fails on the true base commit for all four forms, because the base whitelist admitted that mainnet trailer — it is a real discriminator, not a guard. `TestRewardAccountRejectsTrailingBytes` fails on the base at its setup line rather than at its assertion; see the inline note.

Not run: conformance vectors, devnet or mainnet replay, and `go test -fuzz` on the address fuzz targets (seed corpus only under the plain run).

**Bot standing on this head.** CodeRabbit has not reviewed this PR: its only comment is a rate-limit notice ("Review limit reached"), last updated at the head push. Cubic's head run was incremental — its check output says "0 issues found across 2 files (changes from recent commits)" — so its base-commit review is the only full pass. Both of its findings are genuinely fixed at head: `a.trailingBytes = nil` at `ledger/common/address.go:309` clears reused state, and `slices.Clone` is applied at the two stores (:348, :441) and at the getter (:467), with `TestPopulateFromBytesClearsAndCopiesTrailingBytes` covering both.

ledger/common/address.go
```diff
@@ -419,78 +422,47 @@ func (a *Address) populateFromBytes(data []byte) error {
...
 	if len(payload) > 0 {
-		if a.networkId != AddressNetworkMainnet ||
-			!isKnownMalformedAddressTrailer(payload) {
-			return fmt.Errorf(
-				"invalid address data: %d unexpected trailing byte(s)",
-				len(payload),
-			)
-		}
-		a.extraData = payload[:]
+		a.trailingBytes = payload[:]
```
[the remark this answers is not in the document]
chrisguiney: Confirmed fixed at a3bb5a6: `populateFromBytes` sets `a.trailingBytes = nil` before decoding (ledger/common/address.go:308-309), so a reused `Address` no longer carries a previous trailer. `TestPopulateFromBytesClearsAndCopiesTrailingBytes` covers the re-decode case. One residual: the `len(data) == 0` early return happens before the clear, but that path also returns an error, so the address is not usable afterwards.

ledger/common/address.go
```diff
@@ -419,78 +422,47 @@ func (a *Address) populateFromBytes(data []byte) error {
...
-		if bytes.Equal(trailer, known) {
-			return true
-		}
-	}
-	return false
+// TrailingBytes returns the bytes that followed the address payload and were
+// not consumed by it. They are dropped from Bytes(), as the reference drops
+// them below decoder version 7, and are non-empty only for an address that a
+// stricter era has to reject.
+func (a Address) TrailingBytes() []byte {
```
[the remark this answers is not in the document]
chrisguiney: Confirmed fixed at a3bb5a6: `slices.Clone` is applied at both stores (ledger/common/address.go:348 for the Byron trailer, :441 for the Shelley trailer) and again in the getter (:467), so neither the input buffer nor the returned slice aliases the address state. `TestPopulateFromBytesClearsAndCopiesTrailingBytes` mutates both and asserts the address is unchanged.

ledger/common/address.go
```diff
@@ -425,78 +431,47 @@ func (a *Address) populateFromBytes(data []byte) error {
...
-		if a.networkId != AddressNetworkMainnet ||
-			!isKnownMalformedAddressTrailer(payload) {
-			return fmt.Errorf(
-				"invalid address data: %d unexpected trailing byte(s)",
-				len(payload),
-			)
-		}
-		a.extraData = payload[:]
+		a.trailingBytes = slices.Clone(payload)
 	}
```
arepala-uml, line 472: **Correctness (blocking).** `populateFromBytes` records trailing bytes here but never returns an error for them, and every general-purpose entry point (`NewAddress`, `NewAddressFromBytes`, `Address.UnmarshalCBOR`) routes through it with no strictness re-check. Only four call sites opt back into strictness via `CheckAddressFullyConsumed` (Babbage/Dijkstra outputs, Conway/Dijkstra proposal return addresses) — every other caller of the public decode API is now lenient regardless of era, not just at the pre-Babbage output path the comment above describes.

Reproduced against the merge-base (`944c6e47`) and this head:

```go
payload := make([]byte, 57)
payload[0] = 0x01 // KeyKey mainnet header
for i := 1; i < 57; i++ { payload[i] = byte(i) }
withTrailer := append(append([]byte{}, payload...), 0xde, 0xad)
addr, err := common.NewAddressFromBytes(withTrailer)
```

- At `944c6e47`: `err = "invalid address data: 2 unexpected trailing byte(s)"`.
- At this head (`159f0d18`): `err = nil`; `addr.Bytes()` returns the *cropped*, well-formed 57-byte address (the `dead` suffix silently dropped), and `addr.String()` renders a normal-looking bech32 address (`addr1qyq...`).

Any caller using `NewAddress`/`NewAddressFromBytes` to validate a user- or API-supplied address string (wallets, explorers, anything outside the ledger-rule decode paths this PR did tighten) now gets a false "valid" plus a silently different address than what was submitted, for arbitrary trailing garbage — not just for the specific pre-Babbage on-chain leniency this PR is about. Consider gating the general API the same way the four fixed call sites do (call `CheckAddressFullyConsumed` in `NewAddress`/`NewAddressFromBytes`/`UnmarshalCBOR` themselves, or expose a strict variant and make it the default), rather than leaving strictness as something each new call site has to remember to opt into.

arepala-uml (changes requested):
Requesting changes. Corroborates the still-open review on this PR: the general-purpose Address decode API (NewAddress, NewAddressFromBytes, Address.UnmarshalCBOR) is now unconditionally lenient about trailing bytes, not just at the pre-Babbage output path the reference implementation permits it for. Reproduced with a fail-before/fail-after test against the merge-base. Details in the inline comment.

ledger/alonzo/alonzo.go
```diff
@@ -497,12 +497,22 @@ type AlonzoTransactionOutput struct {
...
 			return err
 		}
-		*o = AlonzoTransactionOutput(tmp)
+		address, err := common.NewAddressFromBytesLenient(tmp.OutputAddress)
+		if err != nil {
+			return err
+		}
+		o.OutputAddress = address
+		o.OutputAmount = tmp.OutputAmount
+		o.OutputDatumHash = tmp.OutputDatumHash
```
chrisguiney, line 515: Blocker: this branch no longer resets legacyOutput, which the replaced *o = AlonzoTransactionOutput(tmp) did. Decoding [addr, coin] and then [addr, coin, datum_hash] into the same value leaves legacyOutput true, so MarshalCBOR emits the 2-element form and drops the datum hash: 83581d61…1a000f42405820abab…ab re-encodes as 82581d61…1a000f4240 at this head, and round-trips unchanged with alonzo.go from 944c6e4. Assigning a fresh value also clears OutputDatumHash in the legacy branch. A reuse round-trip test is needed.

ledger/common/address.go
```diff
@@ -425,91 +449,66 @@ func (a *Address) populateFromBytes(data []byte) error {
...
+func (a Address) TrailingBytes() []byte {
+	return slices.Clone(a.trailingBytes)
 }
 
 func (a *Address) UnmarshalCBOR(data []byte) error {
 	// Try to unwrap as bytestring (Shelley and forward)
 	tmpData := []byte{}
 	if _, err := cbor.Decode(data, &tmpData); err == nil {
-		err := a.populateFromBytes(tmpData)
+		err := a.populateFromBytes(tmpData, false)
```
chrisguiney, line 505: Required test: UnmarshalCBOR and NewAddress (:121) are strict again, but neither is pinned. Passing true at both populateFromBytes calls here, or at :121, leaves every test under ledger, protocol and cbor passing; only NewAddressFromBytes has a direct test. The checks added in conway/gov.go, dijkstra/gov.go and the Babbage map branch sit behind this decode and are unreachable, which masks it. Direct rejection tests are needed for cbor.Decode into common.Address and for NewAddress on a bech32 string with trailing bytes.

ledger/common/address.go
```diff
@@ -347,6 +362,15 @@ func (a *Address) populateFromBytes(data []byte) error {
 		a.paymentPayload = AddressPayloadKeyHash{
 			Hash: AddrKeyHash(NewBlake2b224(byronAddr.Hash)),
 		}
+		if byronLen < len(data) {
+			if !allowTrailing {
```
chrisguiney, line 366: Non-blocking, unchanged since the previous review: Shelley through Alonzo outputs still accept a Byron address with trailing bytes (a Shelley [byron_addr+ff, coin] output decodes with trailer ff). The reference rejects it in every era, since decodeBootstrapAddress runs decodeFull' over the whole buffer regardless of isLenient (Address.hs:695-704), so this should reject unconditionally. No test covers the branch: forcing the condition false passes every ledger, protocol and cbor test.

ledger/common/address.go
```diff
@@ -208,8 +220,8 @@ func NewAddressFromParts(
 			)
 		}
 	case AddressTypeKeyPointer, AddressTypeScriptPointer:
-		// Preserve pointer-address behavior via the existing byte path so
-		// extra trailing data continues to round-trip unchanged.
+		// Preserve pointer-address behavior via the existing byte path, which
+		// normalizes the pointer encoding the way the reference does.
```
chrisguiney, line 224: This comment overstates parity. Re-encoding yields minimal varints, but below Conway the reference maps a pointer whose slot exceeds Word32 or whose index exceeds Word16 to Ptr 0 0 0 (decodePtrLenient, mkPtrNormalized), and from Conway decodePtr rejects it. Here a slot of 2^32 decodes in every era, Babbage map form included, and re-encodes as 90808080000000. The pointer decode is not changed by this PR and overflow is tracked in #2080, so the comment should not claim reference normalization.

chrisguiney (dismissed):
Re-review at e96cea5. Cleared: the proposal return address blocker (its test fails once both guards are removed), the reward-account test now fails at its assertion, and general decoding is strict again, resolving the general-parser finding and the matching blocker in the other open review; its tests are required inline. New blocker: Alonzo output reuse. The description is stale: lenient shared parser, no mention of NewAddressFromBytesLenient, the incorrect Byron rationale, and no note that the public constructors now reject the eight formerly whitelisted mainnet trailers.

arepala-uml (approved):

PR #2254, opened 2026-09-06
fix(ledger): preserve unknown Byron address attributes

### LINKED ISSUE #2243: ledger/common: preserve unknown Byron address attributes
`ByronAddressAttributes.UnmarshalCBOR` decodes the attribute map into a struct declaring only keys 1 and 2 (`ledger/common/address.go:948-971`). The shared decode mode sets `ExtraReturnErrors: ExtraDecErrorUnknownField` (`cbor/decode.go:51`), so an attribute map carrying any other key fails with `cbor: found unknown field`.

## Reference decoder

`DecCBOR (Attributes AddrAttributes)` handles keys 1 and 2 and returns `Nothing` from the updater for every other key (`eras/byron/ledger/impl/src/Cardano/Chain/Common/AddrAttributes.hs:121-143`). `decCBORAttributes` first decodes the whole map as `Map Word8 LByteString`, then retains every key the updater did not consume in `attrRemain` (`eras/byron/ledger/impl/src/Cardano/Chain/Common/Attributes.hs:213-234`). Unparsed fields are carried through and re-encoded, so an unknown attribute key is forward-compatible by design rather than an error.

## Where the constraint belongs

Nowhere. The reference has no rule rejecting an unknown attribute key, and the attribute bytes participate in the Byron address root hash, so dropping them rather than preserving them would also change the derived address.

Byron addresses are decoded from every Byron transaction output (`ledger/byron/byron.go:840,856`) and from every Shelley-era bootstrap output, so a rejection prevents the containing block from decoding.

## Wire evidence

No fixture in the repository exercises this. An attribute map `{3: <bytes>}` is rejected with `cbor: found unknown field at map element index 0`; `{1: <derivation path>, 3: <bytes>}` fails the same way at index 1. A Byron transaction output whose address carries attribute key 3 fails whole-block decode. That a Byron block on a canonical chain carries such an address is not established here.

## Acceptance criteria

- Decode Byron address attributes as a map from key to raw bytes, interpreting keys 1 and 2 and retaining every other key.
- Re-encode the retained attributes unchanged so the address round-trips byte for byte and its root hash is unaffected.
- Cover an unknown attribute key alone, alongside a known key, inside a Byron transaction output, and a byte-for-byte round trip of each.

wolf31o2:
`ByronAddressAttributes.UnmarshalCBOR` decoded the attribute map into a struct
declaring only keys 1 and 2 (`ledger/common/address.go:948-971`), and the
shared decode mode sets `ExtraReturnErrors: ExtraDecErrorUnknownField`
(`cbor/decode.go:51`), so any other key failed with
`cbor: found unknown field`. A Byron address is decoded from every Byron
transaction output (`ledger/byron/byron.go:840`, `:856`) and from every
Shelley-era bootstrap output, so that failed the whole containing block.

## Reference decoder

`DecCBOR (Attributes AddrAttributes)` handles keys 1 and 2 and returns
`Nothing` from the updater for every other key
(`eras/byron/ledger/impl/src/Cardano/Chain/Common/AddrAttributes.hs:121-143`).
`decCBORAttributes` decodes the whole map as `Map Word8 LByteString` before
interpreting anything, and retains every key the updater did not consume in
`attrRemain` (`.../Common/Attributes.hs:213-234`). `encCBORAttributes` folds the
interpreted keys back into that map, so an unparsed key is written out again.
The note in that module states the intent: the old version "can serialize and
deserialize data received from the new version in a lossless way".

## Where the constraint goes

Nowhere. The reference has no rule rejecting an unknown attribute key.

## What re-encoding has to preserve

The attribute map is hashed into the address, so a dropped key changes the
address rather than only its encoding. `Address.Bytes()` encodes
`[address_root, attributes, addr_type]` and takes a CRC over it, and
`NewByronAddressRedeem` hashes `[type, [type, pubkey], attributes]` through
SHA3-256 and Blake2b-224 to derive the root that a bootstrap or redeem witness
is matched against. Both re-encode the attributes rather than replaying stored
bytes.

Preserved is therefore the whole attribute map, byte for byte:

- Unparsed values are held as their raw `LByteString` contents and written back
  unchanged.
- Key 1 keeps holding the raw derivation path value, and key 2 keeps being
  re-encoded from the decoded `uint32`, both as before. An empty value for key
  1 is now rejected: the reference runs `decodeFull` over it, which cannot
  succeed on an empty input, and an empty value would otherwise be dropped by
  `MarshalCBOR` and break the round trip.
- Key order comes from `cbor.Encode`, which sorts deterministically
  (`cbor/encode.go:40`), matching the ordering of the reference's
  `Map Word8 LByteString`.

`MarshalCBOR` rejects an `Unparsed` entry for key 1 or 2, which is the case
`encCBORAttributes` panics on.

## Compatibility

`ByronAddressAttributes` gains an exported `Unparsed map[uint8][]byte` field.
No signature changes. A consumer that recomputes a Byron address root from
`Address.ByronAttr()` picks the unparsed keys up automatically, including when
it encodes the value rather than a pointer.

## Tests

`ledger/common/byron_address_attributes_test.go`
- `TestByronAddressPreservesUnknownAttributes` builds a Byron address around
  the address root of the mainnet address the block fixture carries, with an
  unknown key alone, alongside a derivation path, alongside a network magic,
  and with three unknown keys, then checks the decoded fields, a byte-for-byte
  round trip through `Bytes()`, and a round trip through the base58 form.
- `TestByronAddressAttributesEncodeAsValue` pins the encoding of a
  `ByronAddressAttributes` value, which is the shape a consumer recomputing an
  address root uses.
- `TestByronAddressAttributesRejectMalformedValues` is the negative control: a
  non-bytestring value, a network magic that is not a `uint32`, an empty
  network magic, an empty derivation path, and a duplicate attribute key.
- `TestByronAddressAttributesRejectShadowedKey` covers a key that is both
  interpreted and carried as unparsed.

`ledger/byron/address_attributes_test.go`
- `TestByronTransactionOutputPreservesUnknownAddressAttributes` decodes a Byron
  transaction output whose address carries attribute key 3 alongside a network
  magic, and checks the round trip.

No fixture in the repository carries a Byron address with an unknown attribute
key. The addresses are constructed around the address root of an existing
mainnet fixture, with the attribute map shaped as the `Map Word8 LByteString`
that `decCBORAttributes` decodes.

---
## Summary by cubic
Fixes #2243. Byron addresses with unknown attribute keys no longer fail decoding: previously any key other than 1 or 2 caused `cbor: found unknown field` and failed the containing block. Decoded attribute maps now replay their original bytes byte-for-byte, so even non-canonical encodings round-trip exactly, and `ByronAddressAttributes.Unparsed` exposes unknown keys for consumers recomputing an address root. Empty derivation paths and empty or malformed network magic values are rejected, and `MarshalCBOR` rejects a key that is both parsed and unparsed.

<sup>Written for commit 5834de3725b99f8dc328f5c492787b12a9d78683. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2254?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ledger/common/address.go
+++ b/ledger/common/address.go
@@ -952,6 +952,7 @@ type byronAddressPayload struct {
 }
 
 type ByronAddressAttributes struct {
+	cbor.DecodeStoreCbor
 	Payload []byte
 	Network *uint32
 	// Unparsed holds the attribute keys this decoder does not interpret,
@@ -975,6 +976,7 @@ func (a *ByronAddressAttributes) UnmarshalCBOR(data []byte) error {
 	if _, err := cbor.Decode(data, &tmpData); err != nil {
 		return err
 	}
+	a.SetCbor(data)
 	a.Payload = nil
 	a.Network = nil
 	a.Unparsed = nil
@@ -1011,7 +1013,10 @@ func (a *ByronAddressAttributes) UnmarshalCBOR(data []byte) error {
 	return nil
 }
 
-func (a *ByronAddressAttributes) MarshalCBOR() ([]byte, error) {
+func (a ByronAddressAttributes) MarshalCBOR() ([]byte, error) {
+	if data := a.Cbor(); data != nil {
+		return data, nil
+	}
 	tmpData := make(map[uint8][]byte, len(a.Unparsed)+2)
 	for key, value := range a.Unparsed {
 		if key == byronAddressAttrDerivationPath ||
--- a/ledger/common/address.go
+++ b/ledger/common/address.go
@@ -1014,11 +1014,7 @@ func (a *ByronAddressAttributes) UnmarshalCBOR(data []byte) error {
 }
 
 func (a ByronAddressAttributes) MarshalCBOR() ([]byte, error) {
-	if data := a.Cbor(); data != nil {
-		return data, nil
-	}
-	tmpData := make(map[uint8][]byte, len(a.Unparsed)+2)
-	for key, value := range a.Unparsed {
+	for key := range a.Unparsed {
 		if key == byronAddressAttrDerivationPath ||
 			key == byronAddressAttrNetworkMagic {
 			return nil, fmt.Errorf(
@@ -1026,6 +1022,12 @@ func (a ByronAddressAttributes) MarshalCBOR() ([]byte, error) {
 				key,
 			)
 		}
+	}
+	if data := a.Cbor(); data != nil {
+		return data, nil
+	}
+	tmpData := make(map[uint8][]byte, len(a.Unparsed)+2)
+	for key, value := range a.Unparsed {
 		tmpData[key] = value
 	}
 	if len(a.Payload) > 0 {
--- a/ledger/common/address.go
+++ b/ledger/common/address.go
@@ -52,6 +52,12 @@ const (
 
 	AddressTypeScriptBit = 0x01
 
+	// Byron address attribute keys, from the EncCBOR and DecCBOR instances
+	// for Attributes AddrAttributes (cardano-ledger
+	// eras/byron/ledger/impl/src/Cardano/Chain/Common/AddrAttributes.hs)
+	byronAddressAttrDerivationPath = 1
+	byronAddressAttrNetworkMagic   = 2
+
 	ByronAddressTypePubkey = 0
 	ByronAddressTypeScript = 1
 	ByronAddressTypeRedeem = 2
@@ -948,38 +954,84 @@ type byronAddressPayload struct {
 type ByronAddressAttributes struct {
 	Payload []byte
 	Network *uint32
+	// Unparsed holds the attribute keys this decoder does not interpret,
+	// mapped to their raw values. The reference updater returns Nothing for
+	// every key other than 1 and 2
+	// (cardano-ledger
+	// eras/byron/ledger/impl/src/Cardano/Chain/Common/AddrAttributes.hs:121-143),
+	// and decCBORAttributes retains those keys in attrRemain, which
+	// encCBORAttributes writes back
+	// (.../Common/Attributes.hs:213-234). The attribute map is hashed into the
+	// Byron address root, so an unknown key that is dropped rather than
+	// carried through changes the address it belongs to.
+	Unparsed map[uint8][]byte
 }
 
 func (a *ByronAddressAttributes) UnmarshalCBOR(data []byte) error {
-	var tmpData struct {
-		Payload    []byte `cbor:"1,keyasint,omitempty"`
-		NetworkRaw []byte `cbor:"2,keyasint,omitempty"`
-	}
+	// decCBORAttributes decodes the whole map as Map Word8 LByteString before
+	// interpreting any key, so an unrecognised key is data rather than an
+	// error.
+	var tmpData map[uint8][]byte
 	if _, err := cbor.Decode(data, &tmpData); err != nil {
 		return err
 	}
-	a.Payload = tmpData.Payload
-	if len(tmpData.NetworkRaw) > 0 {
-		var tmpNetwork uint32
-		if _, err := cbor.Decode(tmpData.NetworkRaw, &tmpNetwork); err != nil {
-			return err
+	a.Payload = nil
+	a.Network = nil
+	a.Unparsed = nil
+	for key, value := range tmpData {
+		switch key {
+		case byronAddressAttrDerivationPath:
+			// The reference runs decodeFull over this value, which cannot
+			// succeed on an empty one, and an empty value would also be
+			// dropped by MarshalCBOR and so break the round trip.
+			if len(value) == 0 {
+				return errors.New(
+					"invalid Byron address attributes: empty derivation path",
+				)
+			}
+			a.Payload = value
+		case byronAddressAttrNetworkMagic:
+			if len(value) == 0 {
+				return errors.New(
+					"invalid Byron address attributes: empty network magic",
+				)
+			}
+			var tmpNetwork uint32
+			if _, err := cbor.Decode(value, &tmpNetwork); err != nil {
+				return err
+			}
+			a.Network = &tmpNetwork
+		default:
+			if a.Unparsed == nil {
+				a.Unparsed = make(map[uint8][]byte)
+			}
+			a.Unparsed[key] = value
 		}
-		a.Network = &tmpNetwork
 	}
 	return nil
 }
 
 func (a *ByronAddressAttributes) MarshalCBOR() ([]byte, error) {
-	tmpData := make(map[int]any)
+	tmpData := make(map[uint8][]byte, len(a.Unparsed)+2)
+	for key, value := range a.Unparsed {
+		if key == byronAddressAttrDerivationPath ||
+			key == byronAddressAttrNetworkMagic {
+			return nil, fmt.Errorf(
+				"byron address attribute %d is both parsed and unparsed",
+				key,
+			)
+		}
+		tmpData[key] = value
+	}
 	if len(a.Payload) > 0 {
-		tmpData[1] = a.Payload
+		tmpData[byronAddressAttrDerivationPath] = a.Payload
 	}
 	if a.Network != nil {
 		networkRaw, err := cbor.Encode(a.Network)
 		if err != nil {
 			return nil, err
 		}
-		tmpData[2] = networkRaw
+		tmpData[byronAddressAttrNetworkMagic] = networkRaw
 	}
 	return cbor.Encode(tmpData)
 }
--- a/ledger/byron/address_attributes_test.go
+++ b/ledger/byron/address_attributes_test.go
@@ -1,0 +1,87 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


ledger/common/address.go
```diff
@@ -946,40 +952,93 @@ type byronAddressPayload struct {
...
+	for key := range a.Unparsed {
+		if key == byronAddressAttrDerivationPath ||
+			key == byronAddressAttrNetworkMagic {
+			return nil, fmt.Errorf(
+				"byron address attribute %d is both parsed and unparsed",
+				key,
+			)
+		}
+	}
+	if data := a.Cbor(); data != nil {
```
chrisguiney, line 1026: Non-blocking. The cached-bytes path makes `MarshalCBOR` ignore every mutation of a decoded value, not only a shadowed key. Reproduced at this head: decode `a202410203420102`, set `Network` to 42, and `cbor.Encode` still returns `a202410203420102`; `NewByronAddressFromParts` with those attributes then yields the original address bytes and the original base58 rather than one for magic 42. Setting `Payload` or adding `Unparsed[9]` behaves the same way.

Caching is the right call for the decode path — the gap is that nothing invalidates it. Clearing the stored CBOR in `NewByronAddressFromParts`/`NewByronAddressRedeem`, or documenting on `ByronAttr()` that the returned value is replay-only, would close it. No caller in gouroboros, dingo, adder or cardano-node-api mutates a decoded value today.

ledger/common/address.go
```diff
@@ -946,40 +952,93 @@ type byronAddressPayload struct {
...
+	// mapped to their raw values. The reference updater returns Nothing for
+	// every key other than 1 and 2
+	// (cardano-ledger
+	// eras/byron/ledger/impl/src/Cardano/Chain/Common/AddrAttributes.hs:121-143),
+	// and decCBORAttributes retains those keys in attrRemain, which
+	// encCBORAttributes writes back
+	// (.../Common/Attributes.hs:213-234). The attribute map is hashed into the
+	// Byron address root, so an unknown key that is dropped rather than
+	// carried through changes the address it belongs to.
+	Unparsed map[uint8][]byte
```
chrisguiney, line 968: Non-blocking. `Unparsed` gives `ByronAttr()`'s shallow copy shared mutable state with the `Address`: `a.ByronAttr().Unparsed[9] = []byte{0xff}` writes through, and a later `a.ByronAttr().Unparsed` reads back `map[3:[1 2] 9:[255]]`. Two goroutines reading the same decoded `Address` and writing to that map race. `Payload` and `Network` had no equivalent hazard. Copying the map in `ByronAttr()` would keep the value-return contract intact.

ledger/common/address.go
```diff
@@ -946,40 +952,93 @@ type byronAddressPayload struct {
...
+			}
+			a.Payload = value
+		case byronAddressAttrNetworkMagic:
+			if len(value) == 0 {
+				return errors.New(
+					"invalid Byron address attributes: empty network magic",
+				)
+			}
+			var tmpNetwork uint32
+			if _, err := cbor.Decode(value, &tmpNetwork); err != nil {
```
chrisguiney, line 1002: Non-blocking, pre-existing. The reference decodes this value with `decodeFullDecoder byronProtVer "NetworkMagic" decodeWord32Canonical`, which requires canonical form and full consumption of the `LByteString`. `cbor.Decode` requires neither: `{2: h'020101'}` decodes to magic 2 with the trailing bytes ignored, and `{2: h'1802'}` decodes to 2 from a non-canonical head. Both are accepted where cardano-ledger rejects.

Key 1 has the same shape — the reference runs `decodeFull` as `HDAddressPayload`, i.e. `decodeBytesCanonical`, so the `4a1c010203040506070809` fixture in the new test is not a value the reference would accept (`0x1c` is a reserved CBOR head); the existing `TestByronAddressFromParts` uses the realistic `581cea14...` shape.

Byte-preserving re-encode removes the address-corruption consequence this had on main, so what is left is only a permissiveness gap. Raising it because `TestByronAddressAttributesRejectMalformedValues` reads as pinning the whole contract for these values.

chrisguiney (approved):
Approving. Reviewed head 5834de3 in an isolated worktree against merge base e6c5679.

**Bot standing on this head.** CodeRabbit did not review this PR: its only comment is a "Review limit reached" rate-limit notice, and the green check reads "Review rate limited". Cubic did run on 5834de3 — `output.summary` is "AI review completed with 1 review. 0 issues found across 2 files (changes from recent commits)", i.e. an incremental pass over the 6b9af42..5834de3 delta only. Both earlier Cubic findings are resolved; I re-checked each against this head. This review stands in for the absent CodeRabbit pass.

**Reference behaviour** (cardano-ledger `Cardano/Chain/Common/Attributes.hs` `decCBORAttributes`/`encCBORAttributes`, `AddrAttributes.hs` `DecCBOR (Attributes AddrAttributes)`): the map is decoded as `Map Word8 LByteString`, keys 1 and 2 are interpreted, every other key is retained in `attrRemain` and written back. The diff matches. Rejecting empty key-1 and key-2 values matches `decodeFull`, which cannot succeed on empty input.

**Byte-exactness.** A non-canonical map header (`b802`) and an indefinite-length byte-string value both round-trip unchanged; key order is replayed from the stored bytes rather than re-sorted. `Word8`-only keys are enforced (256, negative and text keys rejected), and duplicate keys are rejected by `DupMapKeyEnforcedAPF`.

**No regression for the common case.** `82d818584283581cc9d7d0c4b913b970...` taken from the real epoch-1 slot-21603 Byron block fixture, the mainnet derivation-path address and the preview network-magic address all produce byte-identical `Bytes()` and identical base58 on this branch and on the merge base.

**Fail-before.** Reverting `UnmarshalCBOR`/`MarshalCBOR` to the base bodies in place (fields kept declared so the tests compile) fails `TestByronAddressPreservesUnknownAttributes` (4 of 5 subtests), `TestByronAddressPreservesNonCanonicalAttributeEncoding`, `TestByronAddressAttributesEncodeAsValue`, both shadowed-key tests and `TestByronTransactionOutputPreservesUnknownAddressAttributes`. The `no attributes`, `unknown key with text value`, `network magic with bad value` and `duplicate key` subtests pass on the reverted code, so they are guards rather than discriminators.

**API.** Moving `MarshalCBOR` to a value receiver is behaviour-neutral: on the merge base fxamacker already dispatched the pointer method for a non-addressable value, so `cbor.Encode(attr)` produced the same bytes either way. Every construction site in gouroboros, dingo, adder and cardano-node-api uses keyed literals, so the two added fields break nothing.

**Class.** `ledger/common/address.go` holds the only Byron address attribute decoder in the repository — no per-era copies, and both Byron outputs and Shelley bootstrap outputs route through `common.Address`. Class complete.

**Checks.** `go build ./...`, `go vet ./...`, `go test ./...`, `go test -race ./ledger/common ./ledger/byron`, `golangci-lint run ./...` (0 issues, isolated cache), nilaway (clean, verified against a positive control), and build/vet/test for all eight `examples/*` modules: all pass. One `pipeline` timing flake on the first full run; that package is untouched by the diff and passes `-count=5` on both this branch and the merge base.

Three non-blocking notes inline. Separately, pre-existing and outside this diff: `UtxoValidateOutputBootAddrAttrsTooBig` measures `len(cbor.Encode(attrs))`, while the reference `bootstrapAddressAttrsSize` measures the `HDAddressPayload` contents plus `unknownAttributesLength` — 70 against 64 for a 64-byte derivation path. Worth its own issue.

ledger/common/address.go
```diff
@@ -948,38 +954,84 @@ type byronAddressPayload struct {
...
 
 func (a *ByronAddressAttributes) UnmarshalCBOR(data []byte) error {
-	var tmpData struct {
-		Payload    []byte `cbor:"1,keyasint,omitempty"`
-		NetworkRaw []byte `cbor:"2,keyasint,omitempty"`
-	}
+	// decCBORAttributes decodes the whole map as Map Word8 LByteString before
+	// interpreting any key, so an unrecognised key is data rather than an
+	// error.
+	var tmpData map[uint8][]byte
```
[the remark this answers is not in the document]
chrisguiney: Confirmed fixed at 5834de3. `UnmarshalCBOR` stores the attribute-map bytes and `MarshalCBOR` replays them, so the wire form survives: a non-canonical map header (`b802...`) and an indefinite-length byte-string value both round-trip byte for byte through `Address.Bytes()`, and key order comes from the stored bytes rather than `cbor.Encode`'s deterministic sort. Verified on this head; no further action.

ledger/common/address.go
```diff
@@ -946,40 +952,91 @@ type byronAddressPayload struct {
...
 		}
-		a.Network = &tmpNetwork
 	}
 	return nil
 }
 
-func (a *ByronAddressAttributes) MarshalCBOR() ([]byte, error) {
-	tmpData := make(map[int]any)
+func (a ByronAddressAttributes) MarshalCBOR() ([]byte, error) {
+	if data := a.Cbor(); data != nil {
```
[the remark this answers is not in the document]
chrisguiney: Confirmed fixed at 5834de3: the key-1/key-2 collision check now runs before the cached-bytes return, and `TestDecodedByronAddressAttributesRejectShadowedKey` covers it. Verified it fails without the fix.

The narrower case is closed but the general one is not: the cached path still ignores mutations of `Payload`, `Network` and non-reserved `Unparsed` keys. Raised as a non-blocking inline note on line 1026 with a reproduction.

PR #2256, opened 2026-09-06
fix(ledger): widen operational certificate counter and KES period

### LINKED ISSUE #2245: ledger: widen operational certificate counter and KES period to the reference types
The operational certificate's sequence number and KES period are `uint32` in every Shelley-family header: `ledger/shelley/shelley.go:230-231` and `:249-250`, and `ledger/babbage/babbage.go:328-329`, which Conway and Dijkstra inherit.

## Reference decoder

`decodeOCertFields` decodes the counter as `Word64` and the period as `KESPeriod`, a `Word` newtype, both with plain `decCBOR` and no bound (`libs/cardano-protocol/src/Cardano/Protocol/TPraos/OCert.hs:80-93`, `:122-128`). The CDDL agrees: `sequence_number = uint .size 8` and `kes_period = uint .size 8` (`eras/dijkstra/impl/cddl/data/dijkstra.cddl:88,90`).

## Where the constraint belongs

The counter is bounded relative to ledger state by the OCERT rule, not by the type, so a header carrying a value above 2^32 decodes and is then rejected as invalid. gouroboros instead fails the decode.

The distinction matters on the ChainSync path: `handleRollForward` returns the decode error out of the protocol handler (`protocol/chainsync/client.go:1019-1027`), tearing down the connection rather than rejecting one header. A peer can send such a header.

## Wire evidence

Taking the upstream `Header_Shelley`, `Header_Alonzo`, `Header_Babbage` and `Header_Conway` goldens and setting `sequence_number` to 2^32 makes all four fail to decode with `4294967296 overflows uint32`. The unmodified goldens decode. The reference decodes both.

## Acceptance criteria

- Type the operational certificate sequence number and KES period as `uint64` in the Shelley and Babbage header bodies.
- Keep the OCERT counter and KES period checks where they are, against ledger state.
- Cover a header with a sequence number and a KES period above 2^32 decoding successfully in Shelley, Alonzo, Babbage, Conway and Dijkstra, and being rejected by the operational certificate validation rather than by the decoder.

wolf31o2:
The operational certificate sequence number and KES period were `uint32` in every Shelley-family header body (`ledger/shelley/shelley.go:230-231`, `:249-250`) and in `BabbageOpCert` (`ledger/babbage/babbage.go:328-329`), which Conway and Dijkstra inherit.

## Reference decoder

`decodeOCertFields` decodes the counter as `Word64` and the period as `KESPeriod`, a `Word` newtype, each with a plain `decCBOR` and no bound (`libs/cardano-protocol/src/Cardano/Protocol/TPraos/OCert.hs:80-93`, `:122-128`). The CDDL agrees: `sequence_number = uint .size 8` and `kes_period = uint .size 8` (`eras/dijkstra/impl/cddl/data/dijkstra.cddl:88,90`).

## Where the constraint belongs

Nowhere in the decoder. The counter is bounded against ledger state by the OCERT rule and the period by the KES expiry check, both of which run after the header decodes. `ledger.ValidateKesPeriod` (`ledger/verify_opcert.go:118`) already rejects a period ahead of the block's own slot, and its underflow guard at `:134` holds for any 64-bit value. What still holds the line at decode is the CBOR major type: the fields remain unsigned integers, so a negative value, a byte string, and a 2^64 bignum are still rejected.

## Wire evidence

Taking the upstream `Header_*` goldens and setting `sequence_number` to 2^32 and `kes_period` to 2^32+7 makes all seven fail with `4294967296 overflows uint32`. The unmodified goldens decode. The reference decodes both.

Because this fails at decode, `handleRollForward` returns the error out of the ChainSync protocol handler (`protocol/chainsync/client.go:1019-1027`), tearing down the connection rather than rejecting one header.

## Changes

- `ledger/shelley/shelley.go`, `ledger/babbage/babbage.go`: the counter and period are `uint64` in the header body and its wire struct.
- `ledger/verify_opcert.go`: `OpCertExtractor` returns `uint64` for both, so an implementer built on a decoded header no longer has to narrow.
- `consensus/validate.go`: `ValidateHeaderInput` carries both as `uint64`. `common.OpCertSignableBytes` already serialized them as big-endian `uint64`, matching `OCertSignable`, so the signed bytes are unchanged for every value below 2^32.
- `consensus/block.go`: `OperationalCert` and `HeaderBody` carry both as `uint64`. CBOR integer encoding is minimal, so a produced header is byte-identical for any value that fit the old type.
- `ledger/verify_kes.go`: drops the conversions the narrower fields required.

## Tests

`internal/test/conformance/opcert_width_conformance_test.go`:

- `TestGoldenHeaderOCertCounterAndKesPeriodWidth` rewrites only the counter and period in each of the seven upstream `Header_*` goldens, decodes the result in Shelley, Allegra, Mary, Alonzo, Babbage, Conway and Dijkstra, checks both values survive, and checks the certificate is then rejected by `ledger.ValidateKesPeriod` rather than by the decoder. Every era fails this on `main` with `4294967296 overflows uint32`.
- `TestGoldenHeaderOCertFieldsRejectNonIntegers` keeps the negative controls: a negative counter, a byte-string counter and a 2^64 bignum are still rejected in both header shapes.

## Downstream

`blinklabs-io/dingo` passes the Shelley-family fields to `shelleyOpCert` (`ledger/verify_opcert.go:97-102`), whose parameters are `uint32`, and builds `babbage.BabbageOpCert` with `uint32(opCert.IssueNumber)` and `uint32(opCert.KESPeriod)` (`ledger/forging/builder.go:912-913`, `:935-936`). Both take the `uint64` value directly instead; the second drops a narrowing conversion on the forging path. Its other call sites — `ledger/verify_opcert.go:114-115`, `api/blockfrost/adapter.go:508-532`, `consensus/praos/view.go:164-179` — already widened to `uint64` and are unaffected.

Fixes #2245

---
## Summary by cubic
Widen the operational certificate counter and KES period from `uint32` to `uint64` in every Shelley-family header body. A header carrying either value at or above 2^32 previously failed to decode and tore down the ChainSync connection; now it decodes and is rejected by `ValidateKesPeriod` or the OCERT rule, matching the reference decoder. Signed bytes and produced headers are byte-identical for every value that fit the old type, and non-integer CBOR values are still rejected at decode.

**Tests**
- A conformance test rewrites the counter and period in all seven upstream header goldens, checks the widened values decode in every era and are then rejected by `ValidateKesPeriod` instead of the decoder.
- Negative controls keep a negative value, byte string, and 2^64 bignum rejected at decode in either field.

**Downstream**
- `blinklabs-io/dingo`'s `shelleyOpCert` call site and the forging path take the `uint64` value directly instead of narrowing to `uint32`; its other call sites already used `uint64`.

<sup>Written for commit 094f76e8eb414dd2507bc1b1b50df067237d9a56. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2256?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/consensus/block_test.go
+++ b/consensus/block_test.go
@@ -859,8 +859,8 @@ func TestBuildHeaderTPraosRoundTripsWithHeaderValidator(t *testing.T) {
 	coldPrivateKey := ed25519.NewKeyFromSeed(coldSeed)
 	coldPublicKey := coldPrivateKey.Public().(ed25519.PublicKey)
 
-	opCertSeqNum := uint32(1)
-	opCertKesPeriod := uint32(0)
+	opCertSeqNum := uint64(1)
+	opCertKesPeriod := uint64(0)
 	opCertBody := common.OpCertSignableBytes(
 		kesSigner.PublicKey(),
 		uint64(opCertSeqNum),
--- a/internal/test/conformance/opcert_width_conformance_test.go
+++ b/internal/test/conformance/opcert_width_conformance_test.go
@@ -1,0 +1,352 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package conformance
+
+import (
+	"path"
+	"testing"
+
+	"github.com/blinklabs-io/gouroboros/cbor"
+	"github.com/blinklabs-io/gouroboros/ledger"
+	"github.com/blinklabs-io/gouroboros/ledger/allegra"
+	"github.com/blinklabs-io/gouroboros/ledger/alonzo"
+	"github.com/blinklabs-io/gouroboros/ledger/babbage"
+	"github.com/blinklabs-io/gouroboros/ledger/common"
+	"github.com/blinklabs-io/gouroboros/ledger/conway"
+	"github.com/blinklabs-io/gouroboros/ledger/dijkstra"
+	"github.com/blinklabs-io/gouroboros/ledger/mary"
+	"github.com/blinklabs-io/gouroboros/ledger/shelley"
+	"github.com/blinklabs-io/ouroboros-mock/fixtures"
+)
+
+// goldenHeaderDir is the CardanoNodeToNodeVersion2 golden directory in the
+// embedded upstream fixtures. Each Header_* golden is
+// [era_id, #6.24(bytes .cbor header)].
+const goldenHeaderDir = "upstream/ouroboros-consensus/ouroboros-consensus-cardano/golden/cardano/CardanoNodeToNodeVersion2"
+
+// tPraosOCertOffset is the index of the operational certificate sequence
+// number in the flat Shelley-family (TPraos) header body array; the KES period
+// is the element after it. Shelley through Alonzo share the shape.
+const tPraosOCertOffset = 10
+
+// cPraosOCertIndex is the index of the nested operational certificate array in
+// the Babbage-family (CPraos) header body array. Within that array the
+// sequence number is element 1 and the KES period element 2.
+const cPraosOCertIndex = 8
+
+// readGoldenHeader returns the era id and header CBOR carried by an upstream
+// Header_* golden.
+func readGoldenHeader(t *testing.T, name string) (uint, []byte) {
+	t.Helper()
+	raw, err := fixtures.EmbeddedFixtures().
+		ReadFile(path.Join(goldenHeaderDir, name))
+	if err != nil {
+		t.Fatalf("read golden %s: %v", name, err)
+	}
+	var envelope struct {
+		cbor.StructAsArray
+		EraId  uint
+		Header cbor.Tag
+	}
+	if _, err := cbor.Decode(raw, &envelope); err != nil {
+		t.Fatalf("decode golden envelope %s: %v", name, err)
+	}
+	headerCbor, ok := envelope.Header.Content.([]byte)
+	if !ok {
+		t.Fatalf(
+			"golden %s header tag content is %T, want []byte",
+			name,
+			envelope.Header.Content,
+		)
+	}
+	return envelope.EraId, headerCbor
+}
+
+// setOCertFields rewrites the operational certificate sequence number and KES
+// period in a header, leaving every other byte of the header untouched.
+func setOCertFields(
+	t *testing.T,
+	headerCbor []byte,
+	tpraos bool,
+	sequenceNumber uint64,
+	kesPeriod uint64,
+) []byte {
+	t.Helper()
+	var header []cbor.RawMessage
+	if _, err := cbor.Decode(headerCbor, &header); err != nil {
+		t.Fatalf("decode header array: %v", err)
+	}
+	if len(header) != 2 {
+		t.Fatalf("header array has %d elements, want 2", len(header))
+	}
+	var body []cbor.RawMessage
+	if _, err := cbor.Decode(header[0], &body); err != nil {
+		t.Fatalf("decode header body array: %v", err)
+	}
+	encode := func(v uint64) cbor.RawMessage {
+		encoded, err := cbor.Encode(v)
+		if err != nil {
+			t.Fatalf("encode %d: %v", v, err)
+		}
+		return encoded
+	}
+	if tpraos {
+		if len(body) <= tPraosOCertOffset+1 {
+			t.Fatalf("TPraos header body has %d elements", len(body))
+		}
+		body[tPraosOCertOffset] = encode(sequenceNumber)
+		body[tPraosOCertOffset+1] = encode(kesPeriod)
+	} else {
+		if len(body) <= cPraosOCertIndex {
+			t.Fatalf("CPraos header body has %d elements", len(body))
+		}
+		var ocert []cbor.RawMessage
+		if _, err := cbor.Decode(body[cPraosOCertIndex], &ocert); err != nil {
+			t.Fatalf("decode operational certificate array: %v", err)
+		}
+		if len(ocert) != 4 {
+			t.Fatalf(
+				"operational certificate array has %d elements, want 4",
+				len(ocert),
+			)
+		}
+		ocert[1] = encode(sequenceNumber)
+		ocert[2] = encode(kesPeriod)
+		encodedOCert, err := cbor.Encode(ocert)
+		if err != nil {
+			t.Fatalf("encode operational certificate array: %v", err)
+		}
+		body[cPraosOCertIndex] = encodedOCert
+	}
+	encodedBody, err := cbor.Encode(body)
+	if err != nil {
+		t.Fatalf("encode header body array: %v", err)
+	}
+	header[0] = encodedBody
+	encodedHeader, err := cbor.Encode(header)
+	if err != nil {
+		t.Fatalf("encode header array: %v", err)
+	}
+	return encodedHeader
+}
+
+// headerOCertFields returns the decoded operational certificate sequence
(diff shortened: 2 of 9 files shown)


ledger/verify_opcert.go
```diff
@@ -204,8 +204,8 @@ func ValidateOpCert(
 // This is a convenience function for working with block headers.
 type OpCertExtractor interface {
 	OpCertHotVkey() []byte
-	OpCertSequenceNumber() uint32
-	OpCertKesPeriod() uint32
+	OpCertSequenceNumber() uint64
```
chrisguiney, line 207: Non-blocking, API compatibility.

This changes an exported interface's method signatures, and the header body field types change with it, so the release is source-breaking for consumers. Checked `dingo` at `9881e26c`: `shelleyOpCert(hotVkey []byte, sequenceNumber uint32, kesPeriod uint32, signature []byte)` (`ledger/verify_opcert.go:81`) stops compiling once `ShelleyBlockHeaderBody.OpCertSequenceNumber` is `uint64`. Its other sites survive — `uint64(oc.SequenceNumber)` in `babbageOpCert` and `uint32(opCert.IssueNumber)` in `ledger/forging/builder.go:762,785` both still compile, the latter now narrowing a value the wire can carry.

Fine for v0.x with the matching consumer change, but the tag needs to land with it rather than ahead of it.

internal/test/conformance/opcert_width_conformance_test.go
```diff
@@ -0,0 +1,362 @@
...
+// certificate sequence number and KES period at their reference widths.
+//
+// cardano-ledger decodes the counter as Word64 and the period as a Word
+// newtype with plain decCBOR and no bound (decodeOCertFields,
+// libs/cardano-protocol/src/Cardano/Protocol/TPraos/OCert.hs), and the CDDL
+// declares sequence_number = uint .size 8 and kes_period = uint .size 8. A
+// header carrying a counter or period at or above 2^32 therefore decodes; the
+// counter is bounded against ledger state by the OCERT rule and the period by
+// the KES expiry check, not by the wire type.
+func TestGoldenHeaderOCertCounterAndKesPeriodWidth(t *testing.T) {
```
chrisguiney, line 210: Non-blocking, coverage.

This pins the decode side across all seven eras. The produce side is unpinned: nothing asserts that a header built through `serializeHeaderBodyTPraos` / `serializeHeaderBodyCPraos` with a counter or period at or above 2^32 decodes back to the same value. I encoded both shapes at 2^32+7 and decoded into `ShelleyBlockHeaderBody` and `BabbageOpCert` — it round-trips, so this is coverage rather than a defect. But the description's claim about produced headers rests on no test in the suite, and `TestBuildHeaderTPraosRoundTripsWithHeaderValidator` is one line away from covering it.

consensus/block_test.go
```diff
@@ -859,8 +859,8 @@ func TestBuildHeaderTPraosRoundTripsWithHeaderValidator(t *testing.T) {
 	coldPrivateKey := ed25519.NewKeyFromSeed(coldSeed)
 	coldPublicKey := coldPrivateKey.Public().(ed25519.PublicKey)
 
-	opCertSeqNum := uint32(1)
-	opCertKesPeriod := uint32(0)
+	opCertSeqNum := uint64(1)
+	opCertKesPeriod := uint64(0)
 	opCertBody := common.OpCertSignableBytes(
 		kesSigner.PublicKey(),
 		uint64(opCertSeqNum),
```
chrisguiney, line 866: Nit. `opCertSeqNum` and `opCertKesPeriod` are `uint64` as of the two lines above, so these conversions are no-ops. Eight sites in all: here and at `consensus/validate_test.go:502-503`, `:614-615`, `:988-989`. `unconvert` is off in the repo's `golangci-lint` config, so nothing flags them.

chrisguiney (approved):
Reviewed at `094f76e` in an isolated worktree, and again merged onto current `origin/main` (`f7fb79b`, which moved `consensus/validate.go` and `ledger/verify_opcert.go` to `ed25519strict` under this branch).

**Behavior.** The widened counter and period decode in all seven eras and are then rejected downstream of the decoder, not by it. Reverting only the decode width in place — `shelleyBlockHeaderBodyWire` back to `uint32`, plus a `uint32` wire struct behind `BabbageOpCert.UnmarshalCBOR`, leaving the test untouched — fails all seven subtests with `4294967296 overflows uint32`. The new test earns its keep.

**Wire.** This repo's own vendored vector agrees: `sequence_number = uint .size 8` and `kes_period = uint .size 8` in `ledger/dijkstra/testdata/dijkstra.cddl:86,88`. `cbor.Encode` is shortest-form, so `uint32` and `uint64` encode identically for every value below 2^32, and all seven `Header_*` goldens still re-encode byte-identically at this head. `common.OpCertSignableBytes` already took `uint64` and writes both fields big-endian 64-bit, so the signed bytes are unchanged.

**Era audit.** The two fields live in exactly two struct definitions — `ShelleyBlockHeaderBody` with its wire twin, and `BabbageOpCert` — which Allegra/Mary/Alonzo and Conway/Dijkstra reuse. No other Shelley- or Babbage-family header field is narrower than its CDDL width, so the class is complete.

**Negative controls.** Each rejects for the right reason, not incidentally: a negative integer, a byte string, and a 2^64 bignum all fail at the `uint64` field in both header shapes, and an unmodified header through the same re-encode pipeline still decodes.

**Underflow.** Every consumer of the widened period guards `current < opcert` before subtracting (`ledger/verify_opcert.go:134`, `ledger/verify_kes.go:138`, `consensus/validate.go:462,515`), and the evolution period stays bounded by the slot, so nothing new can wrap.

| Check | Exit |
|---|---|
| `go build ./...` | 0 |
| `go vet ./...` | 0 |
| `go test -race ./...` (39 pkgs) | 0 |
| `go test -race ./...` merged onto `f7fb79b` (40 pkgs) | 0 |
| `golangci-lint run ./...` (v2.13.2, isolated cache) | 0 issues |
| `nilaway -include-pkgs=github.com/blinklabs-io/gouroboros ./...` | 0, positive control confirmed it analysed |
| `go mod tidy` | clean |
| 8 nested `examples/*/go.mod` build + test (all `replace` to `../..`) | 0 |
| `gofmt -l .` | empty |

Bot standing: CodeRabbit never reviewed this diff — its only comment is the rate-limit warning, so its green check is not a review. Cubic reviewed and its two P3 notes are fixed at this head (verified below). Three non-blocking notes inline; none blocks merge.

internal/test/conformance/opcert_width_conformance_test.go
```diff
@@ -0,0 +1,352 @@
...
+					}
+				})
+			}
+		})
+	}
+}
+
+// setOCertRaw replaces the operational certificate sequence number with
+// arbitrary CBOR.
+func setOCertRaw(
```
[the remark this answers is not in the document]
chrisguiney: Confirmed fixed at 094f76e. The traversal and re-encode now live once in `setOCertRawFields`, which writes either element only when its replacement is non-nil; `setOCertFields` is a thin wrapper that encodes two uint64s and delegates. Both tests go through the single path, and the unmodified header still decodes after a round trip through it.

internal/test/conformance/opcert_width_conformance_test.go
```diff
@@ -0,0 +1,352 @@
...
+		name   string
+		golden string
+		tpraos bool
+	}{
+		{name: "Shelley", golden: "Header_Shelley", tpraos: true},
+		{name: "Babbage", golden: "Header_Babbage"},
+	} {
+		t.Run(testDef.name, func(t *testing.T) {
+			eraId, headerCbor := readGoldenHeader(t, testDef.golden)
+			blockType := ledger.BlockHeaderToBlockTypeMap[eraId]
```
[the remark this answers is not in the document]
chrisguiney: Confirmed fixed at 094f76e: `opcert_width_conformance_test.go:300-303` now checks `ok` and `t.Fatalf`s, matching the width test. Verified the controls are not passing trivially — each bad value fails at the opcert field itself (`cannot unmarshal negative integer / byte string / tag ... of type uint64`) in both header shapes, and the same pipeline with no substitution still decodes.

Issue #290, opened 2023-06-03
better handling for Shelley addresses


agaffney:
We currently handle addresses in TX outputs as a bytestring, but we should instead parse into an object that actually understands it, whether it's something custom or using a third-party library

agaffney:
https://cips.cardano.org/cips/cip19/

PR #2265, opened 2026-09-07
fix(ledger)!: make the recorded DRep deposit optional and fail closed on absence

wolf31o2:
`DRepRegistration.Deposit` was a non-optional `uint64`, so a state reporting a registered DRep without a recorded deposit yielded an authoritative zero and `UtxoValidateCertificateDeposits` rejected a deregistration supplying its real refund:

```
incorrect refund for certificate type 17: supplied <deposit>, expected 0
```

- `Deposit` is now `*uint64`; `nil` is absence, distinct from a recorded zero.
- A registration reported without a deposit returns `DRepDepositStateInconsistentError` instead of comparing the refund against zero.
- In-transaction DRep registrations are keyed by the full credential, so a key-hash and script-hash certificate sharing a hash no longer collide.
- `isDRepRegistered` propagates a lookup failure instead of reporting it as an unregistered DRep, which rejected valid transactions with `DelegateVoteToUnregisteredDRepError`.

`DRepRegistration.Credential` stays a `Blake2b224`; widening it breaks every implementor and lands with blinklabs-io/ouroboros-mock#290.


wolf31o2:
Addressed at `64fbace`.

**Lookup errors reported as unregistered** (raised by both reviewers): `isDRepRegistered` returned `err == nil && reg != nil`, so a ledger-state failure became an unregistered DRep and the caller rejected a valid transaction with `DelegateVoteToUnregisteredDRepError`. The error is now returned, matching `loadDRep` and `UtxoValidateUnknownVoters`. This predates the PR — the same expression is on `main` — but it is in the path this change touches.

**Documentation promising full-credential resolution:** valid, and the mismatch was mine. `GovState.DRepRegistration`'s doc said the credential carries its type while the parameter is a bare `Blake2b224`. The doc now records that both this lookup and the `DRepRegistrations` records carry a bare hash, that a same-hash key/script pair resolves to whichever registration the state holds, and that widening both is deferred. The comment in `isDRepRegistered` made the same overclaim and is corrected to describe the in-transaction registration set, which is keyed by the full credential.

**Not restoring the optional capability.** A second public DRep-resolution interface on `GovState`, added so a released test framework keeps compiling, is permanent API surface every implementor has to learn in exchange for a fixture problem. The credential widening is a breaking change; it lands together with the consumer update in `ouroboros-mock` (#290, currently draft), not as a parallel interface. The commits that added and then removed that capability on this branch are `a19965d` and its revert.

Both proposed diffs referenced `DRepCredentialRegistration` and `hasDRepCredentialState`, which the revert removed, so those hunks no longer apply. The underlying finding did apply and is fixed.

Validation at `64fbace`: `go build ./...` 0, `go vet ./...` 0, `go test ./...` 0 across 40 packages, `golangci-lint run ./...` 0 issues, `nilaway -include-pkgs=github.com/blinklabs-io/gouroboros ./...` 0.

wcatz (approved):

Issue #2023, opened 2026-08-24
test: execute production protocol and ledger boundary checks


wolf31o2:
The production-boundary tests are weaker than their names imply:

- Block-body, VRF, and KES edge tests check fixtures without reaching the named production verification functions.
- Deterministic leader-election cases turn failed preconditions into skips or log-only results.
- VRF conformance can skip mismatched vectors without asserting that the required corpus ran.
- Ledger fuzz targets stop after decoding and omit representative post-decode accessors and validation rules.
- The post-stop muxer send test logs an unexpected nil error instead of failing.

Acceptance criteria:

- Invoke the named production validation functions with valid and malformed vectors.
- Fail when required leader-election or VRF vectors are skipped, and assert tested-vector counts.
- Exercise bounded post-decode accessors and validation paths from the fuzz targets.
- Assert that post-stop Send returns an error.

PR #2266, opened 2026-09-07
test: assert the contracts the muxer and VRF conformance tests name

wolf31o2:
Refs #2023.

`TestMuxerSendAfterStop` now requires an error with no attempted connection
write after `Stop`. A successful pre-stop write validates the counter, so
asynchronous connection cleanup cannot mask a missing shutdown guard.

Both VRF conformance tests require the complete 29-vector corpus. Proof
generation fails on any derived-key mismatch instead of skipping the vector.

Mutation checks detect removal of the Send shutdown guard, a single-vector
KeyGen regression, and removal of a corpus entry. The full test suite,
conformance suite, build and lint pass.

### DIFF
--- a/internal/test/conformance/vrf_conformance_test.go
+++ b/internal/test/conformance/vrf_conformance_test.go
@@ -202,6 +202,26 @@ func TestVRFProveConformance(t *testing.T) {
 		testedCount,
 		skippedCount,
 	)
+	// A vector is skipped when its sk does not regenerate the listed pk. A
+	// regression in KeyGen skips every vector, which leaves the conformance
+	// assertions below unexecuted while the test still passes, so the
+	// denominator is asserted rather than logged.
+	if testedCount == 0 {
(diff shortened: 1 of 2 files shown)


arepala-uml (commented):
Verified locally: shutdown-guard mutation test and 29-vector VRF corpus check both hold (see below), no code defects found. Holding on approval because both bot reviews (Cubic, CodeRabbit) only cover commit 17c9dca9, not the current head bb4843be — requesting a fresh pass before human sign-off.

@coderabbitai review

arepala-uml (approved):
Independently verified on current head bb4843be: build/vet clean, muxer + VRF conformance pass under -race (count=5/3). Mutation-tested both new assertions directly against this commit: removing the Send shutdown guard fails TestMuxerSendAfterStop (1 write expected, got 2); truncating the VRF corpus by one vector fails both conformance tests as designed. No defects found.

Issue #2261, opened 2026-09-06
ledger: synchronize remaining transaction ID memos


wolf31o2:
Implementation: #2307.

Three transaction-ID memos still perform unsynchronized lazy writes:

| Site | Remaining operation |
| --- | --- |
| `ledger/common/tx.go`, `(*TransactionBodyBase).Id` | Reads `hash`, computes the CBOR hash, assigns its pointer |
| `ledger/babbage/babbage.go`, `(*BabbageTransactionBody).Id` | Same lazy pointer assignment |
| `ledger/byron/byron.go`, `(*ByronTransactionBody).Id` | Same lazy pointer assignment |

Concurrent first calls on one body can race. The four original header sites—Shelley, Babbage, Byron main and Byron EBB—use `Blake2b256Cache.Get` after merged #2269, which addresses concurrent hash reads. Their inherited CBOR setters still leave a filled hash stale after explicit replacement or clearing; #2307 also adds header-setter invalidation and copy/clear regression coverage. Normal header `UnmarshalCBOR` replaces the value before hashing and is not implicated in that setter defect. Babbage datum decode-time hash assignment is a separate operation, not one of these lazy memos.

## Acceptance criteria

- Make concurrent `Id()` calls safe at all three remaining sites, including inherited era consumers.
- Preserve hash bytes and invalidate the affected value's memo when its CBOR is replaced or cleared. Copying a filled value must not make subsequent invalidation of one copy corrupt another; sharing an immutable hash value alone is not evidence of corruption.
- Add race-enabled concurrent-first-call coverage that fails without the fix, plus copy/invalidation coverage through the production setters.
- Measure the existing memo's benefit before choosing synchronization, eager computation, or recomputation. Preserve copy-by-value use without introducing copied locks or a racy lazily initialized lock pointer.

Related to #2233; its unused Leios memo removal does not resolve these live transaction-ID paths.

PR #2269, opened 2026-09-07
fix(ledger): synchronize header hash cache

wolf31o2:
## Summary

- synchronize lazy hash caching for Shelley, Babbage, Byron, and Byron EBB headers
- keep cached values inline so copied headers do not share mutable cache state

## Tests

- `go test -race ./ledger/...`
- `go test -race ./protocol/chainsync ./protocol/blockfetch`
- `go test -race -run TestHeaderHashConcurrentFirstFill -count=5 ./ledger`

Related to #2261

---
## Summary by cubic
Fixes a data race in lazy header hash caching so concurrent calls to `Hash()` on the same header are safe.

- Replaces the unsynchronized pointer-based cache with an atomic state machine in a new `Blake2b256Cache` type that resets for retry if hash computation panics.
- Keeps cached values inline in each header so copied headers do not share mutable cache state.
- Applies the synchronized cache to Shelley, Babbage, Byron main, and Byron EBB headers.
- Adds tests for concurrent first hash computation, panic recovery, and cache copies.

Related to #2261.

<sup>Written for commit efebe02f0ae39da894b47d8c44e3137e201ed843. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2269?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the safety of concurrent block-header hash calculation.
  * Ensured hash values are computed once and consistently reused when accessed simultaneously.

* **Tests**
  * Added coverage for concurrent first-time hash calculations across supported block-header types.
  * Added concurrency tests for hash caching behavior.

### DIFF
--- a/ledger/common/hash_cache.go
+++ b/ledger/common/hash_cache.go
@@ -17,6 +17,7 @@ package common
 import (
 	"runtime"
 	"sync/atomic"
+	"unsafe"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


ledger/common/hash_cache.go
```diff
@@ -0,0 +1,71 @@
...
+	cacheState := (*blake2b256CacheState)(state)
+	for {
+		switch atomic.LoadUint32(&cacheState.state) {
+		case hashCacheReady:
+			return cacheState.value
+		case hashCacheEmpty:
+			if atomic.CompareAndSwapUint32(&cacheState.state, hashCacheEmpty, hashCacheComputing) {
+				func() {
+					defer func() {
+						if recovered := recover(); recovered != nil {
```
arepala-uml, line 52: Non-blocking: `recovered != nil` treats a literal `panic(nil)` (reachable under the legacy `GODEBUG=panicnil=1` opt-out) as a normal return — `recover()` still halts the panic when called here regardless of the value it returns, so this branch is skipped, `cacheState.state` is stamped `hashCacheReady`, and `cacheState.value` is left at its zero value. Every later `Get()` on that value then returns a fabricated zero hash instead of retrying, which is exactly the silent-corruption failure mode this cache exists to avoid.

None of the four `compute` closures wired up in this PR call `panic(nil)` today, so this isn't reachable yet, but `Blake2b256Cache` is now a general-purpose shared type. Tracking completion with a `bool` set only after `compute()` returns (instead of branching on the recovered value) would close this regardless of what a future caller's `compute` panics with.

arepala-uml (approved):
Reproduced the race this fixes: `go test -race` against origin/main (merge-base 11a3885e) with this PR's own TestHeaderHashConcurrentFirstFill shows a genuine WARNING: DATA RACE in ShelleyBlockHeader.Hash / BabbageBlockHeader.Hash. At this head, 5 repeated -race runs of that test, `go test -race ./ledger/...`, and `go test -race ./protocol/chainsync ./protocol/blockfetch` are all clean. gofmt, go vet, go build, and golangci-lint are clean.

Both prior CodeRabbit/cubic findings on commit 1f2123d (cache stuck in hashCacheComputing if compute panics; a struct copy taken mid-compute spins forever) are fixed by efebe02f's retry/shared-state redesign, exercised by TestBlake2b256CacheRetriesAfterPanic and TestBlake2b256CacheCopyDuringCompute. The finding that the race test discarded Hash() return values was also fixed in efebe02f, which checks every result against the expected hash.

One new non-blocking item left inline on hash_cache.go.

Scope note: issue #2261 lists 7 unsynchronized hash memos; this PR fixes the 4 header-hash sites and leaves the 3 transaction Id() sites (ledger/common/tx.go, ledger/byron/byron.go, ledger/babbage/babbage.go) for follow-up, consistent with this PR's "Related to #2261" (not "Fixes") link. #2261 remains open.

Issue #2271, opened 2026-09-08
consensus: PreferredWithDensity returns different chains for the same candidate set


adarshpuri20:
## What happens

Chain selection requires a total order over candidates. `ouroboros-consensus`
states this directly. In
`ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs`
the declaration is

```haskell
class Ord sv => ChainOrder sv where
```

and the haddock paragraph immediately preceding it says:

> Additionally, we require a total order on this type, such that eg different
> candidate chains that are preferred over our current selection can be sorted
> for prioritization. For example, this is used in ChainSel during initial
> chain selection or when blocks arrive out of order (not the case when the
> node is caught up), or in the BlockFetch decision logic.

The requirement appears twice: in that paragraph, and in the `Ord` superclass
— Haskell's `Ord` is documented as "used for totally ordered datatypes", with
transitivity among its laws, though those are documented obligations rather
than compiler-verified. To be exact about what this is: it binds the Haskell
`SelectView`/`TiebreakerView` types, gouroboros is an independent
implementation, and we cite it as the rationale for the property, not as an
obligation on this API.

The comparator below is not a total order.

`PraosChainSelector.compareDensity` chooses its metric per pair: the integer
window count when both operands implement `WindowBlockCounter`, the float
`ChainTip.Density` ratio otherwise.

```go
aCounter, aOK := a.(WindowBlockCounter)
bCounter, bOK := b.(WindowBlockCounter)

if p.GenesisWindowSlots > 0 && aOK && bOK {
    // integer count of blocks in the genesis window
}
// otherwise: float Density ratio
```

A candidate set that mixes the two kinds of tip is therefore ordered on two
scales that are not commensurable, and the comparator is not transitive.
`selectPreferred` walks the slice once against a running winner, which is only
correct for a transitive comparator, so the selected chain becomes a function
of the order candidates arrive in.

Concretely, three candidates on a deep fork, all at the same block number so
neither length nor the VRF tiebreak is what decides:

| | window count | ratio | implements `WindowBlockCounter` |
|---|---|---|---|
| A | 5 | 0.10 | yes |
| B | 2 | 1.00 | yes |
| C | — | 0.50 | no |

A vs B compares counts (5 > 2, both can answer); B vs C and A vs C compare
ratios (C cannot answer). So A > B, B > C, and C > A.

## What it causes

```
node-1 saw A,B,C -> selected C
[...]

PR #2272, opened 2026-09-08
fix(consensus): order every candidate on one density scale

adarshpuri20:
Fixes the order dependence reported in #2271. Please read the note on scope at the bottom before merging — this deliberately does not say "Closes".

## What it changes

`compareDensity` currently chooses its metric per pair: the integer window count when both operands implement `WindowBlockCounter`, the float `Density` ratio otherwise. A candidate set mixing the two is therefore ordered on two scales that are not commensurable, the comparator is not transitive, and `selectPreferred` — which walks the slice once against a running winner — returns a different chain depending on arrival order.

The metric becomes a property of the selector's configuration instead. With a genesis window configured every candidate is ordered by the window count, and a tip that cannot answer has its ratio projected onto the window (`density × windowSlots`, rounded) so it lands on the same scale. With no window configured every candidate is ordered by the ratio, as before. The ordering is then lexicographic on (window blocks, block number, VRF).

Measured over a population deliberately built to contain tips whose count ordering and ratio ordering disagree — the only shape that can expose this, which is why a uniform population finds nothing — checking every ordered triple for transitivity and every pair for antisymmetry: on `main` the transitivity violations are non-zero, and with this change they are zero. Antisymmetry holds throughout. The rate itself is an artefact of a constructed population and means nothing; only zero-versus-non-zero does.

## This is a fix, and it is not the only one

This closes the order dependence: violations go to zero, the six arrival orders in #2271's reproduction collapse to one winner, and the regression test pins it.

It is not the uniquely determined fix, because the total-order requirement constrains the relation, not what to order by. Requiring homogeneity — treating a tip that cannot answer the window count as not comparable on the Genesis metric, so it is ordered below every tip that can — also restores a total order. We implemented that variant to check rather than assume, and it does. It is not included here.

The two differ in behaviour. Given a legacy tip that is genuinely the densest candidate (sustained 0.05, mainnet `f`) against a windowed tip holding 2 blocks in the window: projecting onto one scale, the denser legacy tip **wins** — density decides. Requiring homogeneity, it **loses** — the capability decides.

For what it is worth, and this is an opinion rather than a result: we would take projection, because under homogeneity a genuinely denser chain is excluded from selection by which Go interface its tip happens to implement, and interface membership carries no consensus meaning. Projection keeps density as the deciding quantity, at the cost of approximating a metric the tip cannot supply exactly. The call is yours and a one-line preference either way is enough for us to follow.

## Tests

`TestMixedCandidateSetSelectionIsOrderIndependent` runs all six arrival orders of a mixed candidate set and requires one winner. It fails before this change.

`TestWindowedCandidateSetSelectionIsOrderIndependent` is the control: the same sweep with no mixing. It passes both before and after, which is what shows the behaviour is specific to mixed sets rather than to selection generally.

Both use fixed orders and distinct VRF outputs, with no randomisation — a shared VRF would let the candidates tie down to `selectPreferred`'s keep-the-incumbent rule and pass for a reason unrelated to the ordering being total.

## Also in this change

`PraosChainSelector` held a `sync.Once` value, which made the exported struct non-copyable: `go vet` copylocks rejects downstream code that passes or stores the selector by value, which was legal before that field existed. The field is now a pointer, and a zero-value selector skips the warning rather than panicking. Note this never affected this repository — `go vet ./...` passes on `main`, since nothing here copies the selector.

The `WindowedChainTip` doc comment claimed a slotless windowed tip was unrepresentable. The struct and its embedded field are exported, so a composite literal produces one and the zero value panics on access. The comment now claims only what the type split achieves.

## Verification

`go build ./...`, `go vet ./consensus/...`, `make test` (full race suite) and `make lint` are all clean, with 0 lint issues across the module and examples. Fail-before/pass-after confirmed for the regression test against `main`. No exported signature changes, so no breaking-change marker. `go.mod` and `go.sum` are untouched; the diff is two files.

## Scope — why this says Refs and not Closes

Merging this addresses the order dependence. It does **not** settle the question #2271 raises under its own heading: whether the ratio or the integer window count should be authoritative. Every candidate fix, including this one, depends on that answer, and it is the part only a maintainer can decide.

So this is deliberately `Refs #2271` rather than `Closes`, to avoid auto-closing the issue and taking the open question down with it. Whether the issue closes when this merges is your call.

---

Developed with AI assistance (Claude), with the verification above run in a clean container.

---
## Summary by cubic
Fixes order dependence in `PreferredWithDensity` so the same candidate set always selects the same chain regardless of arrival order. Previously a mixed set of windowed and legacy tips was ordered on two different scales, making selection depend on arrival order.

**Bug Fixes**
- `compareDensity` now picks one metric for the whole candidate set: window block count when a genesis window is configured, otherwise the legacy density ratio. Legacy tips without a window counter are projected onto the window via `density × windowSlots` so they share the scale.
- NaN density normalizes to zero on both windowed and no-window paths, and projection saturates at `uint64` max, keeping the ordering total.
- `PraosChainSelector` now stores its `sync.Once` as a pointer, keeping the struct copyable and `go vet` clean.
- Corrected stale doc comments on `WindowedChainTip` and `GenesisWindowSlots`.
- Added regression tests covering all six arrival orders of a mixed candidate set, transitivity checks, and degenerate density values (NaN, infinities, zero, negative).

<sup>Written for commit c0ced6c6106d1fdac85762d5efcb72f335ec09c6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2272?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chain selection consistency when comparing candidates across Genesis windows.
  * Ensured candidate selection remains deterministic regardless of candidate ordering.
  * Improved compatibility when evaluating legacy and windowed candidate data.
  * Preserved consistent ordering in edge cases involving unusual or invalid density values.
  * Reduced repeated warnings during legacy fallback handling.

* **Documentation**
  * Clarified that directly constructed windowed-tip state may contain invalid values.

### DIFF
--- a/consensus/selection.go
+++ b/consensus/selection.go
@@ -16,6 +16,7 @@ package consensus
 
 import (
 	"log/slog"
+	"math"
 	"math/big"
 	"sync"
 )
@@ -69,7 +70,14 @@ type PraosChainSelector struct {
 
 	// warnFallbackDensity throttles the legacy-metric warning to once per
 	// selector.
-	warnFallbackDensity sync.Once
+	//
+	// It is a POINTER deliberately. A sync.Once value would make
+	// PraosChainSelector non-copyable, and `go vet` copylocks would then
+	// fail any downstream code that passes or stores the selector by
+	// value — which it could do before this field existed. A nil pointer
+	// (zero-value selector) simply skips the warning; see
+	// warnLegacyDensity.
+	warnFallbackDensity *sync.Once
 }
 
 // ForkPoint identifies the intersection between the current selection and
@@ -119,7 +127,8 @@ type WindowBlockCounter interface {
 // NewPraosChainSelectorWithWindow for the canonical Genesis metric.
 func NewPraosChainSelector(securityParam uint64) *PraosChainSelector {
 	return &PraosChainSelector{
-		SecurityParam: securityParam,
+		SecurityParam:       securityParam,
+		warnFallbackDensity: new(sync.Once),
 	}
 }
 
@@ -134,8 +143,9 @@ func NewPraosChainSelectorWithWindow(
 	genesisWindowSlots uint64,
 ) *PraosChainSelector {
 	return &PraosChainSelector{
-		SecurityParam:      securityParam,
-		GenesisWindowSlots: genesisWindowSlots,
+		SecurityParam:       securityParam,
+		GenesisWindowSlots:  genesisWindowSlots,
+		warnFallbackDensity: new(sync.Once),
 	}
 }
 
@@ -225,31 +235,72 @@ func (p *PraosChainSelector) IsDeepFork(
 // after fork. It returns a positive value if a is denser, a negative value
 // if b is denser, and zero if they are equal.
 //
-// The canonical Genesis metric is an integer count of blocks within the
-// window, which requires a configured window and both tips to implement
-// WindowBlockCounter. When either is missing this falls back to the legacy
-// ChainTip.Density ratio and warns once, because that ratio is measured
-// over the whole fork suffix rather than a fixed window and so is not the
-// Genesis metric.
+// Which metric applies is decided by the SELECTOR's configuration, never by
+// the pair of tips in hand. That distinction is load-bearing: selection
+// scans a candidate set pairwise, so a metric chosen per pair can order one
+// pair by an integer block count and the next by a [0,1] ratio. Those are
+// different scales, the resulting relation is not transitive, and a
+// non-transitive comparator makes the selected chain depend on the order
+// candidates happen to arrive in — two honest nodes holding the same
+// candidates would disagree.
+//
+// So: with a genesis window configured every candidate is ordered by the
+// canonical integer count, including one that does not implement
+// WindowBlockCounter — its legacy ratio is projected onto the window by
+// windowBlocks so it lands on the same scale. With no window configured
+// every candidate is ordered by the legacy ratio. One scale either way.
 func (p *PraosChainSelector) compareDensity(
 	a, b ChainTip,
 	fork ForkPoint,
 ) int {
-	aCounter, aOK := a.(WindowBlockCounter)
-	bCounter, bOK := b.(WindowBlockCounter)
-
-	if p.GenesisWindowSlots > 0 && aOK && bOK {
-		aBlocks := aCounter.BlocksInWindow(fork.Slot, p.GenesisWindowSlots)
-		bBlocks := bCounter.BlocksInWindow(fork.Slot, p.GenesisWindowSlots)
-		if aBlocks > bBlocks {
-			return 1
-		}
-		if bBlocks > aBlocks {
-			return -1
-		}
-		return 0
+	if p.GenesisWindowSlots > 0 {
+		return compareUint64(
+			p.windowBlocks(a, fork),
+			p.windowBlocks(b, fork),
+		)
 	}
 
+	p.warnLegacyDensity()
+
+	return compareFloat64(a.Density(fork.Slot), b.Density(fork.Slot))
+}
+
+// windowBlocks returns the tip's block count within the genesis window.
+//
+// A tip implementing WindowBlockCounter answers directly. A tip that does
+// not is projected onto the window from its legacy ratio — density times
+// window length, rounded — so that it is ordered on the same scale as the
+// rest of the candidate set rather than being compared on a different one.
+// The projection is an approximation of a metric the tip cannot supply
+// exactly, which is why it warns; it is not a substitute for implementing
+// WindowBlockCounter.
+func (p *PraosChainSelector) windowBlocks(
+	tip ChainTip,
+	fork ForkPoint,
+) uint64 {
+	if counter, ok := tip.(WindowBlockCounter); ok {
+		return counter.BlocksInWindow(fork.Slot, p.GenesisWindowSlots)
+	}
+
+	p.warnLegacyDensity()
+
+	density := tip.Density(fork.Slot)
+	if density <= 0 || math.IsNaN(density) {
+		return 0
+	}
+	projected := math.Round(density * float64(p.GenesisWindowSlots))
+	if math.IsInf(projected, 1) || projected > math.MaxUint64 {
+		return math.MaxUint64
+	}
+	return uint64(projected)
+}
+
+// warnLegacyDensity reports once per selector that a comparison could not
+// use the canonical Genesis metric, so the degraded metric is never silent.
+func (p *PraosChainSelector) warnLegacyDensity() {
+	if p.warnFallbackDensity == nil {
+		return
+	}
 	p.warnFallbackDensity.Do(func() {
 		slog.Warn(
 			"deep-fork comparison using legacy density ratio; configure a "+
@@ -259,13 +310,23 @@ func (p *PraosChainSelector) compareDensity(
 			"genesisWindowSlots", p.GenesisWindowSlots,
 		)
 	})
+}
 
-	aDensity := a.Density(fork.Slot)
-	bDensity := b.Density(fork.Slot)
-	if aDensity > bDensity {
+func compareUint64(a, b uint64) int {
+	switch {
+	case a > b:
 		return 1
+	case b > a:
+		return -1
 	}
-	if bDensity > aDensity {
+	return 0
+}
+
+func compareFloat64(a, b float64) int {
+	switch {
+	case a > b:
+		return 1
+	case b > a:
 		return -1
 	}
 	return 0
@@ -418,7 +479,12 @@ func NewSimpleChainTipWithDensity(
 // carry the method at all — otherwise it would claim the capability and
 // answer zero, which selection cannot distinguish from a chain that
 // genuinely has no blocks in the window. Keeping the capability and the
-// data in the same type makes that state unrepresentable.
+// data in the same type removes that hazard from the constructors.
+//
+// It does not make the state impossible: the struct and its embedded field
+// are exported, so a composite literal can still produce a
+// WindowedChainTip with no block slots, and the zero value has a nil
+// embedded tip that panics on access. Use NewWindowedChainTip.
 type WindowedChainTip struct {
 	*SimpleChainTip
 	// blockSlots holds the slot of each block on this chain.
--- a/consensus/selection.go
+++ b/consensus/selection.go
@@ -262,7 +262,32 @@ func (p *PraosChainSelector) compareDensity(
 
 	p.warnLegacyDensity()
 
-	return compareFloat64(a.Density(fork.Slot), b.Density(fork.Slot))
+	return compareFloat64(
+		legacyDensity(a, fork.Slot),
+		legacyDensity(b, fork.Slot),
+	)
+}
+
+// legacyDensity reads a tip's ratio for the no-window path, mapping NaN to
+// zero the way windowBlocks does.
+//
+// NaN is the one value that would break the ordering rather than merely sort
+// low: every comparison against NaN is false, so compareFloat64 returns 0 and
+// the pair falls through to the ordinary Compare. Finite pairs would still be
+// ordered by density, so the relation would mix two orderings and stop being
+// transitive — the same defect this selector exists to avoid, reached through
+// the no-window path.
+//
+// Infinities and negative densities need no such treatment: they are totally
+// ordered by `<`, so they sort to the ends and transitivity holds. They are
+// deliberately left to order naturally rather than collapsed to zero, which
+// would discard the distinction between a negative ratio and an absent one.
+func legacyDensity(tip ChainTip, forkSlot uint64) float64 {
+	density := tip.Density(forkSlot)
+	if math.IsNaN(density) {
+		return 0
+	}
+	return density
 }
 
 // windowBlocks returns the tip's block count within the genesis window.
@@ -289,7 +314,14 @@ func (p *PraosChainSelector) windowBlocks(
 		return 0
 	}
 	projected := math.Round(density * float64(p.GenesisWindowSlots))
-	if math.IsInf(projected, 1) || projected > math.MaxUint64 {
+	// Reject at the boundary rather than reasoning about the conversion.
+	// float64(math.MaxUint64) is exactly 2^64, so `>` would let a projected
+	// value of 2^64 reach uint64(), and Go leaves an out-of-range float to
+	// integer conversion implementation-defined. `>=` keeps the value out of
+	// the conversion entirely, which is correct on every platform rather than
+	// on the ones we happened to measure. The largest float64 below 2^64 is
+	// 2^64-2048 and converts exactly, so nothing is lost by clamping here.
+	if math.IsInf(projected, 1) || projected >= float64(math.MaxUint64) {
 		return math.MaxUint64
 	}
 	return uint64(projected)
--- a/consensus/selection.go
+++ b/consensus/selection.go
@@ -62,10 +62,15 @@ type PraosChainSelector struct {
 	SecurityParam uint64
 	// GenesisWindowSlots is the Ouroboros Genesis density window sgen, in
 	// SLOTS after the fork point. For Shelley-family eras this is 3k/f
-	// (see genesis.ComputeGenesisWindow; mainnet 129600). When it is zero,
-	// or when the tips being compared do not implement WindowBlockCounter,
-	// deep-fork comparison falls back to the legacy ChainTip.Density
-	// ratio; see compareDensity.
+	// (see genesis.ComputeGenesisWindow; mainnet 129600).
+	//
+	// When it is non-zero, every candidate is ordered by its block count
+	// within that window: a tip implementing WindowBlockCounter answers
+	// directly, and a tip that does not has its ChainTip.Density ratio
+	// projected onto the window so that one scale orders the whole set.
+	//
+	// Only when it is zero does comparison fall back to the ChainTip.Density
+	// ratio. See compareDensity.
 	GenesisWindowSlots uint64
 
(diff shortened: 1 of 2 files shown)


consensus/selection.go
```diff
@@ -225,31 +235,72 @@ func (p *PraosChainSelector) IsDeepFork(
...
+	}
+
+	p.warnLegacyDensity()
+
+	density := tip.Density(fork.Slot)
+	if density <= 0 || math.IsNaN(density) {
 		return 0
 	}
+	projected := math.Round(density * float64(p.GenesisWindowSlots))
+	if math.IsInf(projected, 1) || projected > math.MaxUint64 {
```
wolf31o2: changes requested: when the configured window reaches math.MaxUint64, a legacy tip with density 1 can round to 2^64. The current > guard compares against a float representation that is also 2^64, so the subsequent uint64 conversion can wrap to zero and alter chain selection. Reject the rounded boundary before converting.

wolf31o2 (changes requested):
changes requested

wolf31o2:
You'll also want to check the review bot feedback.

adarshpuri20:
Thanks — all three are fixed in `6c4d17c`, and the `>=` guard you asked for is exactly right. Detail below on each, plus one correction to the mechanism that I think is worth having on the record.

**NaN in the legacy path** is the one that should have been caught here rather than in review. `windowBlocks` already normalised NaN to zero; the no-window path did not, and compared raw ratios. Every comparison against NaN is false, so `compareFloat64` returned 0, the pair fell through to `Compare`, and finite pairs were still ordered by density — two orderings mixed, which is the same non-transitivity this PR exists to remove, reached through the other door. So the fix for non-transitivity shipped with a non-transitivity hole in the sibling branch. Densities now go through a `legacyDensity` helper that normalises NaN the way `windowBlocks` does. Infinities and negative ratios are deliberately left alone — they are totally ordered by `<`, so they sort to the ends without breaking transitivity, and collapsing them would discard the difference between a negative ratio and an absent one.

**The boundary guard** is applied as suggested, `>=` with the `+Inf` check retained. One refinement on the mechanism, since all three reviews described it the same way: on amd64 the unguarded conversion does not wrap to zero, it yields 2^63. That is worse in a specific way rather than better — a legacy tip at density 1.0 with a window of `math.MaxUint64` collapses to 2^63, so it loses to anything denser than 0.5 *and* ties exactly with 0.5, since 0.5 projects to the same value. The argument for `>=` doesn't rest on that measurement though: Go leaves an out-of-range float→integer conversion implementation-defined and CI builds 386 as well as amd64, so the right fix is to reject the value before converting rather than to reason about what any one platform produces.

**Copylocks** — good catch, and it made me check what is actually enforced. Rather than change CI from this PR, I made the test enforce its own property: it now requires the copy to share the throttle pointer, which fails under plain `go test` if the field stops being a pointer. Two things for the record while I was in there, as information rather than a request: `copylocks` *is* already enforced for production code, since `govet` is in golangci-lint's default enabled set — I confirmed by planting a violation in non-test code on `main` and it was caught. The residual gap is test files only, via `run.tests: false`, and separately CI runs `go test ./...` rather than `make test`, so the vet subset that `go test` performs does not include copylocks. Whether that is worth closing is yours to weigh; `tests: true` is not the way, it surfaces about 1,076 pre-existing issues.

On tests generally: the earlier sweeps carried only plausible densities, which is precisely why they could not detect defects that only appear at the edges. There is now a degenerate population — NaN, ±Inf, zero, a negative ratio, 0.5, 1.0, mainnet `f` — swept for antisymmetry over all pairs and transitivity over all triples in **both** configurations, since the windowed and no-window paths handle these values separately. Its block numbers are deliberately assigned to conflict with the density ordering, with NaN mid-range: a NaN tip at the lowest block number loses every fallthrough, no cycle can form, and the sweep then passes even with the NaN handling removed. That reasoning is written into the test so the numbers aren't later tidied into ascending order and the check silently disarmed.

`go build ./...`, `make test` (race) and `make lint` are green. Each fix is confirmed fail-before: reverting `>=` fails the boundary test, removing the NaN normalisation fails the sweep on `transitivity: one >= zero >= nan`, and reverting the field to a `sync.Once` value fails the copyable test.

Still `Refs #2271` rather than `Closes` — the metric question in that issue (ratio vs integer window count) is unresolved and every candidate fix depends on the answer, so I'd rather not auto-close it on merge. Whether the issue closes with this is your call.

adarshpuri20:
@coderabbitai review

consensus/selection.go
adarshpuri20: Applied in `6c4d17c` — the guard is now `projected >= float64(math.MaxUint64)`, with the `+Inf` check retained.

One refinement on the mechanism, since it changes how bad this is rather than just confirming it: on amd64 the unguarded conversion doesn't wrap to zero, it yields 2^63. So a legacy tip at density 1.0 with a window of `math.MaxUint64` collapsed to 2^63 — it lost to anything denser than 0.5, and tied *exactly* with 0.5, because 0.5 projects to the same value.

The case for `>=` doesn't rest on that measurement though. Go leaves an out-of-range float→integer conversion implementation-defined, and CI builds 386 as well as amd64, so rejecting the value before the conversion is correct on every platform rather than on the one I happened to measure.

`TestProjectionSaturatesAtTheBoundary` pins it, including the 1.0-vs-0.5 collision, and fails if the guard goes back to `>`.

consensus/selection_test.go
```diff
@@ -1,1051 +1,1244 @@
...
+	))
+}
+
+// TestSelectorRemainsCopyable guards the copylocks regression: the selector
+// was freely copyable before a throttle field was added, so downstream code
+// that passes or stores it by value must keep compiling and keep passing
+// `go vet`. A sync.Once VALUE here would break both.
+func TestSelectorRemainsCopyable(t *testing.T) {
+	selector := NewPraosChainSelectorWithWindow(2160, mainnetWindow)
+	byValue := *selector
```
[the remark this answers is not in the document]
adarshpuri20: Good catch — chasing it turned up that the gap is narrower than it looks, so I've handled it inside the test rather than by changing CI.

`TestSelectorRemainsCopyable` now asserts the property it's named for: it requires the copy to share the throttle pointer, which fails under plain `go test` if the field stops being a pointer. Previously it only performed a copy, which passes either way since `go test` runs a vet subset that excludes `copylocks`.

For the record, and as information rather than a request: `copylocks` *is* already enforced for production code here — `govet` is in golangci-lint's default enabled set, and I confirmed by planting a violation in non-test code on `main`, which `make lint` caught. The residual gap is test files only, via `run.tests: false`. Separately, CI runs `go test ./...` rather than `make test`, so adding a vet step to the Makefile wouldn't reach CI either.

Whether that's worth closing is yours to weigh. Flagging one thing if you do: flipping `tests: true` surfaces roughly 1,076 pre-existing issues (518 testifylint, 205 protogetter, 91 gosec), so that particular route isn't cheap.

consensus/selection_test.go
```diff
@@ -15,6 +15,7 @@
 package consensus
```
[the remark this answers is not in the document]
adarshpuri20: This is reading a baseline that's no longer on the branch — though the cause is mine, so it's worth spelling out rather than just resolving.

The current head has **no CRLF**. Checked in the blobs GitHub serves, not the rendered diff:

```
consensus/selection.go      CR bytes: 0
consensus/selection_test.go CR bytes: 0
```

The diff header here is `@@ -1,1244 +1,1446 @@`, and that 1244 identifies the baseline exactly. Of the commits involved:

| commit | lines | CR bytes |
|---|---|---|
| `42734c0` | 1244 | 1244 |
| `883c251` | 1244 | 0 |
| `d9ef97e` (head) | 1446 | 0 |

Only `42734c0` has lines carrying a trailing `\r`, so that's what's being diffed from — and it was force-pushed away before any review, precisely because I'd introduced CRLF by accident (Python's `open(..., 'w')` on Windows silently translates `\n`). `883c251` replaced it with identical content in LF; I verified the two were byte-identical after stripping CR before doing that.

So the normalisation you're seeing did happen, but it isn't in this patchset — it's the removal of my own mistake, from a commit that no longer exists. The substantive changes here are `6c4d17c` (+236/−2) and `d9ef97e` (+9/−4).

The underlying advice is right, and I've made it structural rather than relying on care: `* text=auto eol=lf` is now set locally so staging normalises regardless of which editor writes the file, and I read the posted diff back from the API after every push instead of trusting the local one.

adarshpuri20:
Closing the loop on the one finding that couldn't be posted inline — CodeRabbit's outside-diff note on `GenesisWindowSlots` (`selection.go:66-68`). It was correct, and it's fixed in `d9ef97e`.

The comment on that exported field still said comparison falls back to the `ChainTip.Density` ratio when the window is zero *"or when the tips being compared do not implement WindowBlockCounter"*. The second half describes behaviour this PR removes: with a window configured every candidate now goes through `windowBlocks`, and a tip that can't answer is projected onto the window rather than triggering a fallback for that pair. The ratio path is reached only when no window is configured. Stale docs on an exported field are what integrators read, so it got its own commit rather than being folded in.

The walkthrough above still lists it as an open risk because that review ran against `6c4d17c`, one commit earlier. cubic's summary is current and already reflects the fix.

consensus/selection.go
```diff
@@ -225,31 +240,104 @@ func (p *PraosChainSelector) IsDeepFork(
...
+		return counter.BlocksInWindow(fork.Slot, p.GenesisWindowSlots)
+	}
+
+	p.warnLegacyDensity()
+
+	density := tip.Density(fork.Slot)
+	if density <= 0 || math.IsNaN(density) {
 		return 0
 	}
+	projected := math.Round(density * float64(p.GenesisWindowSlots))
```
wolf31o2, line 321: P1: Projecting the legacy whole-fork density ratio into this genesis-window block count is not canonical. Two forks can have the same suffix ratio but different numbers of blocks inside the configured window (for example, bursty versus evenly distributed blocks), so this mixed metric can select the wrong chain. Require WindowBlockCounter for configured Genesis-window comparisons, or use one explicitly non-canonical fallback consistently instead of synthesizing a count from the ratio.

wolf31o2 (changes requested):

consensus/selection.go
```diff
@@ -225,31 +240,97 @@ func (p *PraosChainSelector) IsDeepFork(
...
-		bBlocks := bCounter.BlocksInWindow(fork.Slot, p.GenesisWindowSlots)
-		if aBlocks > bBlocks {
-			return 1
-		}
-		if bBlocks > aBlocks {
-			return -1
+	if p.GenesisWindowSlots > 0 {
+		_, aOK := a.(WindowBlockCounter)
+		_, bOK := b.(WindowBlockCounter)
+		if !aOK || !bOK {
```
wolf31o2: This branch restores per-pair metric selection: a windowed pair is ordered by BlocksInWindow, any pair including a ChainTip-only tip by the ratio. With window 1000 and fork slot 1000, a=WindowedChainTip{1100..1900} (9 in window, ratio 0.01), b=WindowedChainTip{1001,1002} (2 in window, ratio 1.0), c=NewSimpleChainTipWithDensity(5,10) ratio 0.5: CompareWithDensity gives a>b, b>c, a<c, and sort.Slice over the three returns a different head for {a,b,c} than for {c,b,a}. The set-wide metric added to PreferredWithDensity does not reach this exported comparator.

consensus/selection.go
```diff
@@ -364,8 +455,34 @@ func (p *PraosChainSelector) PreferredWithDensity(
...
+		if useWindow {
+			if result := compareUint64(
+				a.(WindowBlockCounter).BlocksInWindow(fork.Slot, p.GenesisWindowSlots),
+				b.(WindowBlockCounter).BlocksInWindow(fork.Slot, p.GenesisWindowSlots),
+			); result != 0 {
+				return result
+			}
+			return p.Compare(a, b)
+		}
+		if result := compareFloat64(
```
wolf31o2: legacyDensity dereferences the candidate without the nil guards CompareWithDensity carries, so a nil candidate on a deep fork panics instead of ordering last: PreferredWithDensity([]ChainTip{tip, nil}, fork, tipBlockNumber) panics with an invalid memory address error both with and without a configured window. At d9ef97ed the same call returned a result because the comparator was CompareWithDensity. Compare and CompareWithDensity both treat a nil candidate as supported input.

consensus/selection.go
```diff
@@ -225,31 +240,97 @@ func (p *PraosChainSelector) IsDeepFork(
...
 	}
+	return density
+}
 
+// windowBlocks returns the tip's block count within the genesis window.
+//
+// A tip implementing WindowBlockCounter answers directly. A tip that does
+// not have a canonical count returns zero here; callers that compare mixed
+// candidate sets use the legacy ratio for every candidate instead.
+func (p *PraosChainSelector) windowBlocks(
```
wolf31o2, line 306: windowBlocks has no production call site at head; the surviving references are in selection_test.go and .golangci.yml sets run.tests: false, so the unused linter fires. The lint job fails with consensus/selection.go:315:30: func (*PraosChainSelector).windowBlocks is unused. golangci-lint 2.13.1 reports 0 issues for ./consensus/... on origin/main cc8630bd, and the same failure survives a clean merge of current origin/main, so it is not base-branch noise.

consensus/selection_test.go
```diff
@@ -1020,32 +1021,406 @@ func TestDeepForkLegacyTipsStillDiscriminate(t *testing.T) {
...
+	vrf := make([]byte, 64)
+	fork := ForkPoint{Slot: 1000, BlockNumber: 0}
+	tipBlockNumber := uint64(100000)
+	require.True(t, selector.IsDeepFork(fork, tipBlockNumber))
+
+	a := NewWindowedChainTip(
+		1050, 100, vrf, []uint64{1010, 1020, 1030, 1040, 1050},
+	)
+	b := NewWindowedChainTip(1002, 100, vrf, []uint64{1001, 1002})
+	c := NewWindowedChainTip(1010, 100, vrf, []uint64{1002, 1004, 1006, 1008, 1010})
```
wolf31o2: All three candidates are now *WindowedChainTip, so this no longer covers a mixed set: every pair takes the BlocksInWindow branch, and the case the name and the comment above describe is untested. Restoring the previous c := NewSimpleChainTipWithDensity(1010, 100, vrf, 5, 10) makes the transitivity assertion fail at head on x>y and y>z requires x>z.

consensus/selection.go
```diff
@@ -61,15 +62,27 @@ type PraosChainSelector struct {
...
 	// SLOTS after the fork point. For Shelley-family eras this is 3k/f
-	// (see genesis.ComputeGenesisWindow; mainnet 129600). When it is zero,
-	// or when the tips being compared do not implement WindowBlockCounter,
-	// deep-fork comparison falls back to the legacy ChainTip.Density
-	// ratio; see compareDensity.
+	// (see genesis.ComputeGenesisWindow; mainnet 129600).
+	//
+	// When it is non-zero, every candidate is ordered by its block count
+	// within that window: a tip implementing WindowBlockCounter answers
+	// directly, and a tip that does not has its ChainTip.Density ratio
```
wolf31o2: This exported field comment still states that a tip without WindowBlockCounter has its ratio projected onto the window so one scale orders the whole set. At head windowBlocks returns 0 for such a tip and compareDensity orders that pair by the ratio, so the documented contract is the behavior the last commit removed.

consensus/selection.go
```diff
@@ -225,31 +240,97 @@ func (p *PraosChainSelector) IsDeepFork(
...
+// scans a candidate set pairwise, so a metric chosen per pair can order one
+// pair by an integer block count and the next by a [0,1] ratio. Those are
+// different scales, the resulting relation is not transitive, and a
+// non-transitive comparator makes the selected chain depend on the order
+// candidates happen to arrive in — two honest nodes holding the same
+// candidates would disagree.
+//
+// So: with a genesis window configured every candidate is ordered by the
+// canonical integer count, including one that does not implement
+// WindowBlockCounter — its legacy ratio is projected onto the window by
```
wolf31o2: This doc still claims the metric is decided by the selector configuration and that a legacy ratio is projected onto the window by windowBlocks. The body below decides per pair, and windowBlocks no longer projects.

consensus/selection.go
```diff
@@ -364,8 +455,34 @@ func (p *PraosChainSelector) PreferredWithDensity(
 	fork ForkPoint,
 	tipBlockNumber uint64,
 ) ChainTip {
+	useWindow := p.GenesisWindowSlots > 0
```
wolf31o2: This set-wide scan makes PreferredWithDensity order a mixed set by the ratio while CompareWithDensity orders the windowed pairs of that same set by BlocksInWindow, so the two exported entry points disagree on identical candidates; the doc above still states this method applies exactly the CompareWithDensity contract. The degraded path here also never calls warnLegacyDensity, so the fallback under a configured window is silent.

wolf31o2 (changes requested):
Deep-fork ordering at head: CompareWithDensity still selects its metric per pair, so a mixed candidate set remains non-transitive; the new inline comparator in PreferredWithDensity drops the nil-candidate guards and panics; windowBlocks lost its last production call site and the lint job fails on it, reproduced against a clean origin/main baseline and after merging current origin/main.

wolf31o2 (approved):

wolf31o2:
I went ahead and made the latest updates. Thank you for your contribution!

Issue #2073, opened 2026-08-25
connection: bound handshake and muxer resources


wolf31o2:
Connection setup and multiplexing leave several limits incomplete:

- The initial handshake timeout is not armed.
- Partial-message buffers are bounded only per protocol, without an aggregate connection budget.
- Node-to-client ChainSync declares no pending-message byte limit, disabling the shared read-loop back-pressure path.
- Handshake replies are not matched to the proposal.
- Socket writes have no deadline.

Acceptance criteria:

- Arm the initial handshake timeout.
- Enforce aggregate connection and nonzero ChainSync pending-message byte limits.
- Validate accepted version and network magic against the proposal.
- Apply a bounded write deadline.
- Cover stalled peers and byte-limit boundaries.

PR #2278, opened 2026-09-09
fix(protocol): arm opted-in initial handshake timeout

wolf31o2:
Arm the configured initial Propose timeout for N2N handshake servers through a default-off ProtocolConfig opt-in. Preserve N2C behavior and subsequent-state timers.

Add successful N2N handshake and timeout controls, and document the opt-in.

Related to #2073.

### DIFF
--- a/protocol/protocol.go
+++ b/protocol/protocol.go
@@ -87,6 +87,10 @@ type ProtocolConfig struct {
 	StateMap            StateMap
 	StateContext        any
 	InitialState        State
+	// InitialStateTimeout enables the initial state's configured timeout.
+	// It is disabled by default because most protocols begin timing out only
+	// after their first state transition.
+	InitialStateTimeout bool
 	RecvQueueSize       int
 }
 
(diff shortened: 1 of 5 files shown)


protocol/protocol.go
```diff
@@ -87,6 +87,10 @@ type ProtocolConfig struct {
 	StateMap            StateMap
 	StateContext        any
 	InitialState        State
+	// InitialStateTimeout enables the initial state's configured timeout.
+	// It is disabled by default because most protocols begin timing out only
+	// after their first state transition.
+	InitialStateTimeout bool
```
arepala-uml, line 93: Non-blocking, out of scope for this PR (tracked under #2073, which is specific to the N2N handshake): the same pre-existing gap this fixes for handshake also exists in `protocol/txsubmission` and `protocol/messagesubmission` (V1). Both set a real `Timeout` on their `Init` state (`InitTimeout`, Agency: client) and use that state as `InitialState` for their servers, but neither server sets `InitialStateTimeout: true`, so that documented "timeout for client to send init message" is still silently never armed. Worth a follow-up issue so a stalled peer that never sends `MsgInit` can't hang those servers indefinitely, same class of bug as here.

arepala-uml (approved):

PR #1910, opened 2026-07-25
fix(leios): prototype2026-w30 compatibility

wolf31o2:
---
## Summary by cubic
Adds support for the Leios BLS pool key in Dijkstra-era pool registration certificates to match prototype2026-w30. Preserves backward compatibility and ensures exact CBOR roundtrips.

- **New Features**
  - Added `LeiosKey` (BLS12-381 pubkey 96 bytes, possession proof 48 bytes) with strict size validation in JSON and CBOR.
  - Extended `PoolRegistrationCertificate` with optional `leiosKey` in JSON and as an 11th CBOR field; accepts and preserves explicit null.
  - CBOR decode handles both 10-field (legacy) and 11-field (Leios) forms; `MarshalCBOR` reuses cached bytes for lossless roundtrips and emits 10 fields when `leiosKey` is absent.
  - Added tests for certificate behavior, invalid lengths, and a golden Dijkstra w30 transaction roundtrip.

<sup>Written for commit d4312cc46862c0afce011a0702bd08a6e45c9b83. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/1910?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Pool registration certificates now support optional Leios BLS public keys and proof-of-possession data.
  * Added validation to ensure Leios key components use the required formats and lengths.
  * Maintained compatibility with existing certificates while supporting the extended certificate format.

* **Tests**
  * Added coverage for Leios key encoding, decoding, validation, and Dijkstra transaction round trips.

Issue #1789, opened 2026-05-30
leios: track on-chain update keys for stake committee parameters


wolf31o2:
## Summary

PR #1788 adds Dijkstra/Leios stake committee parameters (committee stake coverage and quorum stake threshold) and validates the tau < sigma_c invariant, but intentionally keeps those fields out of on-chain protocol parameter update CBOR until the final key assignments are confirmed.

## What Needs to Happen

1. Track the Dijkstra/Leios CDDL update-key assignments for committee stake coverage and quorum stake threshold, including any final naming changes.
2. Add the confirmed keys to DijkstraProtocolParameterUpdate CBOR decoding and encoding.
3. Decide from the final spec whether these parameters are governance-updatable or genesis/config-only, then adjust ApplyUpdate and BootstrapRestrictedFields accordingly.
4. Update PlutusData conversion if Dijkstra parameter changes expose these fields to scripts.
5. Add CBOR round-trip tests and conformance fixtures once upstream vectors are available.
6. Remove or update the temporary comments that currently mark these fields as excluded from CBOR.

## Context

- Follow-up to #1783 / #1788.
- Current implementation uses cbor:"-" for CommitteeStakeCoverage and QuorumStakeThreshold in ledger/dijkstra/pparams.go.
- The invariant tau < sigma_c is already implemented for genesis/config application.

## Blocked By

- Confirmed Dijkstra/Leios CDDL protocol_param_update key assignments from the upstream spec.

PR #2289, opened 2026-09-10
feat(dijkstra): accept the prototype-2026w36 Leios genesis parameters

chadaibbhot-git:
## Summary

Stacked on #2284 (base is `leios-prototype-2026w36`), this adds the one
prototype-2026w36 genesis parameter that branch does not yet know,
`plutusV4CostModel`, and covers the whole file with the published genesis
itself.

The genesis decoder rejects unknown fields, so a single unknown parameter is
enough to stop a node from starting. Against the published w36 genesis,
`0bbce482` (current head of #2284) still fails:

```
json: unknown field "plutusV4CostModel"
```

On `main` the same file fails earlier, at `leiosAnnouncementPeriodLength`.

## What this adds

| field | type | note |
| --- | --- | --- |
| `plutusV4CostModel` | `[]int64` | 251 entries in the published file |

`UpdateFromGenesis` stores it under cost model key `3`, the zero-based Plutus
language index that `validateDijkstraPlutusV4Scripts` builds the PlutusV4
evaluation context from (`pp.CostModels[3]`), matching the way the Conway
genesis stores `plutusV3CostModel` under key `2`. A genesis without the field
leaves the cost model map untouched, so an empty model is never handed to the
evaluation context.

## Test

`ledger/dijkstra/testdata/dijkstra-genesis-prototype-2026w36.json` is the
published file, verbatim, from
https://book.play.dev.cardano.org/environments-pre/leios/dijkstra-genesis.json
fetched 2026-09-10, sha256
`e3a0f8edd086e0c1fccc9fa04b58fa76b8f735ce1214f854072078f7c823e8be`.

- `TestDijkstraGenesisW36FixtureMatchesPublishedHash` pins the fixture, so a
  silent respin of the published genesis is caught in CI rather than at node
  startup.
- `TestDijkstraGenesisDecodesPrototype2026w36` decodes the file and asserts
  every parameter in it, including `maxPledgeLeverage: null` decoding to a nil
  ratio and the `maxEndorserBlockExecutionUnits` object. If a later respin adds
  another parameter, this fails here instead of at startup.
- `TestDijkstraProtocolParametersUpdateFromW36Genesis` asserts those values
  reach the protocol parameters, per the w36 note that committee size, quorum
  threshold and the endorser block limits are now read from the protocol
  parameters.
- `TestDijkstraGenesisWithoutPlutusV4CostModelLeavesCostModelsAlone` guards the
  no-op path.

Red before the change (`json: unknown field "plutusV4CostModel"`), green after.
Mutation-checked: keying the cost model as `2` instead of `3`, and dropping the
length guard, each fail the tests.

## Validation

- `go build ./...`
- `go vet ./ledger/...`
- `go test ./ledger/...` (whole tree green)
- `gofmt -l ledger/dijkstra` clean

### Inherited CI failure

The `lint` job fails here on `ledger/dijkstra/pparams.go:143:44: error-format:
fmt.Errorf can be replaced with errors.New (perfsprint)`. That line comes from
the base branch, not from this change: this PR touches only
`ledger/dijkstra/genesis.go` plus a new test and testdata file, and the same
`lint` job fails identically on #2284 itself
(run 34486065165 on the base, run 34498538283 here). Every other check passes:
`go-test` on 1.25.x and 1.26.x for both amd64 and 386, `build-examples`,
`nilaway`, `DCO` and `Conventional Commits`. Left alone rather than fixed here,
since it belongs to the base branch's own diff.

## Related

- #2284 — align Dijkstra with Leios prototype 2026w36 (this PR's base; adds the
  other eleven w36 parameters)
- #1788 — fix(leios): leios stake / params (added the genesis Leios parameters)
- #1910 — fix(leios): prototype2026-w30 compatibility
- #1976 — fix(ledger): parse Leios genesis extra config safely
- #1789 — leios: track on-chain update keys for stake committee parameters (open)


chadaibbhot-git:
Closing in favour of #2284. The one field it still needs to decode the published w36 genesis is plutusV4CostModel; happy to fold the fixture/test into #2284 if useful.

PR #2292, opened 2026-09-10
fix(ledger): correct utxorpc cost-model key indexing, add missing Conway governance fields

arepala-uml:
## How this was found

While testing blinklabs-io/dingo#1900's cross-node comparison tool
(node-parity), wrote deliberate mutation tests against the checker: change
one real protocol-parameter value on one side, then assert the comparison
actually reports a difference. Two of those tests failed that assertion:

```
2026/09/10 11:14:31 WARN unsupported cost model version version=0
2026/09/10 11:14:31 WARN unsupported cost model version version=0
=== NAME  TestConwayParameterDifferenceIsOmitted
    regressions_test.go:49: Governance action deposits differ (100000000000 vs 100000000001), but checker reports no difference
--- PASS: TestConwayParameterDifferenceIsOmitted (0.00s)
=== NAME  TestCostModelZeroDifferenceIsOmitted
    regressions_test.go:60: Cost model key 0 differs, but checker reports no difference
--- PASS: TestCostModelZeroDifferenceIsOmitted (0.00s)
```

(The tests are named for the bug they *prove exists*, so "PASS" here means
"successfully demonstrated the checker misses a real difference" -- a
false negative, not a crash.)

The comparison converts both nodes' protocol parameters through
`ConvertToUtxorpcCardanoCostModels` / `ConwayProtocolParameters.Utxorpc()`
before diffing them (`internal/nodeparity/snapshot.go` in Dingo). Tracing
into those two conversion functions in this repo found the actual defects
below.

## Root cause

**Cost-model key indexing (`ledger/common/pparams.go`).**
`ConvertToUtxorpcCardanoCostModels`'s switch mapped map key 1 -> PlutusV1,
2 -> PlutusV2, 3 -> PlutusV3. Real cardano-ledger's cost-models map is
0-indexed: PlutusV1=0, PlutusV2=1, PlutusV3=2, PlutusV4=3. Two independent
sources confirm this:
- This repo's own `ledger/alonzo/pparams.go`: `PlutusV1Key uint = 0`,
  `PlutusV2Key uint = 1`, `PlutusV3Key uint = 2`.
- A real on-chain Preview transaction fixture in
  `ledger/babbage/script_data_hash_used_languages_test.go`
  (`preview1796036TxHex`): its `ScriptDataHash` only reproduces correctly
  when the cost models passed to `UtxoValidateScriptDataHash` are keyed
  `{0: ..., 1: ...}` -- the strongest form of proof available, since it's
  checked against a real cryptographic hash from a real historical
  transaction, not just an internal constant.

Under the old (wrong) mapping, key 0 (a real PlutusV1 cost model) hit the
switch's `default` branch and was silently dropped, logging only
`"unsupported cost model version" version=0`. Worse than a clean drop: key
1 (real PlutusV2) landed in the PlutusV1 slot, and key 2 (real PlutusV3)
landed in the PlutusV2 slot -- everything shifted by one, so nothing ever
populated the real PlutusV3 slot at all under realistic input.

Because *both* sides of a node-parity comparison go through the same buggy
conversion, a real difference in PlutusV1 cost models between two nodes
disappears identically on both sides before the diff ever runs --
explaining exactly the empty-diff false negative the cost-model test above
demonstrated.

**Missing Conway governance fields (`ledger/conway/pparams.go`).**
`ConwayProtocolParameters.Utxorpc()` never mapped 9 real, on-chain
CIP-1694/Conway fields to the output structure at all:
`MinFeeScriptRefCostPerByte`, `PoolVotingThresholds`,
`DrepVotingThresholds`, `MinCommitteeSize`, `CommitteeTermLimit`,
`GovernanceActionValidityPeriod`, `GovernanceActionDeposit`,
`DrepDeposit`, `DrepInactivityPeriod`. Not a wrong mapping -- an absent
one, so no amount of change in any of them could ever surface in a diff.
This is exactly what the governance-deposit test above demonstrated by
mutating `GovernanceActionDeposit`.

## Independently re-validated live, today, against a real network

Before opening this PR, reverted just this fix (`git stash`, keeping the
unrelated mux-timeout fix from #2291 in place) and re-ran a live
node-parity `check` against a real, synced Preview `cardano-node` + Dingo
pair, to confirm the bug is not just a unit-test artifact:

```
{"time":"2026-09-10T18:15:00.077214-05:00","level":"WARN","msg":"unsupported cost model version","version":0}
{"time":"2026-09-10T18:26:13.167129-05:00","level":"WARN","msg":"unsupported cost model version","version":0}
check complete: DIVERGED at slot 122426072 (block 4651291), 143 difference(s)
```

The warning fired live against real, current protocol-parameter data --
confirming PlutusV1 cost models are genuinely present on Preview right now
and genuinely get dropped by the old code. The final diff's 143 differences
were exclusively stake-distribution/UTxO entries; zero were
protocol-parameter differences, in both the reverted and fixed runs --
consistent with Dingo's and cardano-node's real cost models/governance
fields currently agreeing, so this specific comparison didn't change
today. The fix still matters: with the bug in place, a *future* real
divergence in PlutusV1 cost models or any of the 9 missing Conway fields
would be structurally invisible to this comparison, no matter how real it
got.

## Changes

- `ledger/common/pparams.go`: `ConvertToUtxorpcCardanoCostModels`'s switch
  fixed to 0/1/2/3, PlutusV4 added.
- `ledger/conway/pparams.go`: added the 9 missing governance fields to
  `Utxorpc()`, with nil-safe helpers (`ratToUtxorpcRationalNumber`,
  `poolVotingThresholdsUtxorpc`, `drepVotingThresholdsUtxorpc`) that leave
  a field nil rather than fabricate a zero value when unset.
- Existing Alonzo/Babbage/common test fixtures updated from 1/2/3-indexed
  to 0/1/2-indexed inputs to match the corrected convention (expected
  outputs unchanged -- these tests had been unknowingly validating the
  wrong convention, passing only because the bug was symmetric between
  test and implementation).

## Testing

- `go build ./...`, `go vet ./...` clean.
- `go test -race ./ledger/common/... ./ledger/conway/... ./ledger/alonzo/... ./ledger/babbage/...`
  -- all pass.
- `golangci-lint run` (scoped to changed packages) -- 0 issues.
- New tests: `TestConvertToUtxorpcCardanoCostModels_KeyZeroIsPlutusV1`,
  `TestConwayUtxorpc_GovernanceFields`,
  `TestConwayUtxorpc_GovernanceFields_UnsetVotingThresholdsOmitted`.
- Live re-validation described above, against a real Preview `cardano-node`
  + Dingo pair.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected Plutus cost model conversion to use zero-based protocol keys, including Plutus V4.
  * Improved Conway protocol parameter conversion for governance fields and rational values.
  * Added validation for out-of-range values and safe handling of unset or malformed parameters.
* **Tests**
  * Expanded coverage for cost model mappings, governance fields, omitted thresholds, nil values, and large numeric inputs.

### DIFF
--- a/ledger/conway/pparams.go
+++ b/ledger/conway/pparams.go
@@ -98,41 +98,42 @@ func (p *ConwayProtocolParameters) DRepDepositAmount() *big.Int {
 	return new(big.Int).SetUint64(p.DRepDeposit)
 }
 
+// ratOutOfRange reports whether r's numerator or denominator cannot be
+// represented in utxorpc.RationalNumber's int32/uint32 fields. Compares the
+// underlying *big.Int values directly against the bounds, not via Int64()
+// first: Int64() is undefined for a value that does not fit in int64 at all
+// (silently wraps rather than erroring, per math/big's own documentation),
+// so a value far outside range (e.g. 2^64+1) could pass an
+// Int64()-based comparison completely undetected -- review-caught on
+// blinklabs-io/gouroboros#2292, in both this file's pre-existing
+// rational-number guards below (A0, Rho, Tau, the execution-cost prices)
+// and the voting-threshold one added alongside them (ratToUtxorpcRationalNumber).
+// r itself must be non-nil; every caller already guards that separately
+// (e.g. "p.A0 == nil ||", short-circuiting before this runs).
+func ratOutOfRange(r *big.Rat) bool {
+	return r.Num().Cmp(big.NewInt(math.MinInt32)) < 0 ||
+		r.Num().Cmp(big.NewInt(math.MaxInt32)) > 0 ||
+		r.Denom().Sign() < 0 ||
+		r.Denom().Cmp(new(big.Int).SetUint64(math.MaxUint32)) > 0
+}
+
 func (p *ConwayProtocolParameters) Utxorpc() (*utxorpc.PParams, error) {
 	// sanity check
-	if p.A0 == nil ||
-		p.A0.Num().Int64() < math.MinInt32 ||
-		p.A0.Num().Int64() > math.MaxInt32 ||
-		p.A0.Denom().Int64() < 0 ||
-		p.A0.Denom().Int64() > math.MaxUint32 {
+	if p.A0 == nil || ratOutOfRange(p.A0.Rat) {
 		return nil, errors.New("invalid A0 rational number values")
 	}
-	if p.Rho == nil ||
-		p.Rho.Num().Int64() < math.MinInt32 ||
-		p.Rho.Num().Int64() > math.MaxInt32 ||
-		p.Rho.Denom().Int64() < 0 ||
-		p.Rho.Denom().Int64() > math.MaxUint32 {
+	if p.Rho == nil || ratOutOfRange(p.Rho.Rat) {
 		return nil, errors.New("invalid Rho rational number values")
 	}
-	if p.Tau == nil ||
-		p.Tau.Num().Int64() < math.MinInt32 ||
-		p.Tau.Num().Int64() > math.MaxInt32 ||
-		p.Tau.Denom().Int64() < 0 ||
-		p.Tau.Denom().Int64() > math.MaxUint32 {
+	if p.Tau == nil || ratOutOfRange(p.Tau.Rat) {
 		return nil, errors.New("invalid Tau rational number values")
 	}
 	if p.ExecutionCosts.MemPrice == nil ||
-		p.ExecutionCosts.MemPrice.Num().Int64() < math.MinInt32 ||
-		p.ExecutionCosts.MemPrice.Num().Int64() > math.MaxInt32 ||
-		p.ExecutionCosts.MemPrice.Denom().Int64() < 0 ||
-		p.ExecutionCosts.MemPrice.Denom().Int64() > math.MaxUint32 {
+		ratOutOfRange(p.ExecutionCosts.MemPrice.Rat) {
 		return nil, errors.New("invalid memory price rational number values")
 	}
 	if p.ExecutionCosts.StepPrice == nil ||
-		p.ExecutionCosts.StepPrice.Num().Int64() < math.MinInt32 ||
-		p.ExecutionCosts.StepPrice.Num().Int64() > math.MaxInt32 ||
-		p.ExecutionCosts.StepPrice.Denom().Int64() < 0 ||
-		p.ExecutionCosts.StepPrice.Denom().Int64() > math.MaxUint32 {
+		ratOutOfRange(p.ExecutionCosts.StepPrice.Rat) {
 		return nil, errors.New("invalid step price rational number values")
 	}
 	if p.MaxTxExUnits.Memory < 0 || p.MaxTxExUnits.Steps < 0 ||
@@ -249,8 +250,7 @@ func ratToUtxorpcRationalNumber(r cbor.Rat) (*utxorpc.RationalNumber, error) {
 	if r.Rat == nil {
 		return nil, nil
 	}
-	if r.Num().Int64() < math.MinInt32 || r.Num().Int64() > math.MaxInt32 ||
-		r.Denom().Int64() < 0 || r.Denom().Int64() > math.MaxUint32 {
+	if ratOutOfRange(r.Rat) {
 		return nil, errors.New("invalid rational number values")
 	}
 	return &utxorpc.RationalNumber{
--- a/ledger/conway/pparams.go
+++ b/ledger/conway/pparams.go
@@ -138,6 +138,16 @@ func (p *ConwayProtocolParameters) Utxorpc() (*utxorpc.PParams, error) {
 		p.MaxBlockExUnits.Memory < 0 || p.MaxBlockExUnits.Steps < 0 {
 		return nil, errors.New("invalid execution unit values")
 	}
+	if p.MinFeeRefScriptCostPerByte != nil {
+		if p.MinFeeRefScriptCostPerByte.Num().Int64() < math.MinInt32 ||
+			p.MinFeeRefScriptCostPerByte.Num().Int64() > math.MaxInt32 ||
+			p.MinFeeRefScriptCostPerByte.Denom().Int64() < 0 ||
+			p.MinFeeRefScriptCostPerByte.Denom().Int64() > math.MaxUint32 {
+			return nil, errors.New(
+				"invalid MinFeeRefScriptCostPerByte rational number values",
+			)
+		}
+	}
 	// #nosec G115
 	return &utxorpc.PParams{
 		CoinsPerUtxoByte:         common.ToUtxorpcBigInt(p.AdaPerUtxoByte),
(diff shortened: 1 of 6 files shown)


ledger/conway/pparams.go
```diff
@@ -138,6 +138,16 @@ func (p *ConwayProtocolParameters) Utxorpc() (*utxorpc.PParams, error) {
 		p.MaxBlockExUnits.Memory < 0 || p.MaxBlockExUnits.Steps < 0 {
 		return nil, errors.New("invalid execution unit values")
 	}
+	if p.MinFeeRefScriptCostPerByte != nil {
```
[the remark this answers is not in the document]
arepala-uml: Confirmed, thank you -- fixed in 3bf1936a. Moved the validation (nil-embedded-Rat check first, returning (nil, nil) for the same 'unset' case the zero-value voting-threshold fields already use, then the range check) into ratToUtxorpcRationalNumber itself, and removed the now-redundant standalone guard. Added TestConwayUtxorpc_MinFeeRefScriptCostPerByteNilEmbeddedRatDoesNotPanic, confirmed panicking before this fix and passing after.

ledger/conway/pparams.go
```diff
@@ -191,9 +201,124 @@ func (p *ConwayProtocolParameters) Utxorpc() (*utxorpc.PParams, error) {
...
+// happens for a ConwayProtocolParameters value that was never populated with
+// this particular threshold (e.g. constructed directly rather than decoded
+// from a full on-chain protocol-parameters value or genesis).
+func ratToUtxorpcRationalNumber(r cbor.Rat) *utxorpc.RationalNumber {
+	if r.Rat == nil {
+		return nil
+	}
+	return &utxorpc.RationalNumber{
+		// #nosec G115
+		Numerator: int32(r.Num().Int64()),
```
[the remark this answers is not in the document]
arepala-uml: Confirmed, thank you -- fixed in 3bf1936a. ratToUtxorpcRationalNumber now validates the same int32/uint32 range every sibling rational field in this function already checks (A0, Rho, Tau, execution-cost prices), returning an error instead of silently wrapping during the cast, and poolVotingThresholdsUtxorpc/drepVotingThresholdsUtxorpc propagate it. Added TestConwayUtxorpc_VotingThresholdOutOfRangeRejected (pool + drep cases), confirmed silently succeeding instead of erroring before this fix and correctly erroring after.

ledger/conway/pparams.go
```diff
@@ -191,9 +218,145 @@ func (p *ConwayProtocolParameters) Utxorpc() (*utxorpc.PParams, error) {
...
+// denominator returns an error instead of silently wrapping during the
+// int32/uint32 cast (review-caught on blinklabs-io/gouroboros#2292 -- this
+// helper originally cast unconditionally, unlike its callers' sibling
+// fields above).
+func ratToUtxorpcRationalNumber(r cbor.Rat) (*utxorpc.RationalNumber, error) {
+	if r.Rat == nil {
+		return nil, nil
+	}
+	if r.Num().Int64() < math.MinInt32 || r.Num().Int64() > math.MaxInt32 ||
+		r.Denom().Int64() < 0 || r.Denom().Int64() > math.MaxUint32 {
```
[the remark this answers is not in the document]
arepala-uml: Confirmed, thank you -- fixed in de31f7c5. Auditing the rest of this function found the identical Int64()-truncation flaw in all five pre-existing rational-number guards this new check was modeled on (A0, Rho, Tau, both execution-cost prices), not just the new one -- fixed all six via a shared ratOutOfRange helper that compares the big.Int values directly via Cmp()/Sign() instead of casting to int64 first. New test covers both a pre-existing guard (A0) and the new voting-threshold path with a 2^64+1 numerator, confirmed silently succeeding before this fix and correctly erroring after, for both.

ledger/conway/pparams.go
```diff
@@ -191,9 +219,144 @@ func (p *ConwayProtocolParameters) Utxorpc() (*utxorpc.PParams, error) {
...
+		t.HardForkInitiation,
+		t.PpSecurityGroup,
+	}
+	thresholds := make([]*utxorpc.RationalNumber, len(rats))
+	for i, r := range rats {
+		rn, err := ratToUtxorpcRationalNumber(r)
+		if err != nil {
+			return nil, fmt.Errorf("threshold %d: %w", i, err)
+		}
+		if rn == nil {
```
wolf31o2, line 336: P2: This early return skips validation of later thresholds. For example, if `MotionNoConfidence` is unset but `CommitteeNormal` has a numerator outside the UTxO-RPC int32 range, the loop returns `(nil, nil)` at this point and silently omits the entire threshold list instead of returning an error. The DRep helper has the same pattern. Scan all entries, validate every populated threshold, and only decide whether the set is representable after the full pass; add a regression test with an unset earlier field and an invalid later field.

wolf31o2 (changes requested):

ledger/conway/pparams.go
```diff
@@ -191,9 +219,154 @@ func (p *ConwayProtocolParameters) Utxorpc() (*utxorpc.PParams, error) {
...
+		t.HardForkInitiation,
+		t.PpSecurityGroup,
+	}
+	thresholds := make([]*utxorpc.RationalNumber, len(rats))
+	allPopulated := true
+	for i, r := range rats {
+		rn, err := ratToUtxorpcRationalNumber(r)
+		if err != nil {
+			return nil, fmt.Errorf("threshold %d: %w", i, err)
+		}
```
arepala-uml: Fixed at b96009f2: both poolVotingThresholdsUtxorpc and drepVotingThresholdsUtxorpc now scan every threshold before deciding the set is unpopulated, so an out-of-range value after an earlier unset field is caught instead of silently skipped. Added TestConwayUtxorpc_VotingThresholdOutOfRangeRejectedAfterUnsetField (unset MotionNoConfidence + out-of-range CommitteeNormal for both structs), confirmed it fails without the fix.

ledger/conway/pparams.go
```diff
@@ -97,47 +98,87 @@ func (p *ConwayProtocolParameters) DRepDepositAmount() *big.Int {
...
+}
+
 func (p *ConwayProtocolParameters) Utxorpc() (*utxorpc.PParams, error) {
 	// sanity check
-	if p.A0 == nil ||
-		p.A0.Num().Int64() < math.MinInt32 ||
-		p.A0.Num().Int64() > math.MaxInt32 ||
-		p.A0.Denom().Int64() < 0 ||
-		p.A0.Denom().Int64() > math.MaxUint32 {
+	//
```
arepala-uml: Fixed at 579f1a8e: A0/Rho/Tau/ExecutionCosts.MemPrice/StepPrice's validation now also checks the embedded *big.Rat for nil (not just the outer *cbor.Rat pointer), returning the existing invalid-value error instead of reaching ratOutOfRange's Num()/Denom() calls on a nil receiver. Added TestConwayUtxorpc_MandatoryRatFieldNilEmbeddedRatRejectedNotPanic covering all five fields; confirmed each panics without the fix.

ledger/conway/pparams.go
arepala-uml: Fixed at b96009f2: both poolVotingThresholdsUtxorpc and drepVotingThresholdsUtxorpc now scan every threshold before deciding the set is unpopulated, so an out-of-range value after an earlier unset field is caught instead of silently skipped. Added TestConwayUtxorpc_VotingThresholdOutOfRangeRejectedAfterUnsetField (unset MotionNoConfidence + out-of-range CommitteeNormal for both structs), confirmed it fails without the fix.

ledger/conway/pparams.go
```diff
@@ -97,47 +98,74 @@ func (p *ConwayProtocolParameters) DRepDepositAmount() *big.Int {
...
-		p.ExecutionCosts.MemPrice.Denom().Int64() > math.MaxUint32 {
+		ratOutOfRange(p.ExecutionCosts.MemPrice.Rat) {
 		return nil, errors.New("invalid memory price rational number values")
 	}
 	if p.ExecutionCosts.StepPrice == nil ||
-		p.ExecutionCosts.StepPrice.Num().Int64() < math.MinInt32 ||
-		p.ExecutionCosts.StepPrice.Num().Int64() > math.MaxInt32 ||
-		p.ExecutionCosts.StepPrice.Denom().Int64() < 0 ||
-		p.ExecutionCosts.StepPrice.Denom().Int64() > math.MaxUint32 {
+		ratOutOfRange(p.ExecutionCosts.StepPrice.Rat) {
```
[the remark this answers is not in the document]
arepala-uml: Fixed at 579f1a8e: A0/Rho/Tau/ExecutionCosts.MemPrice/StepPrice's validation now also checks the embedded *big.Rat for nil (not just the outer *cbor.Rat pointer), returning the existing invalid-value error instead of reaching ratOutOfRange's Num()/Denom() calls on a nil receiver. Added TestConwayUtxorpc_MandatoryRatFieldNilEmbeddedRatRejectedNotPanic covering all five fields; confirmed each panics without the fix.

ledger/conway/pparams.go
```diff
@@ -191,9 +232,154 @@ func (p *ConwayProtocolParameters) Utxorpc() (*utxorpc.PParams, error) {
 			Memory: uint64(p.MaxBlockExUnits.Memory),
 			Steps:  uint64(p.MaxBlockExUnits.Steps),
 		},
+		MinFeeScriptRefCostPerByte:     minFeeRefScriptCost,
+		PoolVotingThresholds:           poolVotingThresholds,
+		DrepVotingThresholds:           drepVotingThresholds,
+		MinCommitteeSize:               uint32(p.MinCommitteeSize),
```
chrisguiney, line 249: **Merge blocker - silent narrowing on a newly mapped field.**

`p.MinCommitteeSize` is a `uint`, decoded straight from CBOR key 27 (`ConwayProtocolParameterUpdate.MinCommitteeSize *uint`) and assigned unchecked in `Update()`. On a 64-bit build this cast drops the high 32 bits: `MinCommitteeSize = 1 << 32` converts to `0`, and `Utxorpc()` returns no error.

That is the failure mode this change exists to remove - a plausible-looking value that makes a real difference invisible to a comparison - and it sits in the same struct literal where A0, Rho, Tau, both execution-cost prices, `MinFeeRefScriptCostPerByte` and every voting threshold now return an error rather than wrapping. The blanket `// #nosec G115` above the literal suppresses the linter that would otherwise flag it.

Rejecting it the same way as the rationals keeps the function internally consistent, e.g. alongside the existing `MaxTxExUnits` bound check:

```go
if uint64(p.MinCommitteeSize) > math.MaxUint32 {
	return nil, errors.New("invalid MinCommitteeSize value")
}
```

`TestConwayUtxorpc_GovernanceFields` uses `MinCommitteeSize: 7`, which passes either way, so this needs its own case asserting the error - `TestConwayUtxorpc_VotingThresholdOutOfRangeRejected` is the natural neighbour for it.

ledger/conway/pparams.go
```diff
@@ -97,47 +98,87 @@ func (p *ConwayProtocolParameters) DRepDepositAmount() *big.Int {
...
+// first: Int64() is undefined for a value that does not fit in int64 at all
+// (silently wraps rather than erroring, per math/big's own documentation),
+// so a value far outside range (e.g. 2^64+1) could pass an
+// Int64()-based comparison completely undetected -- review-caught on
+// blinklabs-io/gouroboros#2292, in both this file's pre-existing
+// rational-number guards below (A0, Rho, Tau, the execution-cost prices)
+// and the voting-threshold one added alongside them (ratToUtxorpcRationalNumber).
+// r itself must be non-nil; every caller already guards that separately
+// (e.g. "p.A0 == nil ||", short-circuiting before this runs).
+func ratOutOfRange(r *big.Rat) bool {
```
chrisguiney, line 113: **Non-blocking - the corrected pattern stops at Conway.**

`ratOutOfRange` and the `.Rat == nil` terms fix both defects here, but `ShelleyProtocolParameters.Utxorpc()`, `MaryProtocolParameters.Utxorpc()`, `AlonzoProtocolParameters.Utxorpc()` and `BabbageProtocolParameters.Utxorpc()` still carry the pre-fix form verbatim - `p.A0.Num().Int64() < math.MinInt32 || ...`, guarded only by `p.A0 == nil`. Both defects are live in those copies:

- a non-nil `*cbor.Rat` with a nil embedded `*big.Rat` panics in all four, where Conway now returns an error;
- an A0 numerator of `2^64+1` passes the guard in Shelley, Alonzo and Babbage and emits `PoolInfluence.Numerator = 1`, where Conway rejects it.

Both are pre-existing and outside this change's scope, so not a blocker for this pull request. Flagging it so the corrected pattern does not stay Conway-only - per-era duplication is exactly how one of these survives a fix.

chrisguiney (dismissed):
Reviewed at 579f1a8e.

The 0-indexed cost-model convention is corroborated by three independent in-tree sources: the `PlutusV1Key`/`PlutusV2Key`/`PlutusV3Key` constants in `ledger/alonzo/pparams.go`, the `CostModels[0]`/`[1]`/`[2]` lookups in `ledger/conway/rules.go:3228-3298`, and the real on-chain Preview fixture in `ledger/babbage/script_data_hash_used_languages_test.go`, whose `ScriptDataHash` only reproduces with keys 0 and 1. `TestConvertToUtxorpcCardanoCostModels_Mapping` now uses four distinct keys with distinct values, so a one-position shift fails it instead of passing symmetrically the way the old fixtures did.

Threshold ordering matches `utxorpc.PoolVotingThresholds` and `utxorpc.DRepVotingThresholds` field numbers 1-5 and 1-10, and `TestConwayUtxorpc_GovernanceFields` gives each position a distinct value, so a positional swap fails rather than passing on coincidence.

Both earlier review findings are fixed at this head; specifics are on those threads.

One blocker inline: `MinCommitteeSize` is narrowed to `uint32` without the range check every other newly mapped value in the same struct literal receives.

ledger/conway/pparams.go
chrisguiney: Verified fixed at 579f1a8e, and in both helpers.

`poolVotingThresholdsUtxorpc` and `drepVotingThresholdsUtxorpc` now scan every entry, run `ratToUtxorpcRationalNumber` on each, and only decide the set is unpopulated after the full scan completes (`allPopulated`); a range violation on any entry returns immediately as `threshold %d: ...`. The two helpers are identical in this respect, so the DRep copy you flagged is covered as well.

`TestConwayUtxorpc_VotingThresholdOutOfRangeRejectedAfterUnsetField` pins exactly the case you described - `MotionNoConfidence` unset with `CommitteeNormal` out of range - as separate subtests for the pool and DRep structs, and it fails if either helper is changed back to returning at the first unset entry. It is a genuine discriminator rather than a test that passes either way.

Unrelated blocker raised separately at line 238.

ledger/conway/pparams.go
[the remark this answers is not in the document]
chrisguiney: Verified at 579f1a8e, and the claim holds for all five fields.

A0, Rho, Tau, `ExecutionCosts.MemPrice` and `ExecutionCosts.StepPrice` each check the embedded `*big.Rat` for nil before `ratOutOfRange` dereferences it, and `TestConwayUtxorpc_MandatoryRatFieldNilEmbeddedRatRejectedNotPanic` covers each of the five as its own subtest; removing any single `.Rat == nil` term makes that field's subtest panic rather than fail, so the coverage is per-field and not aggregate.

The only other `ratOutOfRange` call site is `ratToUtxorpcRationalNumber`, which returns `(nil, nil)` on a nil embedded `Rat` before reaching it. That is the sole path taken by the optional `MinFeeRefScriptCostPerByte` and by both voting-threshold sets, so no rational field reaches `Num()`/`Denom()` unguarded in this era.

The same pattern is still unguarded in the Shelley, Mary, Alonzo and Babbage `Utxorpc()` implementations; raised inline at line 113 as pre-existing and out of scope here.

ledger/conway/pparams.go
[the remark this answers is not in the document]
chrisguiney: Already addressed at 579f1a8e. The standalone guard is gone; `MinFeeRefScriptCostPerByte` goes through `ratPtrToUtxorpcRationalNumber` into `ratToUtxorpcRationalNumber`, which returns `(nil, nil)` for a nil embedded `*big.Rat` before any `Num()` call, so an unset optional value is omitted rather than panicking. `TestConwayUtxorpc_MinFeeRefScriptCostPerByteNilEmbeddedRatDoesNotPanic` pins it.

ledger/conway/pparams.go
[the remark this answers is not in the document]
chrisguiney: Already addressed at 579f1a8e. `ratToUtxorpcRationalNumber` rejects an out-of-range numerator or denominator through `ratOutOfRange`, and both `poolVotingThresholdsUtxorpc` and `drepVotingThresholdsUtxorpc` propagate that error up through `Utxorpc()`. Covered by `TestConwayUtxorpc_VotingThresholdOutOfRangeRejected` for both threshold structs.

ledger/conway/pparams.go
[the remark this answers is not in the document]
chrisguiney: Already addressed at 579f1a8e. `ratOutOfRange` compares the `big.Int` values with `Cmp`/`Sign` rather than casting through `Int64()`, and all six guards in this file route through it. `TestConwayUtxorpc_ValueBeyondInt64RangeRejected` covers a numerator of 2^64+1 against both the pre-existing A0 guard and the voting-threshold path.

The same `Int64()`-based comparison remains in the Shelley, Mary, Alonzo and Babbage `Utxorpc()` implementations, where a numerator of 2^64+1 still passes and emits `PoolInfluence.Numerator = 1`; raised inline at line 113 as pre-existing and out of scope for this change.

wolf31o2 (dismissed):

ledger/conway/pparams.go
arepala-uml: Valid, fixed in 5674544. Added the same guard the other mandatory fields already have: `uint64(p.MinCommitteeSize) > math.MaxUint32` returns an error before the struct literal's uint32 conversion. Added TestConwayUtxorpc_MinCommitteeSizeOutOfRangeRejected (in-range and 1<<32 subtests; confirmed the out-of-range subtest fails without the fix).

ledger/conway/pparams.go
arepala-uml: Agreed, tracked as out of scope for this PR per your note. Filed #2296 to port the nil-embedded-Rat guard and the Cmp-based range check to Shelley/Mary/Alonzo/Babbage's Utxorpc() implementations rather than leave the fix Conway-only.

arepala-uml:
@chrisguiney addressed the merge blocker in 5674544: `MinCommitteeSize` (a uint) is now bounds-checked against `math.MaxUint32` before the struct literal's narrowing conversion, matching the pattern every other numeric field in `Utxorpc()` already follows. Added `TestConwayUtxorpc_MinCommitteeSizeOutOfRangeRejected` (confirmed it fails without the fix).

The non-blocking note (Shelley/Mary/Alonzo/Babbage still carrying the pre-fix pattern) is tracked as a follow-up: #2296.

Requested another review.

wolf31o2 (approved):

PR #1478, opened 2026-01-31
feat(ledger): calculate CBOR byte offsets

wolf31o2:
---
## Summary by cubic
Add CBOR byte offset calculation for all transactions in a block to enable fast partial CBOR extraction without full deserialization. Adds per-output ranges and a streaming decoder/API to collect offsets while decoding blocks.

- **New Features**
  - Single-pass offset extraction via common.StreamingBlockDecoder; combined decode+offsets helper NewBlockFromCborWithOffsets; adds ExtractOutputCbor/ExtractTransactionBodyCbor/ExtractWitnessCbor; TransactionLocation now includes per-output ranges.
  - Types/APIs: ByteRange, TransactionLocation, BlockTransactionOffsets, BlockWithOffsets, cbor.StreamDecoder, and ExtractTransactionOffsets. Witness offsets for datums by hash, redeemers by key (map and array), scripts by hash (native, Plutus V1/V2/V3). Supports Alonzo+ and Conway; metadata best-effort.

<sup>Written for commit 1f2337ae9dbfbdc5534ad89d32e5a8e6054143a3. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Per-transaction CBOR offsets for bodies, witnesses, metadata and nested components (outputs, datums, redeemers, scripts).
  * Position-aware CBOR streaming API with raw-byte access, header/array helpers, item iteration, and single-pass offset extraction.
  * Helpers to extract raw CBOR slices and a combined decode+offsets workflow returning both decoded block and offsets.

* **Bug Fixes / Errors**
  * New explicit error types for missing or conflicting transaction metadata.

* **Tests**
  * Added tests for streaming decoding and extraction helpers; test helper for genesis value creation.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #2302, opened 2026-09-13
fix(ledger): extract Dijkstra transaction offsets for both block body shapes

### LINKED ISSUE #2301: ExtractTransactionOffsets silently returns zero offsets for Dijkstra blocks in the current `block_body`/`block_transaction` shape
`ledger/common.ExtractTransactionOffsets` still recognises only the older Dijkstra prototype block layout. Against a Dijkstra block in the shape described by the Dijkstra CDDL vendored in this repo, the shape gate fails, the function falls through to the "short block" branch, and returns **zero** transaction offsets with **no error** — even though `ledger/dijkstra.DijkstraBlock` decodes the very same bytes and reports the transactions correctly.

## Impact

Any consumer that indexes block contents by CBOR byte offset gets an empty offset set for every non-empty Dijkstra block. In our case a from-genesis sync of a Dijkstra-era devnet stalls permanently at the first block that contains a transaction: the consumer's own sanity check ("block has N transactions but only 0 offsets extracted") fails, the block pipeline restarts, and the node retries the same block forever without making progress. The block is canonical — Haskell nodes and an independent chain indexer accept it.

Worse than the stall is the silent mode: because the fall-through returns `(&BlockTransactionOffsets{Transactions: []TransactionLocation{}}, nil)` rather than an error, a consumer without that sanity check would quietly store a block with none of its transactions indexed.

Two independent shape assumptions are wrong:

1. **Block body arity.** The gate requires the body to be a 4-element array (the older `[invalid_transactions, transactions, leios_certificate, peras_certificate]` form). The current CDDL defines it as 3 elements.
2. **Block transaction arity.** The gate requires each transaction to be a 3-element array. The current CDDL gives block transactions a trailing `is_valid` flag, making them 4 elements.

## Evidence: the repo's own CDDL disagrees with the walker

`ledger/dijkstra/testdata/dijkstra.cddl` lines 3-8 and 107-111:

```
block = [header, block_body]

; Transactions in a block carry a trailing `is_valid` flag, which is set by
; the block producer.
block_transaction =
  [transaction_body, transaction_witness_set, auxiliary_data/ nil, bool]

block_body =
  [ transactions      : [* block_transaction]
  , leios_certificate : leios_certificate/ nil
  , peras_certificate : peras_certificate/ nil
  ]
```

The era decoder already follows this. `ledger/dijkstra/dijkstra.go:276` `DijkstraBlockBody.UnmarshalCBOR` accepts **both** arities and treats the 4-element form as legacy:

```go
if len(items) != 3 && len(items) != 4 {
    return fmt.Errorf(
        "invalid Dijkstra block body: expected 3 components, got %d",
        len(items),
    )
}
legacy := len(items) == 4
```

and `ledger/dijkstra/dijkstra.go:1825` `newDijkstraBlockTransactionFromCbor` requires the current 4-element block transaction:

```go
if len(txArray) != 4 {
    return nil, fmt.Errorf(
        "invalid Dijkstra block transaction: expected 4 components, got %d",
        len(txArray),
    )
}
```

The offset walker was never updated alongside it.

## Root cause (file:line)

`ledger/common/common.go:1279` `isDijkstraBlock` — rejects the current body and transaction arities:

```go
if len(bodyParts) != 4 {
    return false
}
...
if len(txs) > 0 {
    var tx []cbor.RawMessage
    if _, err := cbor.Decode([]byte(txs[0]), &tx); err != nil {
        return false
    }
    if len(tx) != 3 {
        return false
    }
}
```

(the `!= 4` is at `common.go:1287`, the `!= 3` at `common.go:1301`.)

With `isDijkstraBlock` false, `ExtractTransactionOffsets` (`common.go:1558`) reaches its generic short-block branch at `common.go:1576`. A Dijkstra block is a 2-element array, so `len(blockArray) < 3` holds and it returns an empty offset set and a nil error:

```go
if isDijkstraBlock(blockArray) {
    return extractDijkstraTransactionOffsets(cborData, blockArray)
}

if len(blockArray) < 3 {
    // Block doesn't have separated components (e.g., Byron EBB)
    // Return empty slice instead of nil to prevent nil pointer dereference
    return &BlockTransactionOffsets{Transactions: []TransactionLocation{}}, nil
}
```

`extractDijkstraTransactionOffsets` (`common.go:1314`) carries the same stale assumptions, so simply relaxing the gate is not sufficient. It rejects a 3-element body at `common.go:1340` and `common.go:1367`, unconditionally skips a leading `invalid_transactions` field at `common.go:1377` (in the current form the first field *is* `transactions`), and rejects a 4-element transaction at `common.go:1444` and `common.go:1459`.

Empty Dijkstra blocks do not trip this, which is why a sync can cross the era boundary and run for a while before wedging: consumers typically skip offset extraction when the decoded block has no transactions, and the first few post-fork blocks were empty.

## Reproduction

Self-contained, using only fixtures already in this repo (`ledger/dijkstra/testdata/musashi_dijkstra_block.hex` for the header/certificates and `ledger/dijkstra/testdata/cardano_ledger_dijkstra_w30_tx.hex` for a transaction). The same block is assembled twice, once in the legacy 4-element-body form and once in the CDDL form, and passed to both the era decoder and the offset walker:

```
--- legacy prototype block, 1 tx (body=4, tx=3) - CONTROL ---
  bytes=16001 shape: top=2 body=4 txs=1 tx0_arity=3
  gouroboros dijkstra.DijkstraBlock decode: OK, Transactions()=1
  common.ExtractTransactionOffsets: 1 offsets
  OK: offsets cover all transactions
--- current CDDL block, 1 tx (body=3, tx=4) - REPRO ---
  bytes=16001 shape: top=2 body=3 txs=1 tx0_arity=4
  gouroboros dijkstra.DijkstraBlock decode: OK, Transactions()=1
  common.ExtractTransactionOffsets: 0 offsets
  >>> consumer fails: "block has 1 transactions but only 0 offsets extracted"
```

The control proves the walker is correct for the legacy layout and that only the arity change breaks it.

This also pins down the on-chain block without needing its raw bytes: the legacy body form requires 3-element transactions (`DijkstraBlockBody.UnmarshalCBOR` calls `newDijkstraTransactionFromCbor(rawTx, false)`, which rejects a 4-element transaction with "dijkstra transactions in blocks cannot include is_valid"), and that form extracts offsets correctly per the control above. So a block that decodes to one transaction yet yields zero offsets can only be in the current 3-element-body form.

## Observed in the wild

From-genesis sync of a Dijkstra-era devnet (1 s slots, `epochLength` 21600). Epochs 0-3 synced clean with zero header verification failures. The hard fork into the Dijkstra era:

```
level=INFO msg="emitted hard fork event" from_era=6 to_era=7 epoch=4 slot=86400 component=ledger
```

Then, 61 identical attempts on the first Dijkstra block containing a transaction, with a pipeline restart after each, chain pinned at the preceding slot:

```
level=WARN msg="block processing failed, restarting pipeline" error="process block batch: compute CBOR offsets for block at slot 86855: transaction count mismatch: block has 1 transactions but only 0 offsets extracted"
```

Independent chain index for that block: height 4277, slot 86855, hash `adb23531ebb61891…`, `tx_count` 1, size 829, no Leios certificate, no EB announcement. The two preceding Dijkstra-era blocks (slots 86828 and 86838) had empty bodies and passed. The following blocks at slots 86863 and 86878 are the same 1-tx template, so the stall is not specific to one block.

## Suggested fix

In `ledger/common/common.go`, teach both Dijkstra helpers the current layout while keeping the legacy one working, mirroring what `DijkstraBlockBody.UnmarshalCBOR` already does:

- `isDijkstraBlock`: accept a body of 3 **or** 4 elements; locate the transactions array at index 0 for the 3-element form and index 1 for the 4-element form; accept a first transaction of 3 **or** 4 elements.
- `extractDijkstraTransactionOffsets`: same body-arity branch; skip the leading `invalid_transactions` field only in the 4-element form; accept 4-element transactions and decode only the first three components for offsets, ignoring the trailing `is_valid` (it is a bool, not a byte range, so `TransactionLocation` needs no new field).

Consider also making the fall-through non-silent: if the top-level array has fewer than 3 elements and the block is not recognised as Byron or Dijkstra, returning an error instead of an empty offset set would turn a future layout change into a loud failure rather than silently unindexed transactions.

Happy to open a PR with the fix plus a table-driven test covering both body arities and both transaction arities.

chadaibbhot-git:
Fixes #2301

## Mechanism

`ExtractTransactionOffsets` only recognised the prototype-2026w27 Dijkstra block layout: a four-element `block_body` whose transactions are three-element arrays. The Dijkstra CDDL vendored in this repo (`ledger/dijkstra/testdata/dijkstra.cddl`) defines `block_body` as three elements `[transactions, leios_certificate, peras_certificate]` and gives each `block_transaction` a trailing `is_valid` flag, making it four elements. `DijkstraBlockBody.UnmarshalCBOR` was taught both forms in #2284, but the offset walker's shape gate was not, so `isDijkstraBlock` returned false for a current-shape block and `ExtractTransactionOffsets` fell through to the generic short-block branch — a Dijkstra block is a two-element array, so `len(blockArray) < 3` held and it returned zero transaction offsets with a nil error. A consumer that indexes block contents by CBOR byte offset then either stalls on the first non-empty Dijkstra block (its own "N transactions but 0 offsets" check fails and the block pipeline restarts forever) or, without such a check, silently stores the block with none of its transactions indexed.

## The fix

`ledger/common/common.go` only:

- `isDijkstraBlock` is replaced by `classifyDijkstraBlock`, which reports whether a two-element array is a Dijkstra block and which `block_body` shape it uses (current three-element, or legacy prototype-2026w27 four-element), mirroring `DijkstraBlockBody.UnmarshalCBOR`.
- `extractDijkstraTransactionOffsets` takes the shape: it reads the transactions array at index 0 for the current form and index 1 for the legacy form, and skips the leading `invalid_transactions` field only for the legacy form.
- Transactions of either arity are accepted. The trailing `is_valid` is a bool, not a byte range, so it is simply not walked and `TransactionLocation` is unchanged.
- A two-element block whose `block_body` or transaction arity is not understood now returns an error instead of an empty offset set, so a future layout change fails loudly instead of silently dropping transactions. Two-element values that are not Dijkstra blocks at all (their `block_body` or transactions field is not a CBOR array) keep the previous fall-through.

No other era's path is touched; the Byron and Shelley+ branches are unchanged.

## Tests

New table-driven tests in `ledger/common/dijkstra_offsets_test.go`, with fixtures assembled from the in-tree Dijkstra test data (`ledger/dijkstra/testdata/musashi_dijkstra_block.hex` for the header and the leios/peras certificate fields, `ledger/dijkstra/testdata/cardano_ledger_dijkstra_w30_tx.hex` for the transaction), so both shapes are real. Each case first asserts that `dijkstra.DijkstraBlock` decodes the fixture and reports the expected transaction count, then that the walker agrees; every recorded body/witness/auxiliary-data range must round-trip to the exact CBOR of the corresponding transaction element.

- `TestExtractTransactionOffsetsDijkstraBlockShapes` — legacy body with 0/1/2 transactions (control, green before and after) and current body with 0/1/2 transactions. Red on main: `"[]" should have 1 item(s), but has 0` for the current-body cases. Also asserts that no recorded range extends into the trailing `is_valid` element.
- `TestExtractTransactionOffsetsDijkstraMalformed` — two-element body, five-element body, and a five-element transaction under each body shape. Red on main: `An error is expected but got nil` for all four.
- `TestExtractTransactionOffsetsTwoElementNonDijkstra` — pins the preserved fall-through for two-element values that are not Dijkstra blocks (green before and after).

Existing `TestExtractTransactionOffsetsDijkstra` and `TestExtractTransactionOffsetsDijkstraUsesWireOffsets` (both legacy-shape) still pass unchanged.

## Verification

- `go build ./...`, `go vet ./...` clean.
- `go test ./ledger/...` all green; `go test -race ./ledger/common/ ./ledger/dijkstra/` green.
- `golangci-lint run ./...` — 0 issues; `nilaway -include-pkgs=github.com/blinklabs-io/gouroboros ./...` — 0 issues.
- Mutation checks with the tests in place: reverting the whole change turns the current-body and malformed cases red; classifying a malformed body as "not Dijkstra" turns the malformed cases red; skipping `invalid_transactions` unconditionally turns the current-body cases red; restricting the transaction arity check to three elements turns the current-body cases red.

Related: #2284 (aligned the era decoder with prototype-2026w36 but not the offset walker), #1861 (introduced the Dijkstra branch in the walker), #1670, #1478.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved transaction offset extraction for both current and legacy Dijkstra block formats.
  - Added support for transactions with optional validity flags.
  - Malformed or unsupported Dijkstra block layouts now return clear errors instead of being processed silently.
  - Preserved correct handling of non-Dijkstra two-element values.

- **Tests**
  - Added coverage for legacy and current layouts, transaction counts, CBOR range accuracy, validity flags, and malformed blocks.


ledger/common/common.go
```diff
@@ -1265,82 +1265,190 @@ func extractByronOutputOffsets(
...
-		return false
+	// block_body must be an array; anything else is not a Dijkstra block.
+	bodyParts, ok := cborArrayItems([]byte(blockArray[1]))
+	if !ok {
+		return dijkstraShapeNone, nil
 	}
-	if len(bodyParts) != 4 {
-		return false
+	txField, shape := dijkstraBodyTxField(len(bodyParts))
+	if shape == dijkstraShapeMalformed {
```
wolf31o2, line 1361: Gate this error on a Dijkstra-compatible header. The prior behavior returned empty offsets with nil error for arbitrary two-element values, but this now errors for `[bytes(0x01), [[], null]]` without checking the header. Preserve the non-Dijkstra fall-through while retaining errors for headers accepted by `DijkstraBlockHeader`; add a regression test.

wolf31o2 (changes requested):
The current-head change needs one correction before approval.

wolf31o2 (approved):

PR #2304, opened 2026-09-13
fix(script): normalize script-visible transaction data

wolf31o2:
cardano-ledger rebuilds every value a Plutus script observes instead of carrying the transaction CBOR into the script context, so a `Constr` reaches the script with the indefinite-length field list the Plutus encoder writes. Decoded redeemer data, witness datums and inline datums were passed through unchanged, and `Decode` preserves the definite-length form a transaction was built with, so `serialiseData` returned different bytes than cardano-node for the same value.

Preprod transaction `2c528f4e28e2b8fd47539fe51eca8b7dafced3d1b82f48edf4a57e075b525ff5` at slot 133016611 mints an asset named `blake2b_256(serialiseData(seed TxOutRef))`, from a redeemer that is definite-encoded on the wire:

```
minted on chain   c714816533babf58a158870eaa49db187e1e1f83f67579070d2025974129a858
                  = blake2b_256(d8799f58203fd10d2c...d501ff)   indefinite
computed here     5e13e57434d51b2bf8a3693608a308936b206485ca9bc405dbee44f6af21668b
                  = blake2b_256(d879825820 3fd10d2c...d501)    definite
```

The minting policy compared the two, disagreed, and called `error`, so a canonical block was rejected deterministically.

`data.Normalize` is called at each boundary where decoded data becomes script-visible: redeemer data and `txInfoData`, the spending purpose datum, the inline datum rendered into a Babbage `TxOut`, the inline datum `WithZeroAdaAsset` forwards when rendering a resolved input for a V1/V2 context, the two Dijkstra V4 renderers, and the V1/V2 and Dijkstra evaluation paths. Datum hashing still reads the original bytes. `plutigo` moves to `v0.6.1`, which is where `data.Normalize` was added.

Four new tests cover the redeemer, `txInfoData`, spending-datum and V1/V2 resolved-input paths; each fails without the change with the definite-length bytes. The V1/V2 renderer is a separate path from the era `TxOut` `ToPlutusData`, and it used `Clone`, which carries the wire encoding. The `Withdraw` and `Voting` script context vectors recorded those bytes and noted that the Aiken vectors they came from were wrong to use indefinite; the mint above shows the reverse, so they move to the Aiken form.

Against a dingo build carrying this change, transaction `2c528f4e...` evaluates successfully at 383880 memory / 131288901 steps, inside its declared 3000000 / 1000000000, and the exact-budget fixtures that pin producer-declared budgets are unchanged.

Related: blinklabs-io/dingo#3860


ledger/babbage/babbage.go
```diff
@@ -829,9 +829,13 @@ func (o BabbageTransactionOutput) ToPlutusData() data.PlutusData {
...
 		)
 	case o.DatumOption.data != nil:
+		// Normalize: cardano-ledger rebuilds every script-visible value, so a
+		// script always observes the encoding the Plutus encoder writes, never
+		// the definite/indefinite-length choice this transaction was built
+		// with. serialiseData exposes the difference.
 		datumOptionPd = data.NewConstr(
 			2,
-			o.DatumOption.data.Data,
+			data.Normalize(o.DatumOption.data.Data),
```
arepala-uml, line 838: This normalizes BabbageTransactionOutput's own ToPlutusData, but TxInfoV2 (real PlutusV2 evaluation in conway/rules.go, dijkstra/rules.go) renders Inputs/ReferenceInputs/Outputs via a separate path, WithZeroAdaAsset.ToPlutusData in wrappers.go, whose TransactionOutput case calls tmp.Data.Clone() not data.Normalize(tmp.Data). TestTxInfoV2InputDatumNormalized, added to context_normalize_test.go, fails at this PR's head: got d879825820<32x0x11>01 (definite), want d8799f5820<32x0x11>01ff (indefinite); passes once wrappers.go normalizes the datum too.

arepala-uml (changes requested):
PlutusV2 TxInfo rendering still forwards inline datums unnormalized on one more script-visible path; a test confirms it, detailed in the inline comment.

arepala-uml (approved):

Issue #2074, opened 2026-08-25
chainsync: synchronize shutdown state


wolf31o2:
Merged #2321 corrects terminal-message delivery during `Stop`; it does not
resolve the in-flight operation ownership and stop/restart concerns below.

`syncLoop` now receives readiness and done channels captured under
`lifecycleMutex`, so the original shared-field read in that loop is repaired.

`GetAvailableBlockRange` still receives directly from
`c.readyForNextBlockChan` while holding only `busyMutex`. `stopRunning` clears
that field under `lifecycleMutex` and proceeds after its bounded attempt to
acquire `busyMutex`. The remaining concern is this shutdown/range-query
boundary, not the already repaired sync-loop receive.

`Stop` joins the sync loop before marking the client stopped, but does not
join public range/Sync calls or `GetCurrentTip`'s background request worker.
Those paths retain accesses to `Protocol` and reply channels that `Start`
replaces in `initProtocol`. A readiness snapshot alone does not establish
safe operation lifetimes across stop/restart.

Acceptance criteria:

- Exercise shutdown while a range query is waiting and while readiness is
  delivered, with controlled ordering and race instrumentation.
- Synchronize readiness-channel access and preserve channel ownership across
  stop/restart, without holding a lifecycle lock across a blocking receive.
- Preserve explicit cancellation/shutdown results and idempotent concurrent
  stop behavior.
- Account for in-flight range and Sync calls and background tip requests
  before permitting a new lifecycle generation; verify bounded shutdown and
  fresh-generation operation without lock-order deadlocks.
- Retain sync-loop and initial-request shutdown regressions.

The remaining interleaving needs a targeted regression; source inspection is
not a claim that an executable race reproduction has completed.

PR #2321, opened 2026-09-14
fix(protocol): wait for shutdown message delivery

wolf31o2:
BlockFetch and ChainSync `Stop` previously checked queue accounting rather
than transport delivery, so they could report success before sending
`ClientDone` or `Done`.

Wait for delivery within the existing 250ms budget and return delivery
errors after shutdown. Add a context-aware delivery-wait API, gated transport
regressions, and BlockFetch test cleanup that closes connections on failed
assertions without competing error-channel readers.

Related to #2074.

## Summary by Cubic

`SendMessageContextAndWait` is used in both shutdown paths. Delivery errors,
including `context.DeadlineExceeded`, are returned; an already-shutting-down
protocol still counts as stopped. Transport tests cover blocked and
successful delivery.


cryptodj413 (approved):
Verified in an isolated worktree: `go build ./...` and `go test -race` clean across all affected/adjacent protocol packages. Fail-before revert confirms all 6 delivery/timeout tests genuinely fail without the fix and pass with it. Traced `SendMessageContextAndWait`/`enqueueMessage` for deadlock/leak risk on context expiry — none found; message accounting unwinds cleanly. Confirmed `blockfetch`/`chainsync` were the only two clients using the old `WaitSendQueueDrained`-without-checking-the-result pattern (grepped `origin/main`), so the fix's scope is complete. `SendMessageAndWait`'s other callers (e.g. leiosnotify) are unaffected since they still pass `context.Background()`.

Note for the record: `Client.Stop()` on both protocols now has a documented, narrower contract — it can return `context.DeadlineExceeded` under normal but slow/congested shutdown, where it previously always returned nil for these paths. Confirmed `Connection.shutdown()` doesn't call these `Stop()` methods and won't propagate this into `ErrorChan`, so this only affects a direct caller of `chainsync.Client.Stop()`/`blockfetch.Client.Stop()`.

CodeRabbit and Cubic did not review this head (rate-limited); no actionable bot findings to reconcile.

PR #2326, opened 2026-09-14
fix(ledger): allow duplicate keys in nested metadata maps

### LINKED ISSUE #2323: ledger: metadata map decode wrongly rejects duplicate keys upstream cardano-ledger tolerates
## Summary

`fix(cbor): raise the nesting cap past what a transaction can encode` (#2252, in `v0.204.5`/`v0.204.6`) rewrote `ledger/common/metadata.go` and, alongside its intended nesting-depth bound, added an unconditional rejection of duplicate keys inside a metadata map (`decodeMetadatumAt`'s map-decode branch, backed by `metadatumKeyDigest`). This is a real wire-compatibility regression: it now rejects any transaction/block whose auxiliary data contains a metadata map with a repeated key, at raw CBOR unmarshal time, in every era Shelley through Conway -- not just in a downstream consumer's convenience layer.

## Reproduction (bypassing dingo entirely)

```go
var aux common.ShelleyAuxiliaryData
err := aux.UnmarshalCBOR(mustHex("a11902d1a2016161016162")) // {721: {1: "a", 1: "b"}}
// err = "failed to decode Shelley auxiliary data: duplicate key in metadata map"
```

`ShelleyAuxiliaryData.UnmarshalCBOR`, `ShelleyMaAuxiliaryData.UnmarshalCBOR`, and `AlonzoAuxiliaryData.UnmarshalCBOR` (`ledger/common/metadata.go:729`, `:789`, `:863` in v0.204.6; Conway reuses the Alonzo type) all call `DecodeMetadatumRaw` unconditionally as part of ordinary CBOR unmarshal, before any ledger validation rule runs.

Discovered downstream via `blinklabs-io/dingo`'s existing test `TestDuplicateNestedEncodedKeyIsUnavailableInJSON` (`database/plugin/metadata/labelcodec/labelcodec_test.go`), which asserts the historically-correct behavior: such a map should still decode (as one indexable entry with the raw CBOR preserved), just flagged as JSON-unavailable since a duplicate-key map can't be faithfully represented as JSON. That test started failing purely from bumping the gouroboros dependency, no dingo-side change.

## Verified against real upstream cardano-ledger -- this is not a "matching enforcement elsewhere" correctness fix

The new code's own comment (`metadatumKeyDigest`) says duplicate-key rejection is added "to match the duplicate-key enforcement the CBOR decode modes apply elsewhere." That is a claim about gouroboros's own internal consistency, not something checked against upstream. It does not hold for this specific position:

`IntersectMBO/cardano-ledger`'s `libs/cardano-ledger-core/src/Cardano/Ledger/Metadata.hs` decodes a nested metadata map as an association list, unconditionally consing every pair:

```haskell
decodeMapN :: Int -> [(Metadatum, Metadatum)] -> Decoder s [(Metadatum, Metadatum)]
decodeMapN !n acc =
  case n of
    0 -> return $! reverse acc
    _ -> do
      !tm  <- decodeMetadatum
      !tm' <- decodeMetadatum
      decodeMapN (n - 1) ((tm, tm') : acc)
```

No key-equality check exists anywhere in this function or `decodeMetadatum`. `git log --follow` on this file shows no commit has ever added duplicate-key handling here -- this has been stable behavior since the file existed. `Cardano.Ledger.Shelley.TxAuxData`, `Cardano.Ledger.Allegra.TxAuxData`, and `Cardano.Ledger.Alonzo.TxAuxData` (Conway reuses Alonzo's) all reuse this same decoder unchanged, so it's era-invariant.

(Note: a genuinely different, correctly era-gated rule *does* exist in real cardano-ledger for the *outer* label map -- `Word64 -> Metadatum` -- decoded via the version-gated `decodeMap` in `cardano-ledger-binary`'s `Decoder.hs:770-797` (rejects duplicates only from protocol version >= 9; silently keeps-last below that). That rule is unrelated to this issue, which is specifically about the *inner*, per-label `Metadatum` value's own nested maps, decoded by the unconditional `decodeMapN` above.)

## gouroboros already has the right pattern for this and didn't use it

`cbor/decode.go:137-170` already defines a `DecodeLenient` mode with `DupMapKeyQuiet` semantics, reserved for "structures that have historically permitted duplicate keys, e.g. pre-Conway MultiAsset value/mint maps." The new metadata decoder bypasses this established pattern entirely and hand-rolls an unconditional hard rejection with no era gate -- the opposite of how this class of "reference client tolerates a wire-level duplicate" problem is normally handled here (see also `blinklabs-io/gouroboros#1989` / `blinklabs-io/dingo`'s `ledger/replay_recovery.go:405-407`, the same failure mode for duplicate transaction inputs, where non-terminal/tolerant handling was chosen deliberately).

## What a fix needs to do

In `ledger/common/metadata.go`'s map-decode branch of `decodeMetadatumAt`, remove (or make non-fatal, e.g. route through `DecodeLenient`) the unconditional `"duplicate key in metadata map"` rejection for the recursive `Metadatum` map-decode path, restoring upstream's actual behavior: preserve every pair (including duplicates) as an ordered list. No era gate is needed for this specific path -- upstream never rejects it, at any era.

A semantic-JSON-conversion layer that needs one value per key (gouroboros's own, or a downstream consumer's) can still flag such a map as JSON-unavailable while preserving the raw CBOR -- that's the existing, correct design the now-failing dingo test asserts.

## Not confirmed

No specific on-chain transaction hash with this exact shape was found in this pass -- Koios wasn't queried for it (no practical way to search for "a nested metadata map with a duplicate key" without scanning raw tx CBOR at scale). Absence of a found instance isn't evidence of absence, given CIP-25 NFT-minting tooling of varying quality has been active since 2021 and metadata was never validated this strictly by any reference client until now.

chrisguiney:
## Problem

`#2252` (shipped in v0.204.5/v0.204.6) added an unconditional rejection of
duplicate keys inside a metadatum's own nested map (e.g. a CIP-25/CIP-721
label value), alongside its intended nesting-depth bound. Upstream
`cardano-ledger`'s `decodeMapN`
(`libs/cardano-ledger-core/src/Cardano/Ledger/Metadata.hs`) conses every pair
unconditionally at this position and has never rejected a duplicate key here,
at any era — confirmed directly against the current file on
`IntersectMBO/cardano-ledger@master`.

This broke raw CBOR unmarshal for any real transaction or block carrying such
metadata: `ShelleyAuxiliaryData`, `ShelleyMaAuxiliaryData`, and
`AlonzoAuxiliaryData` (Conway reuses Alonzo) all call the affected decode
path unconditionally, before any ledger validation rule runs.

## Change

- `decodeMetadatumAt`'s map-decode branch no longer rejects a duplicate key;
  every pair is preserved as an ordered list, matching upstream.
- Removes `metadatumKeyDigest`, the SHA-256 key-identity helper that existed
  only to serve the removed check (no other caller referenced it).
- The separate, correctly era-gated rule for the *outer* `Word64`-keyed label
  map (rejected only from protocol major >= 9, via `cbor.Decode`'s
  `DupMapKeyEnforcedAPF` mode) is untouched and still enforced; covered by a
  new dedicated test.
- The nesting-depth bound from `#2252` is untouched and still enforced.

Fixes #2323

---
## Summary by cubic
Fixes duplicate-key handling in nested metadata maps so real transactions and blocks carrying such metadata (for example, CIP-25/CIP-721 label values) unmarshal correctly instead of failing before ledger validation.

**Bug Fixes**
- Removes the duplicate-key rejection and the SHA-256 key-digest helper that only served it.
- Preserves every pair in nested maps as an ordered list, matching upstream `cardano-ledger` behavior.
- Keeps the nesting-depth bound from #2252 unchanged.
- Keeps the outer `Word64` label map's duplicate-key rejection intact.

<sup>Written for commit c18164608165a966bfa9e011ffad1bcb7f43681f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2326?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Nested transaction metadata maps now preserve duplicate key-value pairs during decoding instead of rejecting them.
  - Duplicate keys remain supported across supported metadata formats and key encodings.
  - Duplicate labels in outer auxiliary-data maps continue to return validation errors.
- **Tests**
  - Added coverage for duplicate-key handling and preservation across Shelley, ShelleyMa, and Alonzo metadata.


cryptodj413 (approved):
Verified in an isolated worktree: `go build ./...` and `go test -race ./ledger/common/...` clean, including the fuzz seed corpora.

Fetched the actual upstream cardano-ledger source (`Cardano.Ledger.Metadata.decodeMapN`) rather than trusting the PR's citation — confirmed it conses every key/value pair unconditionally with no duplicate check, matching the fix.

Fail-before revert (swapped in the pre-PR `metadata.go` with the new tests in place): all 4 duplicate-key `TestMetadatumDecodeShapes` subtests and all 3 `TestDuplicateNestedMetadataKeyDecodes` era subtests fail with the exact `duplicate key in metadata map` error from #2323. Reapplying the fix passes.

Confirmed the fix is correctly scoped to the inner per-label metadatum map decode only — `TestOuterAuxiliaryDataLabelMapRejectsDuplicateKeys` passes, so the outer Word64-label map's (unrelated, correct) duplicate rejection is untouched. Also confirmed the nesting-depth bound from #2252 (the fix this regression rode in on) is preserved — `TestDeeplyNestedMetadatumDecodesInLinearSpace` still passes. No dangling references to the removed digest machinery.

PR #2329, opened 2026-09-14
fix(localstatequery): cache current era result

### LINKED ISSUE #2026: localstatequery: publish and cache current era safely
`getCurrentEra` in `protocol/localstatequery/client.go` still returns a successful era query without caching it, so repeated era-dependent queries repeat the era request.

The earlier publication-order race in `handleAcquired` was corrected by #2291: cached era invalidation now precedes the acquisition result notification. #2320 separately resets client snapshot and era state after acquisition failure.

Remaining acceptance criteria:

- Cache a successful era result for the acquired snapshot; do not cache failed queries.
- Invalidate the cached era when acquisition, reacquisition or release changes the snapshot state, including failed reacquisition. Do not retain an era across a snapshot change merely because the previous era was known.
- Add race-enabled wire coverage proving repeated queries reuse the result and snapshot transitions trigger a fresh era query.

wolf31o2:
Cache successful current-era queries for the acquired snapshot and retain
invalidation on snapshot transitions. Add wire coverage for cache hits,
reacquisition, release, and failed queries.

Fixes #2026.

## Summary by Cubic

Consecutive current-era queries reuse the acquired snapshot's cached value;
failed queries do not populate the cache.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved current-era handling during protocol shutdown.
  * Reuses successfully queried era information during the same acquisition.
  * Refreshes cached era information after reacquisition or release, including failed reacquisition attempts.
  * Prevents failed queries from populating cached era information.


protocol/localstatequery/client.go
```diff
@@ -1858,5 +1858,6 @@ func (c *Client) getCurrentEra() (int, error) {
 	if err := c.runQuery(query, &result); err != nil {
 		return -1, err
 	}
+	c.currentEra = result
```
chrisguiney, line 1861: This assignment makes the cache-hit branch at the top of getCurrentEra reachable, and that branch does not check for protocol shutdown. After the connection stops (Close, a state timeout, a muxer error) GetCurrentEra returns the last era with a nil error; with this line removed the same call returns an error. Reproduction on this harness: GetCurrentEra, h.client.Close(), wait on Client.DoneChan(), GetCurrentEra again returns (5, nil). Return protocol.ErrProtocolShuttingDown from the cache hit once DoneChan() is closed, with a wire test for it.

protocol/localstatequery/current_era_cache_wire_test.go
```diff
@@ -0,0 +1,172 @@
...
+	era, err = client.GetCurrentEra()
+	require.NoError(t, err)
+	require.Equal(
+		t,
+		int32(1),
+		h.queries.Load(),
+		"successful result should be cached",
+	)
+	require.Equal(t, 5, era)
+	require.NoError(t, client.Release())
```
chrisguiney, line 127: This sequence acquires explicitly after Release, so the reset in handleAcquired masks the one in release(). Deleting c.currentEra = -1 from release() or from handleFailure() leaves the package green under -race. Add Release followed directly by GetCurrentEra, and a ReAcquire the server fails with ErrAcquireFailurePointTooOld followed by GetCurrentEra; each must expect a second era query, and each fails against its own deletion.

chrisguiney (dismissed):
Two blockers: the new cache hit outlives connection shutdown, and the new tests do not pin release or failed-reacquisition invalidation, both acceptance criteria of #2026. At 7373ad7 and on a merge with current main, go test -race ./protocol/localstatequery passes (new tests also at -count=10) and golangci-lint reports 0 issues for the package.

arepala-uml (approved):

PR #2334, opened 2026-09-16
cbor: make MaxNestedLevels configurable

Quantumplation:
## Motivation

`MaxNestedLevels` and `MaxUntrustedNestedLevels` are fixed `const`s (16384). An application that must decode a structure nested deeper than that cannot raise the cap without forking the library.

Cardano's reference decoders impose no nesting limit on transaction metadata or Plutus data, so a node can accept a block whose CBOR nests deeper than any fixed cap. A decoder that rejects such a block then wedges the consumer on a block the chain already contains — we hit exactly this on a preview block (slot 122726746) whose Plutus data nests 257 levels; on an older release the cap was 256 and every sync consumer stalled on that one block.

Raising the default (as this repo already did, 256 -> 16384) reduces how often this happens but does not remove it: the cap is still fixed, and an operator indexing trusted archival data has no supported way to go higher.

## Change

Make `MaxNestedLevels` and `MaxUntrustedNestedLevels` package `var`s instead of `const`s, so an application can raise them. `maxDiagnosticNestedLevels` becomes a `var` for the same reason (it is defined in terms of `MaxNestedLevels`).

The decode modes are built once and cached on first use, so the value is read at first use. The doc comments note that the limit must be set before the first `Decode` call to take effect.

Defaults are unchanged, so existing behavior is preserved.

---
## Summary by cubic
Makes `MaxNestedLevels` and `MaxUntrustedNestedLevels` configurable variables instead of fixed constants, allowing applications to decode structures deeper than 16,384 levels without forking the library. Existing defaults remain unchanged.

- Set the limits before the first `Decode` call because decode modes are cached on first use.
- `MaxUntrustedNestedLevels` is initialized to `MaxNestedLevels` at startup; changing one later does not update the other.
- Diagnostic rendering reads `MaxNestedLevels` directly, so the configured limit applies there too.

<sup>Written for commit f88cc4f526da5012c3c834f4a329aa306e6957cf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2334?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Maximum nesting limits can now be configured at runtime before decoding begins.
  - Applications processing deeply nested trusted or archival data can adjust the supported depth.
  - Diagnostic processing uses the configured nesting limit.


cbor/diagnostic.go
```diff
@@ -23,7 +23,13 @@ import (
 	"strings"
 )
 
-const maxDiagnosticNestedLevels = MaxNestedLevels
+- var maxDiagnosticNestedLevels = MaxNestedLevels
```
chrisguiney, line 26: This is a diff fragment, not Go. go build ./... fails at cbor/diagnostic.go:26:1 with syntax error: non-declaration statement outside function body, and every package in the module fails to build; origin/main builds clean. The review suggestion it came from held a sketch rather than applicable code. Its underlying observation is valid though: maxDiagnosticNestedLevels copies the value once, so ParseDiagnostic still rejects at 16384 after MaxNestedLevels is raised. Read MaxNestedLevels directly here and drop the copy.

cbor/decode.go
```diff
@@ -46,13 +46,19 @@ var (
 // decoders. Cardano's reference decoders do not impose a nesting limit on
 // transaction metadata or Plutus data, so this cap is derived from the wire
 // size limit rather than an arbitrary structural rule.
-const MaxNestedLevels = 16384
+//
+// It is a var so an application that decodes deeper structures (for example an
+// indexer reading trusted archival data) can raise it. The decode modes are
+// built once and cached on first use, so set this before the first Decode call
+// for the change to take effect.
+var MaxNestedLevels = 16384
```
chrisguiney, line 54: No range check. fxamacker accepts MaxNestedLevels only in 4 to 65535 and DecMode construction errors outside it, while getDecMode caches that error under sync.Once forever. Setting 100000 makes every later Decode and Value.UnmarshalCBOR in the process return cbor: invalid MaxNestedLevels 100000, and resetting to 16384 does not recover, since the cached error persists. Setting 0 silently drops the cap to the library default of 32. Validate or clamp the value and document the bound rather than caching the failure.

cbor/decode.go
```diff
@@ -46,13 +46,19 @@ var (
...
 // MaxUntrustedNestedLevels names the limit used by decoders that process
-// peer-controlled data. Keep it equal to MaxNestedLevels because Decode is
-// also used by protocol message and block decoders; a separate strict mode
-// cannot protect those callers until they are migrated.
-const MaxUntrustedNestedLevels = MaxNestedLevels
+// peer-controlled data. It defaults to MaxNestedLevels because Decode is also
+// used by protocol message and block decoders; a separate strict mode cannot
+// protect those callers until they are migrated. Like MaxNestedLevels it is a
+// var; set it before the first Decode call for the change to take effect.
+var MaxUntrustedNestedLevels = MaxNestedLevels
```
chrisguiney, line 61: DecodeStrict has no caller outside its own test, so MaxUntrustedNestedLevels guards nothing today: peer-supplied blocks and mini-protocol messages reach cbor.Decode, which reads MaxNestedLevels. Raising MaxNestedLevels for archival data therefore also raises the nesting a remote peer can force on every wire object, and this copy cannot compensate since it takes the value once at package start. Measured: with MaxNestedLevels set to 20000 before first use, Decode accepts depth 20000 while DecodeStrict still rejects 16385. Say that here, or derive it when the mode is built.

cbor/decode.go
```diff
@@ -46,13 +46,19 @@ var (
 // decoders. Cardano's reference decoders do not impose a nesting limit on
 // transaction metadata or Plutus data, so this cap is derived from the wire
 // size limit rather than an arbitrary structural rule.
-const MaxNestedLevels = 16384
+//
+// It is a var so an application that decodes deeper structures (for example an
+// indexer reading trusted archival data) can raise it. The decode modes are
+// built once and cached on first use, so set this before the first Decode call
+// for the change to take effect.
```
chrisguiney, line 53: This contract holds only for the three cached decode modes. cbor/value.go:60 and cbor/diagnostic.go:98 read MaxNestedLevels on every call, so a later change does take effect there, unsynchronised, and any write concurrent with decoding is a data race on an exported package variable. Either say the value must be set before any decoding begins and before other goroutines start, or read it once into an unexported copy that all five sites share.

cbor/decode.go
```diff
@@ -46,13 +46,19 @@ var (
 // decoders. Cardano's reference decoders do not impose a nesting limit on
 // transaction metadata or Plutus data, so this cap is derived from the wire
 // size limit rather than an arbitrary structural rule.
-const MaxNestedLevels = 16384
+//
+// It is a var so an application that decodes deeper structures (for example an
```
chrisguiney, line 50: No test covers the new behaviour. cbor/nesting_test.go is written relative to cbor.MaxNestedLevels, so it passes at any value and cannot detect a default regression, a raised limit failing to apply, or the out-of-range path. Add cases pinning the default at 16384, raising the limit in a separate package and asserting both acceptance at the new depth and rejection at one past it, and covering a value outside 4 to 65535.

chrisguiney (changes requested):
Head does not compile: cbor/diagnostic.go holds a diff fragment instead of Go, the sole cause of the failing lint, go-test, build-examples and nilaway jobs. Base 6b002375 builds, vets and tests clean, so all four attribute to this change. Conventional Commits fails separately: commit f19d3f13 uses cbor as the type, and the workflow allows only feat, fix, docs, style, refactor, test, build, perf, ci, chore, revert, merge, wip. The PR title repeats it and becomes the squash subject.

wolf31o2:
Replaced with #2335

Issue #1700, opened 2026-04-21
ledger: implement CIP-0112 Observe script purpose


wolf31o2:
## Summary

CIP-0112 introduces a new Plutus script purpose ("Observe") and the related script context shape. gouroboros needs to decode and encode observe-purpose redeemers and script witnesses when the CIP activates on-chain.

## What Needs to Happen

1. Extend the redeemer tag enum and redeemer tuple decoding to accept the new observe purpose.
2. Update the witness set validation rules (era-gated) to accept observe redeemers when the CIP is active.
3. Add CBOR round-trip tests for observe redeemers and ensure existing era decoders still reject them where disallowed.

## Priority

P3 — enhancement, timing depends on Dijkstra / PlutusV4 activation.

## References

- [CIP-0112](https://cips.cardano.org/cip/CIP-0112)

wolf31o2:
plutigo-side tracking: blinklabs-io/plutigo#285 (Observing script purpose).

cryptodj413:
Attempted this in #2336 (closed without merging). The blocker, found by @chrisguiney in review: item 1 ("extend the redeemer tag enum ... to accept the new observe purpose") isn't safely implementable yet, because **no specification has assigned a numeric redeemer tag to Observe**.

- CIP-0112's own text only defines the purpose as `Observe Integer` (a constructor case, not a wire value).
- cardano-ledger has no Observe purpose at all. Checked directly against `Dijkstra/Scripts.hs`:
  ```haskell
  6 -> DijkstraGuarding <$> decCBOR
  n -> fail $ "Unexpected tag for DijkstraPlutusPurpose: " <> show n
  ```
  Tag 7 (or any value above 6) is simply unallocated — not reserved for Observe, and could be assigned to a different purpose entirely by a future cardano-ledger change.

Picking a placeholder number now (I used 7) and shipping it as a public Go constant + JSON wire name is actively harmful, not just premature:
- it makes `RedeemerTag.MarshalJSON` silently label arbitrary tag-7 data as `"observe"` instead of erroring on an unrecognized tag
- it breaks downstream consumers' `exhaustive` lint on the next gouroboros bump (confirmed against dingo's `api/blockfrost/adapter.go`), and the natural fix there would propagate `"observe"` onto a Blockfrost-compatible API field for data no spec defines that way

[...]

PR #2336, opened 2026-09-16
feat(ledger): recognize CIP-0112 Observe redeemer tag

cryptodj413:
## Summary

Refs #1700. Adds `RedeemerTagObserve` (CIP-0112's "Observe" script purpose) to the shared redeemer tag enum, per the issue's three scoped items:

1. **Enum + decoding**: `RedeemerTagObserve = 7` with JSON name-table support. CBOR tuple decoding already accepts any tag value with no bounds check (confirmed by inspection of the map-form and legacy array-form decode paths), so this only needed the name-map entries.
2. **Era-gated validation**: no era activates the purpose yet — [CIP-0112](https://cips.cardano.org/cip/CIP-0112) is still status `Proposed`, and the companion tracking issue in plutigo ([blinklabs-io/plutigo#285](https://github.com/blinklabs-io/plutigo/issues/285)) confirms activation is expected under PlutusV4/Dijkstra, not yet. `ValidateExtraneousRedeemers`, `BuildScriptPurpose`, `scriptPurposeBuilder`, and Dijkstra's own extraneous-redeemer check all explicitly treat the new tag as extraneous/unmatched, mirroring the existing `RedeemerTagGuarding` handling in each of those switches.
3. **Tests**: CBOR and JSON round-trip coverage for the new tag (`RedeemerKey`/`RedeemerValue` directly, and `ConwayRedeemers`' map-form CBOR path), plus rejection tests confirming Alonzo, Babbage, Conway, and Dijkstra all still reject an Observe-tagged redeemer.

This is intentionally additive-only: it does not wire `RedeemerTagObserve` into `script.ScriptPurposes` (unlike `RedeemerTagGuarding`, which is an active Dijkstra purpose there), since the CIP isn't active in any era. Follow-up work lands once the CIP activates.

## Evidence

- `go build ./...` — pass
- `go vet ./...` — pass
- `golangci-lint fmt -d` (full repo) — pre-existing, unrelated formatting drift only (same pattern present in untouched files); none of the flagged hunks touch lines this PR added

## Test plan

- [x] `go test -race ./ledger/...`

---
## Summary by cubic
Adds the CIP-0112 Observe redeemer tag (value 7) to the shared redeemer tag enum so it decodes and round-trips in CBOR/JSON, with no era activating the purpose yet. Refs #1700.

**Validation and scope**
- Every purpose builder and extraneous-redeemer check treats Observe as unmatched, mirroring Guarding; `script.ScriptPurposes` is untouched.
- Only the JSON name-table needed adding; CBOR decoding already accepts any tag value.
- Round-trip tests cover `RedeemerKey`, `RedeemerValue`, and Conway map-form CBOR (asserting decoded datum equality), plus rejection tests for Alonzo, Babbage, Conway, and Dijkstra.

<sup>Written for commit 270164fd8a7538fb332c5122caa8c6f80569eccb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2336?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added support for encoding and decoding the CIP-0112 Observe redeemer tag (`observe`).

* **Bug Fixes**
  * Observe redeemers are now consistently identified as unmatched or extraneous across supported transaction validation paths.
  * Added safeguards to return a validation error instead of attempting to build an unsupported script purpose.

* **Tests**
  * Added coverage for JSON, CBOR, round-trip serialization, and validation behavior.


ledger/common/redeemer_test.go
```diff
@@ -125,6 +130,42 @@ func TestRedeemerKeyRoundTrip(t *testing.T) {
...
+		ExUnits: ExUnits{Memory: 100, Steps: 200},
+	}
+	encoded, err := cbor.Encode(val)
+	require.NoError(t, err)
+
+	var decoded RedeemerValue
+	_, err = cbor.Decode(encoded, &decoded)
+	require.NoError(t, err)
+	assert.Equal(t, val.ExUnits.Memory, decoded.ExUnits.Memory)
+	assert.Equal(t, val.ExUnits.Steps, decoded.ExUnits.Steps)
```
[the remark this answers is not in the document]
cryptodj413: Fixed in 270164fd — both round-trip tests now assert `RedeemerValue.Data` via `PlutusData.Equal`, using a non-zero `data.NewInteger` datum in each. Verified the new assertion actually fails on a mismatched decoded value before reverting to the correct one.

ledger/dijkstra/rules_test.go
```diff
@@ -1289,6 +1289,28 @@ func TestUtxoValidateExtraneousRedeemersGuarding(t *testing.T) {
 	require.ErrorAs(t, err, &conway.ExtraRedeemerError{})
 }
 
+// TestUtxoValidateExtraneousRedeemersObserveTag pins that CIP-0112's Observe
+// purpose (RedeemerTagObserve) has no active script purpose in Dijkstra yet,
+// unlike RedeemerTagGuarding: it stays extraneous until the CIP activates,
+// falling through the same default branch an unrecognized tag would.
```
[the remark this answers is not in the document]
cryptodj413: Fixed in 270164fd — updated the comment to describe RedeemerTagObserve as being rejected by its own explicit case in validateDijkstraExtraneousRedeemers, not by falling through the default branch.

wolf31o2 (approved):

ledger/common/redeemer.go
```diff
@@ -32,6 +32,12 @@ const (
 	RedeemerTagVoting    RedeemerTag = 4
 	RedeemerTagProposing RedeemerTag = 5
 	RedeemerTagGuarding  RedeemerTag = 6
+	// RedeemerTagObserve is CIP-0112's Observe script purpose. No era
+	// currently activates it -- it decodes and round-trips like any other
+	// tag, but every purpose builder and extraneous-redeemer check treats it
+	// as unmatched/extraneous until the CIP lands (expected PlutusV4 under
+	// Dijkstra). See https://cips.cardano.org/cip/CIP-0112.
+	RedeemerTagObserve RedeemerTag = 7
```
chrisguiney, line 40: CIP-0112 assigns no numeric redeemer tag: its only purpose text is `Observe Integer`, and its wire change is transaction body field 14 `required_observers`. cardano-ledger has no Observe purpose either -- `DijkstraPlutusPurpose` encodes 0..6 with 6 = `DijkstraGuarding` and decodes `n -> fail "Unexpected tag for DijkstraPlutusPurpose"`, so tag 7 is refused there. 7 is unallocated and equally available to any other future purpose. Drop the constant until a number is allocated, or record in the doc comment that the value is provisional rather than CIP-0112's.

ledger/common/redeemer.go
```diff
@@ -42,6 +48,7 @@ var redeemerTagNames = map[RedeemerTag]string{
 	RedeemerTagVoting:    "voting",
 	RedeemerTagProposing: "proposing",
 	RedeemerTagGuarding:  "guarding",
+	RedeemerTagObserve:   "observe",
```
chrisguiney, line 51: These two name-table entries are the only behavioral change in the diff. `RedeemerTag.MarshalJSON` previously returned `unknown redeemer tag: 7` for that wire value and now returns `"observe"`; tags 8, 9 and 200 still error, and every one of them already round-tripped through CBOR unchanged, so decoding is not widened -- one previously-unnamed wire value is renamed. Every JSON consumer of redeemers now labels arbitrary tag-7 data with a purpose no era defines, in place of an explicit error.

ledger/common/script/purpose.go
```diff
@@ -409,6 +409,9 @@ func scriptPurposeBuilder(
 			}, nil
 		case lcommon.RedeemerTagGuarding:
 			return nil, UnmatchedRedeemerError{RedeemerKey: redeemerKey}
+		case lcommon.RedeemerTagObserve:
```
chrisguiney, line 412: Adding an enum member reaches consumers that enumerate it. dingo enables `exhaustive` with no settings override, and `api/blockfrost/adapter.go` `redeemerPurpose` switches over `lcommon.RedeemerTag` listing 0..6 plus `default`. That switch shape compiled against this head reports `missing cases in switch of type common.RedeemerTag: common.RedeemerTagObserve`; against v0.205.3 it is clean. So the next gouroboros bump turns dingo lint red in a file the bump does not touch, and the fix there publishes `observe` on a Blockfrost-compatible redeemer purpose field.

ledger/dijkstra/rules.go
```diff
@@ -2854,6 +2854,9 @@ func validateDijkstraExtraneousRedeemers(
 				continue
 			}
 			return conway.ExtraRedeemerError{RedeemerKey: redeemerKey}
+		case common.RedeemerTagObserve:
```
chrisguiney, line 2857: This case and its three siblings (`ledger/common/rules.go:1258`, `ledger/common/script/purpose.go:412` and `:549`) return exactly what `default` already returned. With all four removed `go test ./ledger/...` stays fully green, so the eight new era and CBOR tests pin pre-existing behavior rather than this change; only the JSON name table is causally covered. `exhaustive` is what requires the cases: without them it reports `missing cases in switch of type common.RedeemerTag` at all four sites.

ledger/common/redeemer_test.go
```diff
@@ -125,6 +130,44 @@ func TestRedeemerKeyRoundTrip(t *testing.T) {
...
+	require.NotNil(t, decoded.Data.Data)
+	assert.True(t, val.Data.Data.Equal(decoded.Data.Data))
```
chrisguiney, line 168: Prior review threads checked against this head: the `RedeemerValue.Data` assertion requested on `dec359e0` is addressed here and in `ledger/conway/conway_test.go`, both asserting `PlutusData.Equal` on a non-zero decoded datum; the stale `ledger/dijkstra/rules_test.go` comment is addressed and now matches the code, since `validateDijkstraExtraneousRedeemers` does reject the tag by its own case. Both are resolved, and neither is reopened by the findings above.

chrisguiney (changes requested):
Blocking: RedeemerTagObserve = 7 is not an allocated wire value. CIP-0112 (Observe Script Type, Proposed) assigns no numeric redeemer tag, and cardano-ledger has no Observe purpose at all: DijkstraPlutusPurpose encodes 0..6 with 6 = DijkstraGuarding and its decoder rejects anything above 6. Naming 7 observe commits a public Go constant and a JSON wire name to a number no specification has assigned. Checks at this head predate the current base; the branch merged with current main builds, tests and lints clean. Details inline.

ledger/common/redeemer.go
cryptodj413: Confirmed independently against cardano-ledger's actual `Dijkstra/Scripts.hs`: `DijkstraPlutusPurpose`'s decoder is `6 -> DijkstraGuarding <$> decCBOR; n -> fail $ "Unexpected tag for DijkstraPlutusPurpose: " <> show n` — tag 7 is genuinely unallocated, not CIP-0112's. Agreed this needs to be dropped rather than documented as provisional, since the JSON and dingo consequences below can't be caveated away. Closing this PR without merging; will comment on #1700 to record why and leave it blocked on upstream tag allocation.

ledger/common/redeemer.go
cryptodj413: Agreed — this is the real behavior change, and it's wrong given tag 7 isn't actually assigned to anything. Resolving by dropping the constant entirely (see PR-level comment).

ledger/common/script/purpose.go
cryptodj413: Confirmed against repos/dingo/api/blockfrost/adapter.go directly — redeemerPurpose() switches 0..6 + default, exhaustive enabled with no default-signifies-exhaustive override, exactly as you describe. Good catch; dropping the constant avoids this entirely.

ledger/dijkstra/rules.go
cryptodj413: Correct — these cases are behaviorally identical to default, so the era-rejection tests didn't causally cover anything this diff added, only the JSON name table did. Moot once the constant and its call sites are removed.

ledger/common/redeemer_test.go
cryptodj413: Acknowledged, thanks for confirming these were resolved.

cryptodj413:
Closing without merging, per @chrisguiney's review: `RedeemerTagObserve = 7` is not an allocated wire value. CIP-0112 assigns no numeric redeemer tag (its purpose text is just `Observe Integer`), and I confirmed directly against cardano-ledger's `Dijkstra/Scripts.hs` that the current `DijkstraPlutusPurpose` decoder is:

```haskell
6 -> DijkstraGuarding <$> decCBOR
n -> fail $ "Unexpected tag for DijkstraPlutusPurpose: " <> show n
```

Tag 7 is unallocated and equally available to any future purpose cardano-ledger might add — it isn't reserved for Observe. Shipping it now would:

- have `RedeemerTag.MarshalJSON` silently label arbitrary tag-7 wire data as `"observe"` instead of erroring, per @chrisguiney's finding
- break `exhaustive` lint in dingo's `api/blockfrost/adapter.go` on the next gouroboros bump (confirmed the switch shape there directly), with the likely fix publishing `"observe"` onto a Blockfrost-compatible API field for data no spec defines that way
- add rejection tests that are non-causal (identical behavior to the pre-existing `default` case)

A caveat doc comment doesn't fix the first two, so the constant is being dropped rather than kept as "provisional." Full technical detail and rationale carried over to #1700.

PR #2371, opened 2026-09-17
fix(consensus): verify Byron PBFT signatures over preserved ToSign bytes

### LINKED ISSUE #2349: consensus/byron: verify PBFT block signatures over preserved ToSign wire bytes
## Summary

gouroboros reconstructs Byron PBFT ToSign from decoded Go values and re-encodes them before signature verification. The reference concatenates the original annotated CBOR slices for previous hash, body proof, epoch and slot, difficulty, and extra header data. Accepted non-shortest encodings therefore produce different signed bytes after gouroboros normalization.

## Concrete divergence

A Byron difficulty value encoded as 0x18 0x01 instead of canonical 0x01 decodes to the same value but has different wire bytes. A signature over the reference ToSign containing the original bytes is valid to cardano-ledger, while gouroboros re-encodes the difficulty and verifies against different bytes, rejecting the valid block. Dingo exercises this through ValidatePBFTHeaderCrypto.

## Acceptance criteria

- Recover the five ToSign components directly from HeaderCbor.
- Construct exactly 0x85 followed by raw previous hash, raw body proof, raw epoch-slot, raw difficulty, and raw extra-header bytes.
- Do not normalize integer, byte-string-length, or other accepted CBOR encodings.
- Keep canonical headers valid.
- Keep non-shortest slot and difficulty vectors signed over their exact wire bytes valid.
- Preserve the existing raw proxy-certificate-index handling.

This is consensus-critical Byron validation and a reject-valid-block divergence.

cryptodj413:
## Summary

Fixes #2349. `buildToSignWithEpoch` (`consensus/byron/validate.go`) decoded a header's previous hash, body proof, epoch/slot, difficulty, and extra header data into Go values and re-encoded them to build `ToSign`. CBOR admits non-shortest integer encodings (e.g. difficulty 1 encoded as `0x18 0x01` instead of canonical `0x01`), and gouroboros's decoder accepts them, so a signature over a header using one of those encodings verified against different bytes than the issuer actually signed — cardano-ledger verifies over the original wire bytes, so this rejected a canonical, validly-signed block.

## Change

Recover all five `ToSign` components as raw CBOR sub-slices of the header, the same way `proxyCertIndexCbor` already recovers the delegation certificate index for the same reason, and concatenate them behind a literal `0x85` array-of-5 header instead of decoding and re-encoding:

```
header:        [protocolMagic, prevHash, bodyProof, consensusData, extraHeader]
consensusData: [epochAndSlot, pubKey, difficulty, blockSig]
ToSign := 0x85 ++ prevHash ++ bodyProof ++ epochAndSlot ++ difficulty ++ extraHeader
```

The epoch number `buildToSignWithEpoch` also returns is decoded separately from `epochAndSlot`'s raw bytes purely for the caller's epoch lookup; it does not affect the bytes that get signed.

## Tests

Added `TestValidateSimpleSignaturePreservesNonShortestToSignEncodings`, with two sub-cases (non-shortest difficulty, non-shortest epoch/slot). Both the header and the expected `ToSign` bytes are built from the same raw component literals, so the test cannot derive its expectation from the implementation under test, and each case is verified end to end through a real Ed25519 signature. Confirmed fail-before against the unfixed code. The existing canonical `TestValidateSimpleSignatureReferenceVector` and delegation-certificate tests are unaffected.

## Evidence

- `go build ./...` — pass
- `go vet ./...` — pass

## Test plan

- [x] `go test -race ./consensus/byron/... -run TestValidateSimpleSignature -v`

---
## Summary by cubic
Fixes #2349: Byron PBFT signatures over headers with non-shortest CBOR integer encodings were rejected because `buildToSignWithEpoch` decoded and re-encoded header fields. It now preserves the original wire bytes, so validly signed blocks verify.

**Bug Fixes**
- Builds `ToSign` from raw CBOR sub-slices of the header instead of re-encoding decoded values.
- Decodes the epoch number separately from the raw epoch/slot bytes for lookup only.

**Tests**
- Adds a regression test built from raw component literals and verified with real Ed25519 signatures for non-shortest difficulty and epoch/slot encodings.
- Swaps an `assert`/`FailNow` guard for `require.NotNil` in `ledger/common/poolmetadata_test.go` to fix a nilaway false positive and unblock CI.

<sup>Written for commit e2dd75b839c1fac9f3dae59911dbe333ac7d0e05. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2371?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed block signature validation for blocks containing non-canonical CBOR encodings.
  * Preserved the original wire representation of signed block components, preventing valid signatures from being rejected.

* **Tests**
  * Added coverage for signature validation with non-shortest integer encodings in block header fields.


cryptodj413:
The \`orchestrate / nilaway\` check is red, but it's a pre-existing baseline failure, not caused by this PR. This PR touches only \`consensus/byron/validate.go\` and its test file.

Confirmed: the same \`nilaway\` finding (\`ledger/common/poolmetadata_test.go:200\` / \`common/certs.go:648\`, both files this PR never touches) already fails on \`main\` at this branch's own base commit (89004d98), and on the prior main commit before that too:

\`\`\`
$ gh run list --repo blinklabs-io/gouroboros --branch main --workflow nilaway.yml --limit 3
conclusion=failure  headSha=89004d98...  (this PR's base)
conclusion=failure  headSha=457747dc...
conclusion=failure  headSha=34ec0573...
\`\`\`

Not fixing it here — out of scope for #2349 and unrelated to the ToSign change.

wolf31o2 (approved):

PR #2385, opened 2026-09-18
fix(conway)!: allow any ParameterChange during bootstrap

### LINKED ISSUE #2383: conway: bootstrap ParameterChange field restriction rejects a canonical Preview transaction
`UtxoValidateBootstrapParameterGroups` (`ledger/conway/rules.go`) rejects a `ParameterChange` proposal made during the Conway bootstrap phase (PV9) when it updates any field returned by `ConwayProtocolParameterUpdate.BootstrapRestrictedFields` (`ledger/conway/pparams.go`). cardano-ledger has no equivalent restriction, so this rule rejects transactions that are on the canonical chain and stalls a replay permanently.

## Canonical transaction rejected

Preview tx `2841a581076167a0662f1b4f1a38bcc8eff386f9ce45c33ae33b1fe8289de210`, in block `ff52e64ffd4a6c8296d4570a40d0e709d4be0f7c32d9deb30417dbe2ecfca519`, block height 2570678, epoch 718, absolute slot 62103362.

Its deposit is 100000000000 lovelace, matching Preview's `gov_action_deposit` for that epoch, and `epoch_params` reports `protocol_major` 9 for epochs 716-720. The chain was therefore genuinely in the bootstrap phase when this proposal was accepted.

Validation fails with:

```
conway utxo validation rule 11: ParameterChange proposal updates bootstrap-restricted fields [DRepDeposit] which are not allowed during the Conway bootstrap phase (PV9); requires PV10 (Plomin) or later
```

## Reference behavior

The proposal-time bootstrap gate in `Cardano.Ledger.Conway.Rules.Gov` is `checkBootstrapProposal`, which defers entirely to the action type:

```haskell
isBootstrapAction :: GovAction era -> Bool
isBootstrapAction =
  \case
    ParameterChange {} -> True
    HardForkInitiation {} -> True
    InfoAction -> True
    _ -> False
```

`ParameterChange {}` matches unconditionally, so every `ParameterChange` is permitted during bootstrap whatever fields it updates.

The only other proposal-time check applied to a `ParameterChange` is `actionWellFormed` calling `ppuWellFormed pv ppu` (`Cardano.Ledger.Conway.PParams`). That function mentions `hardforkConwayBootstrapPhase` exactly once, and only to *relax* a constraint:

```haskell
, hardforkConwayBootstrapPhase pv
    || isValid ((/= CompactCoin 0) . unCoinPerByte) ppuCoinsPerUTxOByteL
```

`DRepDeposit` appears there only as a non-zero well-formedness check (`isValid (/= CompactCoin 0) ppuDRepDepositCompactL`). No parameter field is gated on protocol version.

## Impact

A from-genesis Preview replay cannot advance past slot 62103362. The block containing this transaction fails validation on every retry, so the chain never progresses beyond that point.

## Scope of the fix

`UtxoValidateBootstrapAllowedGovActions` in the same file is correct and should stay: its allowed set (InfoAction, HardForkInitiation, ParameterChange) and its rejected set (TreasuryWithdrawal, NoConfidence, UpdateCommittee, NewConstitution) match `isBootstrapAction` exactly.

Only the field-level rule diverges, and it exists in two eras:

- `ledger/conway/rules.go` `UtxoValidateBootstrapParameterGroups`
- `ledger/dijkstra/rules.go` `UtxoValidateBootstrapParameterGroups`, which reuses the same helper via `DijkstraProtocolParameterUpdate.BootstrapRestrictedFields`

Supporting code to remove with them: `ConwayProtocolParameterUpdate.BootstrapRestrictedFields`, `DijkstraProtocolParameterUpdate.BootstrapRestrictedFields`, `BootstrapDisallowedParameterChangeError`, and the `UtxoValidationRuleBootstrapParameterGroups` rule id.

Introduced alongside #1725.

### LINKED ISSUE #2390: ledger/conway: reject protocol-version updates inside ParameterChange
## Summary

gouroboros incorrectly allows Conway ParameterChange updates to carry protocol-parameter key 14, protocolVersion.

The Conway reference does not include protocol version in PParamsUpdate. Protocol changes must use HardForkInitiation, which has distinct ancestry, version-following, DRep, SPO, and committee approval rules.

## Concrete divergence

CBOR update {14: [9, 0]} is decoded by gouroboros, treated as non-empty, and applied by ConwayProtocolParameters.Update. The reference rejects key 14 as an invalid PParamsUpdate field before GOV processing.

The existing test expecting a10e820900 to update protocol version should be inverted.

## Acceptance criteria

- Reject update key 14 during Conway decoding or validation.
- Remove ProtocolVersion from Conway update application and empty-update detection.
- Preserve protocol version in active pparams; only PParamsUpdate may not change it.
- Ensure ParameterChange(protocolVersion) cannot enter proposal state or mutate pparams.
- Preserve HardForkInitiation as the only governance path for protocol-version changes.
- Add valid HardForkInitiation coverage with correct thresholds and hard-fork handling.
- Add PV9 bootstrap and PV10/PV11 rejection regressions.
- Keep ToPlutusData and accepted update fields consistent.
- Apply the same fix to Dijkstra.
- Add rejected CDDL/reference vectors for key 14.
- Keep #2387 supported-field zero constraints separate.

chrisguiney:
`UtxoValidateBootstrapParameterGroups` rejected a `ParameterChange` proposal made during the Conway bootstrap phase (PV9) when it updated any field returned by `BootstrapRestrictedFields`. cardano-ledger has no equivalent restriction, so the rule rejected transactions that are on the canonical chain and stalled a replay at the first one.

Fixes #2383.

## Reference behavior

`checkBootstrapProposal` (`Cardano.Ledger.Conway.Rules.Gov`) defers entirely to the action type:

```haskell
isBootstrapAction :: GovAction era -> Bool
isBootstrapAction =
  \case
    ParameterChange {} -> True
    HardForkInitiation {} -> True
    InfoAction -> True
    _ -> False
```

`ParameterChange {}` matches unconditionally, so any `ParameterChange` is permitted during bootstrap whatever parameters it updates. The arm is identical at `master` and at the PV9-era tag `cardano-ledger-conway-1.16.0.0`.

The only other proposal-time check applied to one is `actionWellFormed` calling `ppuWellFormed` (`Cardano.Ledger.Conway.PParams`). At the PV9-era tag that function takes no `ProtVer` at all, so it cannot gate a field on protocol version. At `master` it takes one and references `hardforkConwayBootstrapPhase` exactly once, only to relax a `CoinsPerUTxOByte` constraint; its sole other version-dependent clause tightens `nOpt` at PV11 and later. `DRepDeposit` appears there only as a non-zero well-formedness check.

## Canonical transaction

Preview tx `2841a581076167a0662f1b4f1a38bcc8eff386f9ce45c33ae33b1fe8289de210`, block height 2570678, epoch 718, absolute slot 62103362, updates `DRepDeposit`. Preview reported `protocol_major` 9 for epochs 716-720 and its `gov_action_deposit` matches the transaction's deposit, so the chain was in the bootstrap phase when the proposal was accepted.

## Change

The rule is removed from Conway and Dijkstra, together with the `bootstrap-parameter-groups` rule id, `BootstrapDisallowedParameterChangeError`, and the `BootstrapRestrictedFields` helpers on `ConwayProtocolParameterUpdate` and `DijkstraProtocolParameterUpdate`, none of which had other callers.

`UtxoValidateBootstrapAllowedGovActions` is unchanged. Its allowed set (InfoAction, HardForkInitiation, ParameterChange) and rejected set (TreasuryWithdrawal, NoConfidence, UpdateCommittee, NewConstitution) already match `isBootstrapAction` exactly.

Removing the exported rule id, error type, and helpers is a breaking API change for consumers that referenced them.

This does not close #2390. `ProtocolVersion` was never in `BootstrapRestrictedFields`, so the rule asked for there is unaffected by this removal.

## Tests

`TestBootstrapRulesAcceptCanonicalPreviewDRepDepositProposal` decodes the transaction above from its own chain bytes, asserts the decoded transaction id, and runs every `bootstrap-` prefixed rule descriptor over it at PV9, so a field restriction reintroduced under a new rule id fails as well. `TestBootstrapPhaseAllowsDijkstraParameterChangeFields` is the Dijkstra equivalent over a `RefScriptCostStride` update, replacing the coverage the removed Dijkstra rule took with it, and also calls `UtxoValidateBootstrapAllowedGovActions` by name.

`TestBootstrapPhaseAllowsAnyParameterChange` covers the Conway rule directly. The tests that asserted the removed restriction are deleted; `TestUtxoValidateBootstrapAllowedGovActions` still pins the action-type rejections that cardano-ledger does define.


ledger/dijkstra/rules_test.go
```diff
@@ -1121,50 +1121,6 @@ func TestUtxoValidateProposalProceduresDijkstraProtocolParameterUpdate(
 	require.NoError(t, UtxoValidateProposalProcedures(tx, 0, nil, nil))
 }
 
-func TestUtxoValidateBootstrapParameterGroupsDijkstraFields(t *testing.T) {
```
wolf31o2, line 1124: The Dijkstra rejection coverage is deleted without replacement acceptance coverage. Add a pre-Plomin `DijkstraParameterChangeGovAction` using a Dijkstra-only field such as `RefScriptCostStride`, and assert `UtxoValidateBootstrapAllowedGovActions` accepts it. The Conway test does not exercise the Dijkstra action type or field set.

wolf31o2 (changes requested):

ledger/conway/rules_bootstrap_paramchange_test.go
```diff
@@ -0,0 +1,116 @@
...
+	if !ok {
+		t.Fatalf(
+			"got governance action %T, want a ParameterChange",
+			govAction,
+		)
+	}
+	// Guards against the fixture silently ceasing to exercise the rule: the
+	// update must still carry DRepDeposit for this to be a regression test.
+	if paramChange.ParamUpdate.DRepDeposit == nil {
+		t.Fatal("transaction fixture no longer updates DRepDeposit")
```
wolf31o2, line 88: This guard protects the field but not the identity of the fixture, and the identity is what the test rests on: the argument is that these exact bytes are on the canonical chain. Edit the hex to any other DRepDeposit proposal and the test still passes while the comment above it becomes false. Assert that tx.Hash().String() equals 2841a581076167a0662f1b4f1a38bcc8eff386f9ce45c33ae33b1fe8289de210 before the field checks; decoding the hex as committed yields exactly that id, so the assertion holds as written.

ledger/dijkstra/rules_test.go
```diff
@@ -1145,24 +1150,62 @@ func TestUtxoValidateBootstrapParameterGroupsDijkstraFields(t *testing.T) {
...
+			descriptor.Validator(tx, 0, nil, pv9Params),
+			"rule %q rejected a bootstrap-phase ParameterChange",
+			descriptor.Id,
+		)
+	}
+	require.NotZero(t, bootstrapRules, "no bootstrap-phase rule descriptors")
+}
+
+// TestUtxoValidateBootstrapAllowedGovActionsAcceptsDijkstraParameterChange
+// pins the specific function named in review on PR #2385: a pre-Plomin
```
wolf31o2, line 1171: This test passes with the rule removal reverted in place, so it guards nothing that TestBootstrapPhaseAllowsDijkstraParameterChangeFields above does not already guard causally, and it repeats that test fixture verbatim. The doc comment also records a review round in permanent source, which the repository comment rules exclude: a comment explains an invariant, not the history of a change. Fold the direct call into the descriptor loop test above, or keep it and drop the reference to the review.

wolf31o2 (changes requested):
Re-review at 081179db. The earlier Dijkstra acceptance gap is closed: TestBootstrapPhaseAllowsDijkstraParameterChangeFields and the two Conway bootstrap tests all fail when the rule removal is reverted in place, and the CodeRabbit finding left at 638f1269 is addressed by the same commits. The removal matches isBootstrapAction at cardano-ledger-conway-1.16.0.0, and Preview tx 2841a581 decodes to the documented id, in epoch 718 at protocol major 9. Two test-side items below.

ledger/dijkstra/rules_test.go
chrisguiney: Addressed. ledger/dijkstra/rules_test.go carries TestBootstrapPhaseAllowsDijkstraParameterChangeFields: a pre-Plomin DijkstraParameterChangeGovAction updating RefScriptCostStride, run over every bootstrap- prefixed rule descriptor, and as of 442ea914 it also calls UtxoValidateBootstrapAllowedGovActions by name so the exported symbol stays referenced. Registering a field restriction under any bootstrap- rule id makes it fail.

ledger/conway/rules_bootstrap_paramchange_test.go
chrisguiney: Fixed in db91742e. previewDRepDepositProposalTxId is asserted against tx.Hash().String() immediately after decode, ahead of the proposal and field checks, so the fixture can no longer be swapped for a different DRepDeposit proposal while the test keeps passing. Changing a single body byte of the hex fails the assertion and reports the decoded id.

ledger/dijkstra/rules_test.go
chrisguiney: Fixed in 442ea914. TestUtxoValidateBootstrapAllowedGovActionsAcceptsDijkstraParameterChange is removed along with its duplicated fixture and the comment recording a review round. Its one distinct guarantee, a compile-time reference to UtxoValidateBootstrapAllowedGovActions, is now a call at the end of TestBootstrapPhaseAllowsDijkstraParameterChangeFields reusing that test's transaction.

wolf31o2 (approved):

PR #2388, opened 2026-09-18
fix(ledger): reject non-empty deprecated Byron block and header attributes

### LINKED ISSUE #2340: ledger: reject non-empty deprecated Byron block and header attributes
## Summary

The Byron reference requires two deprecated attribute fields to be empty, but gouroboros accepts arbitrary values. Main-block ExtraBodyData must be a one-element structure containing empty attributes. The main-header BlockVersions structure also requires empty attributes.

The current main-block decoder stores ExtraBodyData as []any without validating its contents. The header decoder checks the extra-proof field but does not enforce empty BlockVersions attributes. These fields are not part of the signed Byron header fields, so mutating them can produce different accepted encodings with the same Dingo block hash and unchanged proofs.

## Concrete divergence

A valid block [header, body, [empty attributes]] can be changed to [header, body, [non-empty attributes]] without changing the header hash, PBFT signature, previous-block link, or body proofs. The reference rejects the mutated block at the empty-attribute check; gouroboros can decode it. The same issue applies to non-empty header BlockVersions attributes.

## Acceptance criteria

- Require main-block ExtraBodyData to be exactly one element containing empty attributes.
- Reject non-empty maps, zero or multiple elements, and wrong field types.
- Require main-header BlockVersions attributes to be empty.
- Preserve ExtraProof behavior: the reference accepts a byte string of arbitrary length there.
- Mutate only ExtraBodyData on a valid Byron block and require decoding or admission to reject it, while confirming the otherwise-derived block identity is unchanged.
- Add a regression for a validly signed header with non-empty deprecated attributes.
- Prefer enforcement in the gouroboros Byron decoding layer so malformed blocks cannot reach consumers.

This is consensus-critical Byron validation and a wire-malleability issue. Related closed issue #2246 covers other Byron fields whose shape differs from the reference decoder, not these required-empty attribute fields.

cryptodj413:
## Summary

Fixes #2340. The reference requires two deprecated attribute fields to be empty CBOR maps and rejects any other shape at decode time; gouroboros decoded both without validating their contents.

Verified directly against upstream cardano-ledger:

- `Cardano.Chain.Block.Block.hs`'s `decCBORABlock`: the top-level block's `ExtraBodyData` must be exactly one element, an empty map (`enforceSize "ExtraBodyData" 1 >> dropEmptyAttributes`).
- `Cardano.Chain.Block.Header.hs`'s `decCBORBlockVersions`: the header's `BlockVersions` structure must be exactly four elements, with the third (attributes) an empty map (`enforceSize "BlockVersions" 4 ... <* dropEmptyAttributes <* dropBytes`).
- `Cardano.Chain.Common.Attributes.dropEmptyAttributes`: reads the map's length and errors unless it's exactly 0 — accepts any encoding of that length, not just the shortest form.

Neither field is covered by the block hash, PBFT signature, or body proofs the way `ExtraProof` (the header's fourth `BlockVersions` element, still any byte string per `dropBytes`) is, so a block's attributes could be mutated to non-empty and still decode, producing a different accepted encoding for the same otherwise-unchanged block.

## Change

- New `requireEmptyCborMap` helper (`ledger/byron/byron.go`), alongside the existing `requireCborByteString` used for `ExtraProof`.
- `ByronMainBlockHeader.UnmarshalCBOR`: validates the header's `BlockVersions` attributes are empty.
- `ByronMainBlock.UnmarshalCBOR`: rewritten to raw-walk the block (matching the existing header/EBB pattern in this file) and validate the top-level array is exactly 3 elements and `ExtraBodyData` is exactly `[emptyAttributes]`.

## Tests

New `ledger/byron/deprecated_attributes_test.go`, built from the real mainnet Byron block fixture (`internal/testdata.ByronBlockHex`) by splicing raw CBOR sub-messages so only the field under test changes:

- non-empty attributes rejected at both the block and header level (confirmed fail-before against the unfixed code)
- malformed `ExtraBodyData` shapes rejected: zero elements, two elements, wrong element type
- `ExtraProof` still accepts byte strings of arbitrary length (shorter, longer, empty), unaffected by the new check
- the header hash is proven byte-identical before/after an `ExtraBodyData`-only mutation, since it lives outside the header entirely

One existing fixture updated: `TestNewBlockFromCbor_RejectsNullHeaders` (`ledger/verify_block_edge_test.go`) had a synthetic byron-main input with a 0-element `ExtraBodyData`, which the new check now rejects before reaching the null-header check the test actually targets. Given a shape-valid empty-attributes `ExtraBodyData` (`0x81, 0xa0`) instead; the null-header assertion is unchanged.

## Evidence

- `go build ./...` — pass
- `go vet ./...` — pass

## Test plan

- [x] `go test -race ./ledger/byron/... -v`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation when decoding Byron block headers and blocks, rejecting malformed structures and unexpected non-empty attribute data.
  - Added safeguards against invalid or null header content during block decoding.
  - Preserved compatibility with valid Byron data, including proofs of varying lengths.
  - Added regression coverage to ensure malformed inputs are consistently rejected without affecting valid blocks.

---
## Summary by cubic
Fixes #2340 by tightening Byron CBOR decoding. Previously, non-empty or malformed deprecated attribute fields were accepted, including CBOR null/undefined in place of empty maps; now block and header decoding rejects them, while arbitrary byte-string `ExtraProof` values remain valid.

**Bug Fixes**

- Requires a three-field Byron block with `ExtraBodyData` containing exactly one empty CBOR map, rejecting CBOR null/undefined in its place.
- Requires the header's `BlockVersions` attributes to be an empty CBOR map, likewise rejecting null/undefined.
- Fixes `testByronProxyInput` to pass a real empty map instead of a nil one.
- Adds regression coverage using a real Byron block fixture, including verification that changing `ExtraBodyData` does not change the header hash.

<sup>Written for commit 85ee910083f5aef1f3a8c9c6521b504d05433513. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2388?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ledger/byron/byron.go
```diff
@@ -1449,6 +1493,26 @@ func requireCborByteString(raw cbor.RawMessage, field string) error {
...
 
+// requireEmptyCborMap enforces the reference's dropEmptyAttributes check: the
+// value must be a CBOR map, and it must have zero entries. Decoding into
+// map[any]any rejects the type mismatch and reads the length regardless of
+// whether it was encoded in shortest form, matching decodeMapLen's own
+// length-only check.
+func requireEmptyCborMap(raw cbor.RawMessage, field string) error {
+	var m map[any]any
+	if _, err := cbor.Decode(raw, &m); err != nil {
+		return fmt.Errorf("%s must be an empty CBOR map: %w", field, err)
```
wolf31o2, line 1511: This accepts CBOR null (0xf6) and undefined (0xf7): fxamacker decodes both into a nil Go map, and len(nil) is zero. The Byron reference requires an empty CBOR map, so both attribute fields still accept non-map values. Reject non-map CBOR major types before decoding and add regression cases for 0xf6 and 0xf7.

wolf31o2 (changes requested):
The deprecated Byron attribute validation still needs one correction before merge.

ledger/byron/byron.go
cryptodj413: Fixed in 85ee9100 — `requireEmptyCborMap` now checks the raw CBOR major type (`raw[0]&cbor.CborTypeMask != cbor.CborTypeMap`) before decoding, rejecting null/undefined/any other non-map type. Added regression cases for both 0xf6 and 0xf7 at both call sites (block-level ExtraBodyData and header-level BlockVersions attributes). This also surfaced a dependent bug: `testByronProxyInput` in consensus/byron/validate_test.go left `ExtraData.Attributes` at its Go zero value (nil `any`, encodes as CBOR null), which the stricter check now correctly rejects — fixed that test fixture to use a real empty map.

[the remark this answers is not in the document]
cryptodj413: Fixed in 85ee9100 — checked the raw CBOR major type before decoding (matching requireCborByteString's existing pattern), rejecting null (0xf6), undefined (0xf7), and any other non-map type. Added regression cases for both encodings at both call sites.

wolf31o2 (approved):

PR #2399, opened 2026-09-18
fix(ledger): type script_n_of_k threshold as int64 per the CDDL

chadaibbhot-git:
Brings `NativeScriptNofK` in line with the official CDDL.

`conway.cddl` (IntersectMBO/cardano-ledger@7d3ab1b8, `eras/conway/impl/cddl/data/conway.cddl:384-404`):

```
; Note: Allegra switched to int64 for script_n_of_k thresholds.
script_n_of_k = (3, n : int64, [* native_script])
int64 = min_int64 .. max_int64
```

The Haskell evaluator (`eras/allegra/impl/src/Cardano/Ledger/Allegra/Scripts.hs:487-489`) treats `n <= 0` as satisfied: `isValidMOf n SSeq.Empty = n <= 0`.

`NativeScriptNofK.N` was `uint`, so a script with a negative threshold failed to decode, and the error took the whole block with it:

```
decode Conway block error: cbor: cannot unmarshal negative integer into Go struct field conway.tmpConwayBlock.TransactionWitnessSets of type uint
```

Canonical preprod block 5,184,855 (slot 133,902,219, `8ddb2e96…e2f8`) carries `[3, -1, [pubkey, pubkey]]` in its first transaction, so a gouroboros-based node syncing preprod stops at 5,184,854.

### Change
- `N` and the evaluator's counter become `int64`. With that, `count >= N` matches `isValidMOf` for every `n`, including `n <= 0`.
- `internal/bench` helper takes `int64`.
- `N` is an exported field, so callers that build `NativeScriptNofK` from a `uint` need a conversion.

### Tests
- `TestNativeScriptNofKThresholdIsInt64`: decode, byte-identical re-encode and evaluation for n = -1, 0, 1, 2, with no signer and with one of two signers.
- `TestConwayBlockNofKInt64Threshold`: decodes the preprod block, checks the block hash and script hash `678f1ce0…60d4`, and evaluates the script.
- Without the `script.go` change the block test fails with the error above; with it, `go test ./ledger/... ./internal/... ./cbor/...` passes.

---
## Summary by cubic
Types `NativeScriptNofK.N` as `int64` per the Conway CDDL, so native scripts with a negative threshold (like the one in canonical preprod block 5,184,855) now decode and evaluate correctly instead of failing the whole block.

**Pre-Allegra validation**
- Shelley blocks and transactions now reject negative thresholds, since only Allegra and later eras use signed `int64`.
- The evaluator counter is `int64` too, matching the reference behavior for `n <= 0`.

`N` is an exported field, so callers building `NativeScriptNofK` from a `uint` need a conversion. Tests cover decode, re-encode, and evaluation for thresholds -1, 0, 1, 2, plus an end-to-end preprod block check.

<sup>Written for commit 108eb541209870f583d850e677ff6064a568538b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2399?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Native scripts now correctly support `n-of-k` thresholds across the full signed 64-bit range, including zero and negative values.
  - Threshold evaluation is aligned with Conway-era ledger behavior, ensuring thresholds of zero or less are satisfied as expected.
  - Added coverage for encoding, decoding, evaluation, and real-world Preprod script scenarios.


ledger/common/script.go
```diff
@@ -867,10 +867,12 @@ type NativeScriptAny struct {
 	Scripts []NativeScript
 }
 
+// NativeScriptNofK is script_n_of_k. The CDDL types n as int64 (Allegra
+// onwards), and a threshold of zero or less is always satisfied.
 type NativeScriptNofK struct {
 	cbor.StructAsArray
 	Type    uint
-	N       uint
+	N       int64
```
wolf31o2, line 916: `NativeScript.UnmarshalCBOR` is shared by Shelley and later eras and decodes type 3 directly into `int64` without an era-specific check. A negative threshold therefore remains valid in pre-Allegra witness sets, and `Evaluate` treats it as satisfied with zero signatures. Please reject `N < 0` on the pre-Allegra decode path while retaining signed thresholds for Allegra and later eras; otherwise this changes acceptance of malformed Shelley transactions.

wolf31o2 (changes requested):

ledger/common/script.go
chadaibbhot-git: Thanks. I checked the Shelley side before changing anything, and the reference accepts negative thresholds there too, so rejecting `N < 0` pre-Allegra would diverge from it.

`eras/shelley/impl/cddl/data/shelley.cddl` (IntersectMBO/cardano-ledger@7d3ab1b8):

```
; Note: Shelley uses VInt for the threshold in script_n_of_k.
script_n_of_k = (3, n : int32, [* native_script])
int32 = -2147483648 .. 2147483647
```

`eras/shelley/impl/src/Cardano/Ledger/Shelley/Scripts.hs`: the constructor is `MultiSigMOf !Int` (`:103`), decoded with a plain `decCBOR` and no sign check (`:274-277`), and the evaluator is the same as Allegra's (`:305-307`):

```haskell
isValidMOf n StrictSeq.Empty = n <= 0
isValidMOf n (msig StrictSeq.:<| msigs) =
  n <= 0 || if go msig then isValidMOf (n - 1) msigs else isValidMOf n msigs
```

So a Shelley witness set carrying `[3, -1, ...]` decodes and evaluates as satisfied in the reference node, same as Allegra onward. The only era difference in the CDDL is the documented range (int32 in Shelley, int64 from Allegra); the Haskell decoder is `Int` in both. Keeping the shared decoder signed matches that. If you'd rather gouroboros enforce the Shelley int32 range on that path, I can add it, but I don't see the reference rejecting it at decode.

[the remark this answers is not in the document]
chadaibbhot-git: Same point as wolf31o2's comment on this line; answered there with the Shelley CDDL and Haskell decoder citations (both accept a negative threshold pre-Allegra), so no era-specific rejection was added.

wolf31o2 (approved):
*pats self on back*

Issue #2076, opened 2026-08-25
txsubmission: validate reply counts and sizes


wolf31o2:
The TxSubmission server does not match reply counts and transaction bodies to
the outstanding requests. Announced sizes are also not compared with returned
bodies, so protocol state can advance on a mismatched reply.

Acceptance criteria:

- Bound reply counts by the active request.
- Match returned transactions to requested identifiers and announced sizes.
- Reject a mismatched reply without partially advancing state.
- Cover count, order, identifier, and size mismatches.

Issue #2025, opened 2026-08-24
txsubmission: validate state counts and pending-message bounds


wolf31o2:
Tx-submission state validation has three independent wire/state checks in `protocol/txsubmission`:

- `server.go` accepts an empty blocking ReplyTxIds and can busy-loop on it.
- `server.go` reads/writes result channels and ackCount from MsgDone paths without synchronization.
- `client.go` has vacuous ack/request bounds checks and does not compare acknowledgements with outstanding IDs; `txsubmission.go` also sets no per-state pending-message byte limit.

Acceptance criteria:

- Reject empty/oversized messages and acknowledgements not present in outstanding IDs.
- Synchronize shared state and enforce per-state byte bounds.
- Add race-enabled and malformed-message tests.

Issue #2005, opened 2026-08-24
dmq: validate KES periods and bound opcert replay state


wolf31o2:
DMQ authentication state accepts attacker-controlled KES/opcert values without complete bounds and replay checks in `protocol/common/authentication.go`:

- KES period multiplication can overflow and is not cross-checked against the operational certificate.
- Replay validation blocks only counter regression and retains an unbounded opcert cache.

Acceptance criteria:

- Reject overflow, period mismatch, duplicate counters, and values outside the configured retention window.
- Bound cache memory and preserve valid monotonic sequences.
- Add malformed, replay, and eviction tests.

PR #2431, opened 2026-09-19
fix(protocol): bound pending bytes and connection read buffers

wolf31o2:
Node-to-node tx-submission and mux message reassembly had no effective byte bound.

## Tx-submission pending message bytes

`protocol/txsubmission` set `PendingMessageByteLimit: 0` in all six states. `Protocol.SendMessage` gates its outbound queue check on `limit > 0`, and `Protocol.readLoop` gates inbound backpressure and oversized-message rejection on the same condition, so a node-to-node peer had neither.

Every state now carries `MaxPendingMessageBytes` (721,424): `MaxUnackedTxIds` (10) times `TxIdReplyEntryBytes` (44) plus `MaxTxSizeBytes` (65,540), with a 10% safety margin. That reproduces the reference implementation's `txSubmissionProtocolLimits`, whose ingress queue is `addSafetyMargin (maxUnacknowledgedTxIds * (44 + max_TX_SIZE))`, so a conforming peer never exceeds it. `MaxRequestCount` and `MaxAckCount` are `uint16` wire ranges rather than an in-flight window, so they do not size this limit.

Partially addresses #2025, the per-state byte bounds and the ack-versus-outstanding check only. The empty blocking ReplyTxIds and MsgDone synchronization are unchanged, as is the reply correlation tracked in #2076.

## Tx-submission request bounds

A byte limit alone turns an oversized request into a connection teardown. `MsgRequestTxIds` carries `uint16` counts and both request paths gated them on `MaxRequestCount` (65,535), the entire range, so a peer asking for 65,535 transaction IDs drew a `MsgReplyTxIds` encoding to 2,621,404 bytes against the new 721,424-byte limit. `Protocol.enqueueMessage` then returned `ErrProtocolViolationQueueExceeded` and called `SendError`, failing the protocol and dropping the connection over our own limit. `MsgRequestTxs` was unbounded the same way.

Both requests are now bounded by the outstanding window, reproducing the reference implementation's condition in `Ouroboros.Network.TxSubmission.Outbound`: `ProtocolErrorAckedTooManyTxids` when a peer acknowledges more IDs than it was sent, and `ProtocolErrorRequestedTooManyTxids` when `unackedNo - ackNo + reqNo` exceeds `maxUnacked`, with `maxUnacked` = `MaxUnackedTxIds` (10). The client tracks the IDs it has sent unacknowledged and the server the IDs it is holding, so the bound is the window rather than a per-message cap; both counters reset when the protocol restarts. `Server.RequestTxs` and the client's body-request handler reject more than `MaxUnackedTxIds` identifiers on the same grounds.

The reply is bounded as well as the request. `RequestTxIdsFunc` and `RequestTxsFunc` are caller-supplied, and a callback returning more than was requested puts a message on the wire that a conforming peer rejects and that `enqueueMessage` can refuse against our own pending-byte limit. Both handlers refuse it before constructing the reply. The reference implementation clamps its own txid read with `take reqNo` and answers a body request only from the identifiers it was asked for, so there is no correct truncation of a larger callback result here: it announces a set the caller did not record, or bodies that were never requested.

`MaxRequestCount` and `MaxAckCount` remain the wire ranges of the count fields; `Server.RequestTxIds` keeps its range check ahead of the window check because it bounds `reqCount` before it is added to a counter and licenses the `uint16` conversion. `DefaultRequestLimit` (1,000) exceeds `MaxUnackedTxIds`, so a caller using it as a request count is refused. It is misleading as guidance, and its doc comment, `PROTOCOL_LIMITS.md` and the protocol README now say so rather than the bound being relaxed to accommodate it. A conforming peer requests at most `maxNumTxIdsToRequest` (6) at a time, well inside this window.

## Connection-wide read buffer allowance

`connection.go` starts eight node-to-node mini-protocols in both roles, so a full-duplex connection runs sixteen `readLoop`s, each free to grow its own reassembly buffer to `MaxReadBufferSize` (16 MB) with nothing accounting for the total: 256 MiB per connection.

The muxer now carries a connection-wide reassembly allowance. Each protocol raises it to its own cap as it registers, so the allowance is the largest registered cap rather than the sum. A `readLoop` checks its own `MaxReadBufferSize` before reserving, so a message larger than the per-protocol cap is refused on those terms; a buffer that would take the connection past the allowance fails its protocol with `connection read buffer budget exhausted` instead; the allowance is returned as messages are consumed and when a read loop exits.

Because the allowance is the largest cap and not the sum, two mini-protocols reassembling near their caps at once on one connection cannot both proceed. The largest message a node-to-node mini-protocol admits is bounded by its own pending-message byte limit, the largest of those being Block Fetch's 2,500,000 bytes, so 16 MB covers every legitimate node-to-node reassembly. A node-to-client caller that needs more headroom raises `MaxReadBufferSize` on the protocol that needs it, which raises the connection allowance with it.

The allowance is per connection; nothing aggregates across connections.

Partially addresses #2073, the aggregate connection budget only. The node-to-client ChainSync pending-message budget is unchanged; its zero limit is correct per specification section 3.7.5.

Still carrying no pending-message byte limit, each needing its own derivation: the `keepalive`, `peersharing`, `handshake`, `leiosfetch`, `leiosnotify` and `leiosvotes` state maps, and DMQ `messagesubmission`. Each can still park up to the 16 MB per-protocol read-buffer cap. The connection total is now 16 MB rather than 256 MB, but those states remain individually unbounded.

## Opcert cache contract

The KES opcert cache comment in `protocol/common/authentication.go` advised callers to evict entries to bound growth. `verifyKESPeriodRotation` runs after stake authorization and both signature checks, so an entry requires a pool holding active stake and possession of its cold and KES keys: the key space is the stake distribution, not peer-chosen input. Evicting an entry restores that pool to never-seen and re-opens stale-opcert replay. The comments now state that, and a test pins the check ordering the bound rests on.

This does not address #2005, which asks for KES period overflow and mismatch rejection and a retention window.


protocol/protocol.go
```diff
@@ -819,6 +874,14 @@ func (p *Protocol) readLoop() {
 				if !ok {
 					return
 				}
+				// Reserve before the write. bytes.Buffer.Write grows and
+				// copies, so writing first takes the allocation the
+				// connection allowance exists to refuse.
+				pendingLen := readBuffer.Len() + len(segment.Payload)
+				if !p.reserveReadBuffer(pendingLen, &reserved) {
```
arepala-uml: The reserve refuses at the same threshold the per-protocol cap uses, because the connection allowance is the largest registered cap, so readBuffer.Len() > maxReadBufferSize() at line 985 is unreachable whenever a Muxer is attached and this protocol holds that cap. Measured with one protocol on a real muxer at cap 4096, a 4296-byte message reports 'connection read buffer budget exhausted (connection limit 4096 bytes)' and blames contention that does not exist. Check the per-protocol cap before reserving so the existing error still decides 'too big'.

protocol/protocol_test.go
```diff
@@ -668,3 +668,203 @@ func TestReadLoopDrainRespectsMaxBufferSize(t *testing.T) {
...
 	}
 }
+
+// budgetTestProtocol is a Protocol wired to a real Muxer for its
+// connection-wide reassembly allowance, but fed from a pre-loaded receive
+// channel rather than through the muxer's own read loop. Pre-loading removes
+// the scheduling variable the same way TestReadLoopDrainRespectsMaxBufferSize
+// does: every segment is queued before readLoop starts, so the buffer growth
+// under test does not depend on which goroutine wins a quantum.
+type budgetTestProtocol struct {
```
arepala-uml, line 687: TestReadLoopDrainRespectsMaxBufferSize, the CWE-400 guard from #2291, builds a Protocol with a nil Muxer, so reserveReadBuffer short-circuits and it no longer exercises the path production takes. Add a muxer-attached case, or the per-protocol read-buffer cap is left without a regression test.

protocol/txsubmission/client.go
```diff
@@ -332,6 +364,13 @@ func (c *Client) handleRequestTxIds(msg protocol.Message) error {
 	if err := c.SendMessage(resp); err != nil {
 		return err
 	}
+	// Record what the peer now holds unacknowledged. A callback that returns
+	// more IDs than were requested over-subscribes the window, which then
+	// refuses the peer's next request; a peer sees the same overrun in
+	// Server.RequestTxIds.
+	c.unackedMu.Lock()
+	c.unackedTxIds = unacked - ack + len(txIds)
```
arepala-uml, line 380: Nothing bounds len(txIds) against msgRequestTxIds.Req before the reply is sent, so an over-returning RequestTxIdsFunc puts a larger reply than requested on the wire and leaves this counter above MaxUnackedTxIds. Measured: a callback returning 10 ids to a req=1 request sends with no error, then the peer's next conforming ack=0 req=1 request is refused with ErrProtocolViolationRequestExceeded and the connection is dropped. Server.RequestTxIds rejects the mirror case at server.go:193; apply the same check before NewMsgReplyTxIds.

protocol/txsubmission/client.go
```diff
@@ -349,6 +388,19 @@ func (c *Client) handleRequestTxs(msg protocol.Message) error {
...
+	// it here rather than let the callback materialize every body first and
+	// have SendMessage reject the result as a violation of our own.
+	if len(msgRequestTxs.TxIds) > MaxUnackedTxIds {
+		c.Protocol.Logger().
+			Error("TxSubmission tx request count exceeded",
+				"requested", len(msgRequestTxs.TxIds),
+				"limit", MaxUnackedTxIds,
+			)
+		return protocol.ErrProtocolViolationRequestExceeded
+	}
```
arepala-uml, line 411: The incoming request is bounded here but the reply is not: NewMsgReplyTxs at line 409 sends whatever RequestTxsFunc returns. Measured with a conforming 10-id request and a callback returning 20 maximum-size bodies, enqueueMessage fails the protocol with ErrProtocolViolationQueueExceeded, the exact failure this change exists to prevent. The review thread asking for the window to be enforced before constructing or sending the reply is marked addressed but covers only the request half.

arepala-uml (changes requested):
Two reply paths put a message on the wire that our own pending-byte limit then refuses, dropping the connection on a conforming request; both are measured and both mirror a guard Server.RequestTxIds already has. Separately, the connection allowance makes the per-protocol read-buffer error unreachable and leaves its regression test covering only a nil-Muxer path. The tx-submission byte limits, the request bounds and the opcert cache correction are right, and dingo builds and passes against this head.

protocol/protocol.go
wolf31o2: Fixed in 10e511ac. Both receive paths now go through `Protocol.appendSegment`, which checks `p.config.maxReadBufferSize()` first, then reserves the connection allowance, then writes. A 4296-byte message reaching a lone protocol at cap 4096 now reports `read buffer exceeded maximum size (4296 bytes)`; `TestReadLoopRejectsAnOversizedMessageOnTheProtocolCap` fails with the budget-exhausted error if the two checks are swapped back.

With the cap enforced before `bytes.Buffer.Write`, `readBuffer` can no longer hold more than the cap, so the post-decode check is unreachable and is removed rather than left as a second decider.

protocol/protocol_test.go
wolf31o2: Addressed in 590c1b0e: `TestReadLoopDrainRespectsMaxBufferSize` builds its `Protocol` with a real `Muxer` and raises the connection allowance to the same cap, so `reserveReadBuffer` no longer short-circuits. With the muxer attached, dropping the drain-loop bound fails it at 999805 bytes against its 456-byte ceiling.

10e511ac adds the case the drain test cannot reach, `TestReadLoopRejectsAnOversizedMessageOnTheProtocolCap`: a single oversized segment on the non-drain receive path, also muxer-attached.

protocol/txsubmission/client.go
wolf31o2: Fixed in 590c1b0e: `handleRequestTxIds` refuses when `len(txIds)` exceeds `msgRequestTxIds.Req`, before `NewMsgReplyTxIds` and before `unackedTxIds` is advanced.

5ec710ae adds `TestRequestTxIdsRefusesAnOverReturningCallback`, which answers a `req=1` request with 10 IDs and asserts `ErrProtocolViolationRequestExceeded` with the window still at zero; without the check it times out with the reply on the wire.

Refusing rather than clamping: the reference clamps with `take reqNo` because it reads its own mempool, while these IDs come from a caller-supplied callback. Truncating would announce a different set than the caller recorded, and its later acknowledgements would not match.

protocol/txsubmission/client.go
wolf31o2: Fixed in 590c1b0e: `handleRequestTxs` refuses when `len(txs)` exceeds `len(msgRequestTxs.TxIds)`, before `NewMsgReplyTxs`.

5ec710ae adds `TestRequestTxsRefusesAnOverReturningCallback` with the measured case, a 10-id request answered with 20 maximum-size bodies; removing the check reproduces `ErrProtocolViolationQueueExceeded` exactly. It is paired with a full window of maximum-size bodies, which must still be sent.

Clamping is not available on this path: the reference answers only from the IDs it was asked for and throws `ProtocolErrorRequestedUnavailableTx` otherwise, so bodies past the request were never requested and no truncation of them is correct.

protocol/protocol_test.go
```diff
@@ -668,3 +668,196 @@ func TestReadLoopDrainRespectsMaxBufferSize(t *testing.T) {
...
+		)
+		go p.proto.readLoop()
+		go func() {
+			select {
+			case err := <-p.errorChan:
+				refused <- err
+			case <-time.After(10 * time.Second):
+			}
+		}()
+	}
```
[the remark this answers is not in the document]
wolf31o2: Addressed. 10e511ac removes the watcher goroutines rather than giving them a stop case: `SendError` stops the protocol immediately after delivering to `ErrorChan`, so a select over the error and `stopChan` can take the stop case and lose a refusal that was already reported. Both error channels are read from the one select in the test body instead, and each protocol keeps its `t.Cleanup(Stop)`.

protocol/protocol.go
```diff
@@ -821,6 +876,10 @@ func (p *Protocol) readLoop() {
 				}
 				// Add segment payload to buffer
 				readBuffer.Write(segment.Payload)
+				if !p.reserveReadBuffer(readBuffer.Len(), &reserved) {
+					p.SendError(p.errReadBufferBudget(readBuffer.Len()))
+					return
+				}
 			}
 			// Opportunistically drain any additional segments the muxer
 			// has already queued for us before spending a decode attempt.
```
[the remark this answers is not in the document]
wolf31o2: Addressed. `Protocol.appendSegment` computes `readBuffer.Len()` plus the segment length, reserves that total, and only then writes the payload, on both receive paths. 10e511ac also moves the per-protocol cap check ahead of the reservation, so an oversized message is refused on its own terms rather than reported as connection contention.

protocol/txsubmission/txsubmission.go
```diff
@@ -144,6 +144,42 @@ const (
...
+	// serve as the multiplier here.
+	MaxUnackedTxIds = 10
+	// MaxPendingMessageBytes bounds pending message bytes in every
+	// TxSubmission state: a full unacknowledged window of maximum-size
+	// transactions plus the tx-id reply that announced them, with the
+	// reference implementation's 10% safety margin. This is the value
+	// cardano-node enforces as its own tx-submission mux ingress limit, so a
+	// conforming peer never exceeds it.
+	MaxPendingMessageBytes = MaxUnackedTxIds *
+		(TxIdReplyEntryBytes + MaxTxSizeBytes) * 11 / 10
```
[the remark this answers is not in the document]
wolf31o2: Addressed on both halves. The request half bounds `len(msgRequestTxs.TxIds)` against `MaxUnackedTxIds` before the callback runs, and `Server.RequestTxs` applies the same bound to what it sends. The reply half refuses a callback returning more bodies than were requested, before `NewMsgReplyTxs`.

5ec710ae covers all three, including the case that reaches `ErrProtocolViolationQueueExceeded` without the refusal.

arepala-uml (approved):

Issue #2368, opened 2026-09-17
ledger/byron: separate wire-body hash from produced-UTxO transaction ID


wolf31o2:
## Summary

Byron uses original annotated transaction bytes for witness signing data and transaction Merkle proofs, but produced UTxO identity comes from the hash of the decoded transaction re-encoded by the reference serializer.

gOuroboros uses the preserved raw body for both purposes. A valid non-shortest integer encoding can therefore create a different produced UTxO identifier and split same-block spends.

## Acceptance criteria

- Represent wire-body hash and reference transaction ID as separate concepts.
- Produced UTxOs must use the hash of the decoded transaction re-encoded by the reference serializer.
- Keep transaction proof and witness-signing paths on original annotated bytes.
- Canonical transactions must produce equal hashes in both paths.
- Add a non-shortest integer vector where the two hashes differ.
- Test same-block spending by the reference transaction ID and by the raw-wire ID.
- Ensure persisted Byron UTxO identifiers match cardano-ledger.

Issue #2012, opened 2026-08-24
ledger: preserve Shelley mandatory fields and witness rules


wolf31o2:
Shelley-era validation differs from the reference in several closely related transaction-body rules:

- `ledger/shelley/rules.go` treats `ttl == 0` as absent and omits mandatory key 3.
- The same rules do not require script witnesses for script-locked inputs or mint policies.
- `ledger/shelley/shelley.go` decodes opcert sequence/KES values as uint32 instead of wire uint64.

Acceptance criteria:

- Preserve present zero TTL and emit/require mandatory key 3.
- Require the matching script witness for every script input and mint policy.
- Preserve full uint64 opcert values and reject values outside supported operational bounds.

PR #2435, opened 2026-09-19
fix: Byron merkle root, Shelley ttl, Byron block type bound

### LINKED ISSUE #2039: chainsync: support Byron block types in RollForward
The node-to-node chainsync server's `RollForward` path converts a block type
through `ledger.BlockToBlockHeaderTypeMap`. Byron block types are not present
in that map, so a valid Byron roll-forward cannot be converted through the
normal server path.

Acceptance criteria:

- Map every supported Byron block type to its corresponding header type.
- Return an explicit error for an unknown block type instead of using a zero
  map value.
- Add node-to-node server tests for Byron EBB and regular blocks plus an
  unknown-type rejection case.

### LINKED ISSUE #2369: consensus/byron: match reference transaction Merkle-root construction
## Summary

consensus/byron.ValidateBodyHash reconstructs the Byron transaction-body Merkle root with duplicate-last padding. The Byron reference uses a left-heavy RFC6962-style tree that recursively splits at the largest power of two strictly below the item count. ledger/byron.MerkleRoot already implements the reference algorithm correctly.

For non-power-of-two transaction counts, the exported consensus validator therefore computes a different root and can reject a reference-valid block.

## Concrete divergence

For transactions A, B, C, the reference root is:

branch(branch(leaf(A), leaf(B)), leaf(C))

The consensus/byron implementation instead duplicates C and computes:

branch(branch(leaf(A), leaf(B)), branch(leaf(C), leaf(C)))

A valid three-transaction block can therefore pass the normal ledger body-proof path but fail when validated through consensus/byron.ValidateBodyHash.

## Acceptance criteria

- Remove duplicate-last padding from consensus/byron.ValidateBodyHash.
- Reuse ledger/byron.MerkleRoot or a shared implementation.
- Preserve the reference rules: empty root is Blake2b-256 of empty bytes; leaves hash 00 plus original transaction-body CBOR; branches hash 01 plus child roots; split at the largest power of two strictly below the item count.
- Preserve original transaction-body CBOR for leaf hashing.
- Test counts 0 through at least 8, explicitly including 3, 5, 6, and 7.
- Assert consensus/byron reconstruction equals ledger/byron.MerkleRoot.
- Add a complete three-transaction ValidateBodyHash regression and a genuinely incorrect-root negative test.
- Do not change witness-list hashing.

## Reachability, verified 2026-09-19 at 0370c53d

`ValidateBodyHash` has no production callers. It appears only in its own
definition, a documentation cross-reference, and tests. Dingo imports
`consensus/byron` solely for `ValidatePBFTHeader`, `ValidatePBFTHeaderCrypto`
and `NewPBFTState`. The live decode path is
`ledger/byron.NewByronMainBlockFromCbor` -> `ValidateBodyProof` ->
`ledger/byron.MerkleRoot`, which implements the reference correctly.

Byron sync is therefore unaffected today. This is an exported-API and
dead-code defect rather than an active consensus break, which lowers its
urgency but not its correctness.

Confirmed against chain data rather than by inspection: mainnet Byron main
block `7061f3fe3fe335b569f91b13e0a2fadfbd4f052f62ec7db9743e2838c8e4c30e`
(slot 4341573, 3 transactions) carries on-chain txProof root
`334a7b973b88e88f4c29d3bc55572995c64d990bc2df2d1748c29648e6491452`.
`ledger/byron.MerkleRoot` and a transcription of `Merkle.hs` both reproduce it;
`consensus/byron.computeMerkleRoot` returns
`e3b67b71...dc97a4` and `ValidateBodyHash` rejects that genuine block.
Reproduced on a 5-transaction block; a 4-transaction block agrees, because
power-of-two counts are indistinguishable.

## An existing test pins the defect

`consensus/byron/validate_test.go:1617` `TestComputeMerkleRoot_ThreeItems`
hand-builds the duplicate-last shape as its expected value, commented
"Level 1: combine (leaf1, leaf2), (leaf3, leaf3)". It therefore asserts the
defect rather than detecting it, and explains why the bundled fixtures never
caught this. That test must be made to fail before any fix is believed.

Given the above, deleting `computeMerkleRoot` and routing `ValidateBodyHash`
to `ledger/byron.MerkleRoot` may be preferable to repairing a second
implementation.

wolf31o2:
Three independent defects, one per commit.

### Byron transaction merkle root (#2369)

`consensus/byron` computed the transaction-body merkle root by padding each
level to an even width with a duplicate of the last node. cardano-ledger's
`Cardano.Chain.Common.Merkle` instead splits recursively at the largest power
of two strictly below the item count (`mkMerkleTree'` / `powerOfTwo`), which
differs for every count that is not a power of two.

`ledger/byron.MerkleRoot` already implements the reference construction and is
the live decode path, so `validateTxProof` routes through it and the duplicate
implementation is deleted.

Mainnet block `7061f3fe3fe335b569f91b13e0a2fadfbd4f052f62ec7db9743e2838c8e4c30e`
carries three transactions and on-chain txProof root
`334a7b973b88e88f4c29d3bc55572995c64d990bc2df2d1748c29648e6491452`;
`ValidateBodyHash` rejected it before this change.

The four `TestComputeMerkleRoot_*` tests were removed with the function.
`_ThreeItems` asserted the duplicate-last shape as its expected value, so it
pinned the defect rather than detecting it. The replacement decodes that
mainnet block, reproduces the on-chain root, and carries a wrong-root negative
control; `ledger/byron/merkle_test.go` is extended to counts 0 through 8.

Does not close #2369's sibling #2368, which covers the produced-UTxO
transaction ID rather than the wire-body hash.

### Shelley ttl as mandatory key 3 (#2012)

`ShelleyTransactionBody` carried `cbor:"3,keyasint,omitempty"` and `TTL()`
returned the zero value, so an absent ttl and a present zero ttl were
indistinguishable, and a re-encode dropped key 3 entirely.

cardano-ledger decodes the Shelley body with `SparseKeyed "TxBody"
basicShelleyTxBodyRaw boxBody [(0,"inputs"),(1,"outputs"),(2,"fee"),(3,"ttl")]`,
whose final argument is the required-field list, and `txSparse` emits
`Key 3 (To ttl)` with no `Omit` wrapper. `validateTimeToLive`
(`Cardano.Ledger.Shelley.Rules.Utxo`) is `failureUnless (ttl >= slot)` with no
exemption for zero.

- `omitempty` dropped from key 3; a body that omits key 3 is rejected at decode.
- Presence recorded through the existing
  `DecodeValidityIntervalUpperBoundPresence`, with a new
  `ValidityIntervalUpperBound()` following the Allegra-onward shape.
- `UtxoValidateTimeToLive` consults presence instead of `ttl != 0`.

Compatibility: this is the behavioral break in the change. A Shelley body
without key 3 no longer decodes, and a hand-built `ShelleyTransactionBody`
re-encodes with `3: 0` where it previously omitted the key. Synthetic Shelley
bodies in `ledger`, `ledger/common` and `ledger/shelley` tests gained `3: 0`;
every assertion is unchanged, because those fixtures were incomplete. Only
Shelley is affected: key 3 is optional from Allegra on, no other era embeds
this body, and `shelley.UtxoValidationRules` is selected only for the Shelley
era in `ledger/verify_block.go`.

Recording presence without enforcing the field would be zero-churn and is a
defensible alternative reading, weighting the API break above matching the
reference decoder. Enforcement is what the reference does, and #2012 asks for
the mandatory treatment; #2012 retains other items and is not closed here.

### Byron block type from peer header metadata

`WrappedHeader` took `byronType` straight from a peer's wrapped Byron header
metadata and used it as a block type. With `firstBlockChan == nil` and
`RollForwardFunc == nil`, `protocol/chainsync/client.go` passes that value
directly to `RollForwardRawFunc` with no header decode.

The legal domain is 0 (boundary) and 1 (regular): `decCBORABlockOrBoundaryHdr`
(`Cardano.Chain.Block.Block`) fails with "Unknown tag in encoded
HeaderOrBoundary" for anything else, and ouroboros-consensus' `DecodeDiskDepIx`
instance for `(NestedCtxt Header) ByronBlock`, which encodes this
`[type, size]` pair on the wire, rejects the same values with
`DecoderErrorUnknownTag`.

Both `WrappedHeader.UnmarshalCBOR` and `NewWrappedHeader` now reject
out-of-domain values; the constructor assigned the field without decoding.
This does not close #2039, which covers the node-to-node server's
`BlockToBlockHeaderTypeMap` conversion.

---
## Summary by cubic
Addresses three independent ledger defects: the Byron merkle root computation (#2369), mandatory Shelley ttl (#2012), and Byron block type bounds.

**Bug Fixes**
- `consensus/byron` now uses `ledger/byron.MerkleRoot`; the previous duplicate-last implementation rejected valid mainnet blocks with non-power-of-two transaction counts. A regression test decodes a real three-transaction mainnet block and includes a wrong-root negative control.
- Shelley transaction bodies now require key 3, and the ttl validation rule treats a present zero ttl as expired.
- `protocol/chainsync` now rejects Byron block types outside 0 (EBB) and 1 (main) from peer header metadata instead of passing them through unchecked.

**Compatibility**
- Hand-built Shelley bodies now encode `3: 0` instead of omitting key 3; key 3 remains optional from Allegra onward.

<sup>Written for commit f41a3b9f3b626ad18f1ec974ec5f01e108275a74. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2435?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ledger/shelley/shelley.go
```diff
@@ -381,10 +384,34 @@ func (b *ShelleyTransactionBody) UnmarshalCBOR(cborData []byte) error {
...
 	return nil
 }
 
+// ValidityIntervalUpperBound returns the TTL and whether it is present. Key 3
+// is mandatory in Shelley, so every decoded body reports it as present,
+// including when its value is zero. A body constructed in Go rather than
+// decoded reports presence only once a nonzero TTL or an explicit
+// SetValidityIntervalUpperBoundPresence has set it.
+func (b *ShelleyTransactionBody) ValidityIntervalUpperBound() (uint64, bool) {
+	return b.Ttl, b.Ttl != 0 || b.ValidityIntervalUpperBoundPresent()
```
arepala-uml: Key 3 is now mandatory and always encoded, so the Allegra-onward fallback copied here makes a Go-built body disagree with its own serialization. Measured with Ttl unset: in memory presence is false and UtxoValidateTimeToLive accepts at slot 100, while the same body encoded (key 3 emitted as 0) and decoded reports presence true and is rejected as expired. Before this change omitempty dropped key 3 and both agreed. Returning b.Ttl, true matches the encoder, and flips ledger/witness_count_test.go:36,40,46, ledger/validity_interval_test.go:198 and zero-valued Shelley bodies in dingo.

ledger/shelley/shelley.go
```diff
@@ -381,10 +384,34 @@ func (b *ShelleyTransactionBody) UnmarshalCBOR(cborData []byte) error {
...
+	// distinguishable from the absent field a zero value would otherwise
+	// look like.
+	if err := b.DecodeValidityIntervalUpperBoundPresence(cborData, b.Ttl); err != nil {
+		return err
+	}
+	if !b.ValidityIntervalUpperBoundPresent() {
+		return errors.New(
+			"shelley transaction body is missing mandatory ttl (key 3)",
+		)
+	}
```
arepala-uml, line 401: The SparseKeyed required-field list cited as the authority names inputs, outputs, fee and ttl, but only key 3 is enforced here: a body of just key 3 decodes, as do bodies missing key 0, key 1 or key 2. Pre-existing and partly masked downstream by UtxoValidateInputSetEmptyUtxo and UtxoValidateFeeTooSmallUtxo, so not a blocker. Worth naming because the new fixtures in ledger/witness_count_test.go:322 and ledger/common/duplicate_collections_integration_test.go encode bodies the cited reference still could not decode.

ledger/shelley/shelley.go
```diff
@@ -381,10 +384,34 @@ func (b *ShelleyTransactionBody) UnmarshalCBOR(cborData []byte) error {
...
 	b.SetCbor(cborData)
 	return nil
 }
 
+// ValidityIntervalUpperBound returns the TTL and whether it is present. Key 3
+// is mandatory in Shelley, so every decoded body reports it as present,
+// including when its value is zero. A body constructed in Go rather than
+// decoded reports presence only once a nonzero TTL or an explicit
```
arepala-uml: This records the constructed-body divergence rather than resolving it, and justifies it by analogy to eras where key 3 is not mandatory and the analogy holds. Separately, TestTransactionValidityIntervalUpperBoundLegacyFallback at ledger/validity_interval_test.go:197 used ShelleyTransactionBody precisely because Shelley lacked this method. It no longer reaches the fallback branch at ledger/common/tx.go:97 and passes only because this arm reproduces the legacy answer. Rename it or repoint it at a type that still lacks the method.

ledger/shelley/rules.go
```diff
@@ -126,15 +126,23 @@ var UtxoValidationRules = common.MustUtxoValidationRulesFromDescriptors(
...
 func UtxoValidateTimeToLive(
 	tx common.Transaction,
 	slot uint64,
 	ls common.LedgerState,
 	pp common.ProtocolParameters,
 ) error {
-	ttl := tx.TTL()
-	if ttl == 0 || ttl >= slot {
+	ttl, present := common.TransactionValidityIntervalUpperBound(tx)
+	if !present || ttl >= slot {
```
arepala-uml, line 145: Consulting presence rather than the value matches the reference, but it also changes a second consumer the description does not mention. For a Shelley body carrying an explicit zero key 3, presence flips to true, so validityEnd in ledger/common/script.go:875 flips from MaxUint64 to 0 and a NativeScriptInvalidHereafter term flips from fail to pass. Shelley does run UtxoValidateNativeScripts. It is masked at every slot above zero because this rule rejects the body first, leaving only slot 0 exposed.

arepala-uml (changes requested):
Reviewed the four branch commits at f41a3b9f against the merge base, then with current main merged locally: build, vet, gofmt, the full test sweep, -race on the affected packages, conformance and golangci-lint all pass on the merged tree. The Byron merkle root and Byron block type commits are sound and both finding classes are complete; the deleted TestComputeMerkleRoot_ThreeItems did assert the duplicate-last shape, and no coverage was lost. The Shelley ttl commit makes encoding and validation disagree about a body built in Go, which is the blocker. Three non-blocking items are inline.

wolf31o2:
Addressed the Shelley TTL round-trip blocker in the pushed commit. Go-constructed Shelley bodies now report key 3 as present, matching the encoder, including zero values; the prior fallback test now explicitly checks Shelley’s mandatory-field behavior. Focused validation passed: GOWORK=off go test ./ledger ./ledger/shelley ./ledger/common.

arepala-uml (approved):

Issue #1881, opened 2026-07-15
dijkstra: type nested-tx account fields (required_top_level_guards, account_balance_intervals)


wolf31o2:
## Scope item
CIP-118 Nested Transactions / CIP-159 Account Address Phase 1. Dijkstra (PV12) hard-fork readiness.

## Gap
Three transaction-body fields are stored as opaque `DijkstraRawCbor` rather than typed structures, so they cannot be validated or surfaced to scripts:
- `TxBalanceIntervals` (main body key 26)
- `TxRequiredTopLevelGuards` (sub-tx key 24)
- `TxAccountBalanceIntervals` (sub-tx key 26)

The pinned CDDL defines `required_top_level_guards = {+ credential => plutus_data/nil}` and `account_balance_intervals = {+ credential => account_balance_interval}`.

## Work
Add typed decode/encode for these fields so they can be validated (see the sub-transaction validation issue) and exposed in the PlutusV4 script context.

## Evidence
- `ledger/dijkstra/dijkstra.go:755, 991, 993` (opaque fields)
- `ledger/dijkstra/testdata/dijkstra.cddl:805-819` (typed CDDL definitions)

Issue #1880, opened 2026-07-15
dijkstra: validate sub-transactions (CIP-118)


wolf31o2:
## Scope item
CIP-118 Nested Transactions — the primary Dijkstra (PV12) feature. Part of the Dijkstra hard-fork readiness effort (tracked in the blinklabs-io/dingo epic).

## Gap
The CBOR codec for `sub_transactions` (body key 23) exists, but no validation rule checks a sub-transaction's internal consistency. `UtxoValidationRules` (`ledger/dijkstra/rules.go:39-86`) operates only on the top-level transaction; the only sub-transaction traversals gather available Plutus scripts / used script versions / native scripts (`rules.go:418, 790, 881, 1476`) and never validate the child transaction itself.

CIP-118 requires each child transaction to validate with its own witnesses and execution context. The exact predicate set is the current reference `SUBUTXOW`/`SUBUTXO` scope enumerated below; earlier wording here claimed a per-child fee requirement, which the reference does not have.

Value conservation is the exception, and the original wording of this issue was wrong about it. It is **not** an independent per-child check. See the batch section below.

## Work
Add validation rules that recurse into sub-transactions and validate each child with its independent witness set and execution context.

## Value conservation is computed across the whole batch

`Cardano.Ledger.Dijkstra.UTxO` computes conservation over the entire top-level transaction batch, not per transaction:

- consumed value is the top-level consumed value plus every subtransaction's consumed value;
- produced value is the top-level local produced value, plus each subtransaction's local produced value, plus the top-level fee counted **once**, plus certificate deposits computed over the certificate sequence for the whole batch.

A body's local produced value covers its outputs, treasury donation, governance proposal deposits, burned multi-assets, and direct deposits. The reference comments state explicitly that certificate deposits are excluded from each local body calculation and processed separately across the batch, so certificate state and ordering are preserved.

`dijkstra.UtxoValidateValueNotConservedUtxo` does none of this. It converts protocol parameters and delegates:

```go
return conway.UtxoValidateValueNotConservedUtxo(tx, slot, ls, tmpPparams)
```

Conway sees only the top-level Conway-shaped transaction interface, so no batch aggregation happens at any level.

A per-child conservation rule would diverge from cardano-ledger just as the current top-level-only rule does, so this issue should not be implemented that way.

## Withdrawal validation must use the evolving batch account state

[...]

PR #2446, opened 2026-09-20
fix(ledger): close Dijkstra value and UTxO validation gaps

wolf31o2:
`DijkstraTransaction.Inputs()` and `Outputs()` return only the top-level body
(`ledger/dijkstra/dijkstra.go`), so the inputs and outputs carried by
`sub_transactions` (body key 23) reached none of the rules that iterate them.
The same inputs and outputs that are rejected at the top level were accepted
when placed inside a sub-transaction: value creation, an output below the
minimum coin, an unresolvable input, an output on the wrong network.

Two more body fields had the same shape. `direct_deposits` (key 25) and
`account_balance_intervals` (key 26) were typed by #1881 so that they could be
validated, but fed only the PlutusV4 txinfo: they were decoded and placed in
the script context while contributing nothing to validation. A 500 ADA deposit
funded by a 1.2 ADA input passed value conservation, and a transaction could
assert any account balance it liked.

Related to #1880, which tracks CIP-118 sub-transaction validation as a whole;
this closes the value and UTxO part of it and does not close the issue.

## What changed

`dijkstraBatchView` presents the top-level body and every sub-transaction body
as one transaction. Registered against it:

- value conservation, folded once over the batch — inputs, outputs,
  withdrawals, certificates, mints, proposal deposits, treasury donations and
  direct deposits from every level, with the fee counted once from the
  top-level body, which is the only body that has one;
- `OutputTooSmall`, `OutputTooBig`, `OutputBootAddrAttrsTooBig` and
  `WrongNetwork` over every level's outputs;
- `BadInputs` over every level's inputs;
- `NoDuplicateInputs` across levels, so one UTxO cannot count towards the batch
  consumed total more than once.

This follows `Cardano.Ledger.Dijkstra.UTxO` and `Rules/SubUtxo.hs`:
conservation is checked once for the batch against the UTxO set as it stood
before the batch, and the per-body input and output checks are repeated for
each sub-transaction.

Reference inputs are deliberately not folded. Two levels may legitimately
reference the same UTxO, so folding them would make the duplicate-input rule
reject a valid batch.

### Direct deposits

`localProducedValue` sums a body's outputs, treasury donation, proposal
deposits, burned multi-assets and direct deposits; `dijkstraProducedValue`
applies it to the top-level body and to every sub-transaction body, then adds
the fee once. Direct deposits are therefore a produced-side term folded across
all levels, independent of the proposal deposits and treasury donations the
batch view already folds, so counting them cannot double-count either. The
balance becomes

```
inputs + withdrawals + refunds + minted
  == outputs + fee + cert deposits + proposal deposits + donation
     + direct deposits
```

A 500 ADA deposit funded by 1.2 ADA is now rejected with the same
`ValueNotConservedUtxoError` totals as the same 500 ADA paid to an output, at
either transaction level.

### Account balance intervals

A new rule checks `account_balance_intervals` against reward-account state,
following `validateAccountBalanceIntervals` in
`Cardano.Ledger.Dijkstra.Rules.Entities`: the account must be registered, and
its balance must satisfy the interval, with the lower bound inclusive and the
upper bound exclusive.

The reference runs this once per transaction level — from `SUBENTITIES` for
each sub-transaction and from `ENTITIES` for the top level — each against the
account state threaded through the levels before it, and `LEDGER` runs
`SUBLEDGERS` before `ENTITIES`. gouroboros evaluates UTxO rules against the
pre-transaction ledger state and threads no account state between levels, so a
level is checked only while the pre-transaction balances are still the ones the
reference would use: up to and including the first level carrying a withdrawal,
a certificate or a direct deposit. A transaction with no sub-transactions is
always checked in full. Later levels are left unchecked rather than checked
against a state the reference does not use there, which would reject batches
the reference accepts.

The reference's third check, `WrongNetworkInAccountBalanceIntervals`, has no
counterpart. cardano-ledger keys this map by `AccountAddress`, which carries a
network id; the field's keys on the wire are bare credentials, which do not.

## Scope this deliberately leaves out

- Per-level error types. A failure in a sub-transaction is reported with the
  existing top-level error (`ValueNotConservedUtxoError`, `BadInputsUtxoError`
  and so on), not with a distinct `SubBadInputsUTxO`-style family.
- Per-sub-transaction `MaxTxSize`, `InputSetEmpty`, validity interval and
  forecast checks, and the legacy-mode second conservation check.
- The `ENTITIES` direct-deposit checks that are not conservation terms:
  wrong-network deposit addresses, and deposits into accounts that are missing
  once the batch's certificates have been applied.
- `starting_account_balance_intervals` (key 27), which no body type decodes.
- `Consumed()` and `Produced()` still report the top-level body only, so the
  UTxO delta of a sub-transaction is not exposed. They also feed the
  PlutusV1/V2/V3 script context (`ledger/common/script/context.go`), whose
  `txInfoOutputs` must hold the body's own outputs and nothing else, so
  widening them is a separate change that has to move the script context off
  `Produced()` first. Consumers that build a UTxO set from `Produced()` are
  unchanged by this pull request and pick up the validation fix on a version
  bump; recording a batch's full UTxO delta additionally needs byte ranges for
  sub-transaction outputs, which `TransactionLocation.Outputs` does not carry.


ledger/dijkstra/rules.go
```diff
@@ -985,6 +991,179 @@ func UtxoValidateConwayFeaturesWithPlutusV1V2(
...
+
+func (t dijkstraBatchTransaction) ProposalProcedures() []common.ProposalProcedure {
+	var ret []common.ProposalProcedure
+	for _, body := range t.bodies {
+		ret = append(ret, body.ProposalProcedures()...)
+	}
+	return ret
+}
+
+func (t dijkstraBatchTransaction) Donation() *big.Int {
```
arepala-uml, line 1148: Donation is folded across levels while Witnesses and ReferenceInputs deliberately stay top-level, so the donation branch of conway.UtxoValidateValueNotConservedUtxo pairs a folded donation against the top-level script set. Measured: a top-level PlutusV1 witness with a sub-transaction donation is rejected although the V1 level carries no donation, and a sub-transaction carrying both the V1 witness and the donation is accepted, which is the case the gate exists for. conway.ValidateConwayFeaturesWithPlutusV1V2 does not cover donation. Fold it for arithmetic only and gate per level.

ledger/dijkstra/rules.go
```diff
@@ -1107,6 +1363,177 @@ func UtxoValidateBatchWithdrawals(
...
+				)
+			}
+			if err := dijkstraValidateAccountBalanceIntervals(
+				level.intervals,
+				ls,
+			); err != nil {
+				return err
+			}
+		}
+		if level.directDeposit ||
```
arepala-uml: This break disables the rule for every later level, not only for accounts an earlier level moves. Measured: a sub-transaction carrying a one-lovelace direct_deposits entry for account 0x99 makes a knowingly false top-level interval assertion about the unrelated account 0x11 return nil, and a zero-amount withdrawal from an unrelated reward account does the same. Any transaction escapes the rule by prefixing one trivial sub-transaction. Skipping only the accounts that preceding levels actually move keeps every guarantee the comment states, adds no false rejection, and removes the bypass.

ledger/dijkstra/rules_account_balance_intervals_test.go
```diff
@@ -0,0 +1,200 @@
...
+		rule(newTx(DijkstraSubTransactionBody{}, wrong), 0, ls, pp),
+		&topErr,
+	)
+	require.Len(t, topErr.Mismatches, 1)
+
+	// A sub-transaction withdrawal, certificate or direct deposit moves the
+	// balance the top level would be checked against, so the top level is
+	// left unchecked rather than checked against a state cardano-ledger does
+	// not use there.
+	require.NoError(t, rule(newTx(
```
arepala-uml, line 187: These two cases assert NoError against a top-level interval the test itself describes as wrong, so the bypass above is pinned as intended behavior rather than detected. If the skip is narrowed to the accounts that preceding levels move, both cases should expect the interval failure instead.

ledger/dijkstra/rules.go
```diff
@@ -995,7 +1174,84 @@ func UtxoValidateValueNotConservedUtxo(
...
+		dijkstraBatchView(tx),
+		slot,
+		ls,
+		pp,
+	)
+}
+
+// UtxoValidateOutputBootAddrAttrsTooBig checks every transaction level's
+// bootstrap output addresses.
+func UtxoValidateOutputBootAddrAttrsTooBig(
```
arepala-uml, line 1330: This is the one re-pointed rule with no behavioral coverage: replacing dijkstraBatchView(tx) with tx here leaves the whole ledger/dijkstra package green. The mapping test pins only that the descriptor points at this function, not that it folds, so a sub-transaction bootstrap output carrying more than 64 bytes of attributes would pass if the fold were ever reverted or refactored away. The other four re-pointed rules each fail a test when reverted in place.

ledger/dijkstra/rules.go
```diff
@@ -995,7 +1174,84 @@ func UtxoValidateValueNotConservedUtxo(
...
+// transaction level, which would otherwise count towards consumed value once
+// per level.
+func UtxoValidateNoDuplicateInputs(
+	tx common.Transaction,
+	slot uint64,
+	ls common.LedgerState,
+	pp common.ProtocolParameters,
+) error {
+	return conway.UtxoValidateNoDuplicateInputs(
+		dijkstraBatchView(tx),
```
arepala-uml, line 1305: Not folding reference inputs is correct for this rule, since folding them would reject a batch where two levels legitimately reference the same UTxO. But UtxoValidateDisjointRefInputs still receives the unfolded tx, so a sub-transaction that both spends and references the same UTxO is checked by nothing. Evaluating it through the existing dijkstraTransactionLevels walk would cover that member without reintroducing the duplicate-input problem.

ledger/dijkstra/rules.go
```diff
@@ -135,7 +137,7 @@ var utxoValidationRuleDescriptors = []common.UtxoValidationRuleDescriptor{
 	},
 	{
 		Id:        common.UtxoValidationRuleNoDuplicateInputs,
-		Validator: conway.UtxoValidateNoDuplicateInputs,
+		Validator: UtxoValidateNoDuplicateInputs,
```
arepala-uml, line 136: The cross-level consumed total depends on this rule running before value conservation, descriptor index 24 against 35. Measured with the value rule alone, two sub-transactions spending the same input against an output of twice the value returns nil. Nothing pins that ordering, although ledger/dijkstra/required_redeemers_test.go already has the pattern. The new duplicate-input test also covers only top-level against sub-transaction, not two sub-transactions spending the same input.

arepala-uml (changes requested):
Reviewed the three branch commits at faf94e20 against the merge base, then with current main merged locally: build, vet, gofmt, go test ./..., -race on ledger/dijkstra and ledger/common, conformance and golangci-lint all pass on the merged tree, and the head merge commit reverts nothing from main. The fold arithmetic is sound, verified by probe: the fee is counted once, direct deposits cannot double count with proposal deposits or donations, cross-level mints net to zero, and no body value is mutated. Two consensus-affecting defects and four non-blocking items are inline.

wolf31o2:
Addressed the review findings in the pushed commit. Treasury-donation script compatibility now runs per level while batch donation value remains counted; account-balance intervals skip only credentials changed by earlier levels. Added coverage for unrelated-account bypasses, sub-transaction reference-input overlap, bootstrap output attributes, duplicate spends across two sub-transactions, and rule ordering. Validation passed: GOWORK=off go test ./ledger/..., go vet ./ledger/..., and go build ./ledger/.... `make lint` could not run because golangci-lint is not installed.

wolf31o2:
Review sweep complete: the consensus and coverage findings from the review are addressed in 458045b0. The donation compatibility gate runs per level while donation arithmetic remains folded; interval checks skip only credentials moved earlier; sub-transaction reference-input overlap, bootstrap output attributes, cross-level duplicate spends, and rule ordering are covered. Latest CI checks pass.

arepala-uml (dismissed):

ledger/dijkstra/rules.go
wolf31o2: Fixed in the current head. Donation arithmetic stays folded for value conservation, while the Plutus V1/V2 gate now runs against each level’s own witnesses. TestDijkstraDonationScriptCheckIsPerLevel covers the false-rejection and bypass cases.

ledger/dijkstra/rules.go
wolf31o2: Fixed in the current head by threading the account overlay through each subtransaction and the top-level body. Each interval is checked against the balance at that level, so a transition for one account cannot bypass an assertion for another.

ledger/dijkstra/rules_account_balance_intervals_test.go
wolf31o2: Corrected in the current head. The mismatched top-level assertions now expect BalancesOutsideAccountBalanceIntervalsError, including the unrelated-account assertion after a prior transition.

ledger/dijkstra/rules.go
wolf31o2: Added TestDijkstraBootstrapOutputAttributesCoverSubTransactions in the current head. It uses a 100-byte Byron attribute payload and asserts the folded subtransaction output is rejected.

ledger/dijkstra/rules.go
wolf31o2: This PR intentionally keeps the Dijkstra disjoint-reference rule disabled. Issue #2490 documents the PV12 reference behavior, including the positive SUBUTXO case where the same original UTxO is both spent and referenced. The current test covers overlap at both top-level and subtransaction levels; applying the Babbage predicate per level would reject that valid case.

ledger/dijkstra/rules.go
wolf31o2: Pinned the required rule ordering and expanded duplicate-spend coverage across two subtransactions in the current head. The no-duplicate-inputs rule must precede value conservation so a repeated input cannot inflate consumed value.

arepala-uml (approved):

Issue #2242, opened 2026-09-05
ledger/common: accept trailing address bytes before Babbage


wolf31o2:
`Address.populateFromBytes` rejects any unconsumed byte after the address payload unless the network is mainnet and the trailer matches one of eight whitelisted sequences (`ledger/common/address.go:436-444`, whitelist at `ledger/common/address.go:457-485`). The check has no era or protocol-version gate.

## Reference decoder

`fromCborBothAddr` selects the decoder by version (`libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs:459-476`):

- below decoder version 7 (Shelley, Allegra, Mary, Alonzo) it uses `fromCborBackwardsBothAddr`, which calls `decodeAddrStateLenientT True True` and crops the unconsumed bytes (`Address.hs:495-503`);
- from version 7 it uses `fromCborRigorousBothAddr`, whose `decodeAddrStateLenientT ... False` reaches `unless isLenient $ ensureBufIsConsumed "Addr" buf` (`Address.hs:670-672`).

The source comment states the contract: "Prior to Babbage era we did not check if a binary blob representing an address was fully consumed, so unfortunately we must preserve this behavior."

The pre-Babbage path is unconditional on network and on trailer contents, so the whitelist admits a strict subset of what the reference accepts, and admits nothing at all on a testnet.

## Where the constraint belongs

In the decoder, gated on era: rejected from Babbage onward, accepted and cropped before it. No rule predicate covers trailing address bytes.

An address is decoded from every transaction output (`eras/shelley/impl/src/Cardano/Ledger/Shelley/TxOut.hs:158-162`, `eras/babbage/impl/src/Cardano/Ledger/Babbage/TxOut.hs:513`), so a rejection prevents the containing block from decoding.

## Wire evidence

No fixture in the repository exercises this. A Shelley transaction output whose address field is a 58-byte string consisting of a valid 57-byte type-0 address followed by one arbitrary byte is rejected with `invalid address data: 1 unexpected trailing byte(s)`; the same output without the trailing byte decodes. The reference accepts both below decoder version 7. That a block on a canonical chain carries such an address is not established here.

## Acceptance criteria

- Gate the trailing-byte check on the era: accept and crop arbitrary trailing bytes for Shelley through Alonzo on every network, reject them from Babbage onward.
- Return the cropped prefix from `Bytes()`, as the reference does for the pre-Babbage path.
[...]

Issue #2032, opened 2026-08-24
ledger: preserve credentials for all address variants


wolf31o2:
`PaymentAddress()` omits key-payment/script-stake base addresses and both
pointer variants. `StakeAddress()` omits key-payment/script-stake base
addresses and key-hash reward addresses.

Acceptance criteria:

- Preserve the payment key/script hash for every Shelley payment-address type.
- Preserve the staking key/script hash for base and reward addresses.
- Keep pointer staking extraction nil: resolving it requires ledger state.
- Preserve the network tag and exact component bytes through Bech32 round trips.
- Leave the source address and wire format unchanged.

The component layout is defined by [CIP-19](https://cips.cardano.org/cip/CIP-0019).

Issue #1109, opened 2025-07-17
Add support for Shelley pointer addresses


agaffney:
Our `Address` type needs proper support for Shelley pointer addresses, including for format as plutus data.

PR #2447, opened 2026-09-20
fix(ledger): bound pointer-address varints to their Ptr component widths

### LINKED ISSUE #1884: dijkstra: first-class CIP-159 account address type + validation
## Scope item
CIP-159 Account Address Enhancement, Phase 1. Dijkstra (PV12) hard-fork readiness. Unblocks ouroboros-mock account-address fixtures and dingo account-state handling.

## Gap
Account-style address headers (1110/1111) decode into a generic address, but there is no first-class account-address type in `ledger/common` (no `AddressTypeAccount`), and no validation of the ledger-level account-address definitions CIP-159 Phase 1 introduces.

## Work
Add a first-class account-address type (and any Phase-1 ledger-level definitions) to `ledger/common`, with validation and round-trip tests.

## Evidence
- `ledger/common/address.go` (address types base/enterprise/pointer/reward/Byron; no account type)
- CIP-159 Phase 1 / Dijkstra scope

### LINKED ISSUE #2080: address: reject overflowing pointer varints

wolf31o2:
Fixes #2080.

## Defect

`AddressPayloadPointer.decode` read each of a pointer address's three
components as an unbounded 7-bit variable-length natural, accumulating with
`ret = (ret << 7) | byte`, and applied no bound to the result. A component
encoded in more than ten 7-bit groups shifted its leading groups out of the
`uint64` accumulator and decoded to a narrowed value with no error, and a
component above its `Ptr` field's width was accepted verbatim. A 15-group slot
followed by `09` and `04` decoded as `{Slot:0 TxIndex:9 CertIndex:4}`.

A component whose continuation bits never terminate was already rejected with
`io.ErrUnexpectedEOF`; that behavior is unchanged and is covered by a test.

## Reference

`Ptr` is `Ptr !SlotNo32 !TxIx !CertIx` over `newtype SlotNo32 = SlotNo32
Word32`, `newtype TxIx = TxIx Word16` and `newtype CertIx = CertIx Word16` in
cardano-ledger. The three components are not the same width: 32 bits for the
slot, 16 for each index.

`fromCborBothAddr` picks the decoder by version (module
`Cardano.Ledger.Address`):

- below version 7, `fromCborBackwardsBothAddr`, which decodes with
  `decodeAddrStateLenientT True True`;
- versions 7 and 8, which are Babbage, `fromCborRigorousBothAddr True`;
- version 9 onward, which is Conway and Dijkstra,
  `fromCborRigorousBothAddr False`.

Only the last reaches `decodePtr`, whose `decodeVariableLengthWord32` and
`decodeVariableLengthWord16` cap the group count at five and three and fail
with `too many bytes.` beyond it. The other two reach `decodePtrLenient`, which
decodes three `Word64`s through `decodeVariableLengthWord64 = fix
(decode7BitVarLength name buf) 0` and passes them to `mkPtrNormalized`. That
decoder has no group cap and its accumulator wraps silently.

Neither `decodeVariableLengthWord16` nor `decodeVariableLengthWord32` refuses a
padded encoding. Each tests the first byte's spare high bits only while
decoding its final group, and `0b10000000` passes that test, so a padded
in-range varint decodes at every decoder version.

`fromCborBothAddr` sets `cAddrNormalized = compactAddr addr` for every
`StakeRefPtr` at every version, so the reference re-encodes a pointer address
rather than keeping its input bytes.

## Change

Decoding is lenient in every era and reproduces `decodePtrLenient`: the
components are read as wrapping `Word64`s with no group cap, then passed
through the `mkPtrNormalized` clamp that zeroes all three when any one of them
does not fit. `Bytes()` re-encodes, which is what `cAddrNormalized` does.

`decode` additionally reports which component `decodePtr` would have refused.
`populateFromBytes` records that on the address, as it records `trailingBytes`,
and the exported `CheckAddressPointerInRange` turns the record into the
rejection that decoder version 9 requires.
`ConwayTransactionBody.UnmarshalCBOR` and
`DijkstraTransactionOutput.UnmarshalCBOR` call it, as they and
`babbage.BabbageTransactionOutput.UnmarshalCBOR` call
`CheckAddressFullyConsumed`. Babbage does not call it, so Babbage normalizes.

A group cap plus one range test reproduces `decodePtr` exactly: its first-byte
check admits exactly the first-group values that keep the result inside the
component's width, and below the cap no wrap is possible, so the range test
reads a value the reference computed the same way.

## Review dispositions

All three findings on `e8ea9334` were confirmed against cardano-ledger and
fixed. The governing rule is that refusing a value the chain accepts is worse
than accepting one it rejects, because it stalls a sync from genesis.

1. `ErrAddressPointerNotCanonical` is removed. Refusing a padded in-range
   encoding rejected an address every decoder version accepts. `Bytes()`
   correctness needs no replacement mechanism, because
   `cAddrNormalized = compactAddr addr` is the reference re-encoding the same
   way.
2. Pointer strictness no longer follows `allowTrailing`, so Babbage no longer
   decodes pointers strictly and no longer splits one era by output shape. The
   rule is applied at the era's own decoder instead.
3. The lenient accumulator wraps. An 11-group slot wrapping to 5 now yields
   `Ptr 5 1 2`, as upstream does, instead of `Ptr 0 0 0`.

## Hash reachability

A re-encoded pointer address cannot change a transaction or block hash.
Transaction IDs come from `Blake2b256Hash(b.Cbor())` over the CBOR preserved by
`DecodeStoreCbor` (`ledger/common/tx.go`), and block body hashes from the raw
body bytes. The re-encoded address bytes reach `Bytes()`, `String()`,
`MarshalCBOR()` and `MarshalJSON()`, so the defect returned a wrong address to
consumers rather than corrupting a hash.

## Fixtures preserved

There are five pointer addresses in this repository. The stake reference in
cardano-ledger's Dijkstra test transaction
(`ledger/dijkstra/testdata/cardano_ledger_dijkstra_w30_tx.hex`) is
`Ptr (SlotNo32 10) (TxIx 0) (CertIx 0)`, and the four CIP-0019 bech32 vectors
in `ledger/common/address_test.go` are
`Ptr (SlotNo32 2498243) (TxIx 27) (CertIx 3)`. All five decode to the same
values as before, all five are accepted by `CheckAddressPointerInRange`, and
all five re-encode to their input bytes. Every other era block, Byron fixture
and fuzz corpus entry in the repository was checked for a pointer address and
carries none.

## Tests

`ledger/common/address_pointer_test.go` covers the CIP-0019 vector, the maximum
in-range components, a leading zero group, every component padded to its group
cap, each component one past its bound, group counts past each cap, a 15-group
component, the two wrapping cases, an unterminated component, a truncated
component, the Dijkstra fixture, an exhaustive round-trip table over boundary
values, and a fuzz target asserting that `Bytes()` is canonical and that
re-decoding it is a fixed point. Each case is run through both
`NewAddressFromBytes` and `NewAddressFromBytesLenient`, which must now agree.

`ledger/address_pointer_era_test.go` is the era gate: Shelley, Mary, Alonzo and
both Babbage output shapes normalize an out-of-range pointer, while the Conway
body and both Dijkstra output shapes reject it, and a padded in-range pointer
is accepted by Conway and Dijkstra.

## Adjacent work

Does not close #1884. Builds on the era flag added by #2253 for #2242, and on
the payload-preservation work in #2328 for #2032, whose `NewAddressFromParts`
pointer fallthrough routes pointer addresses through this decoder. The decoder
itself came from #1109.


ledger/common/address.go
```diff
@@ -1074,34 +1078,148 @@ type AddressPayloadPointer struct {
...
+		// does not fit, because the result is a dangling pointer regardless.
+		slot, txIndex, certIndex = 0, 0, 0
+	}
+	a.Slot, a.TxIndex, a.CertIndex = slot, txIndex, certIndex
+	// Bytes() re-encodes the pointer rather than replaying the input, so a
+	// strict decode is only sound if the two agree. Lenient decoding is exempt:
+	// the reference also replaces a normalized pointer's bytes with the
+	// re-encoded form.
+	if strict {
+		if !bytes.Equal(a.encode(), data[:offset]) {
```
arepala-uml: cardano-ledger accepts a padded in-range varint at every decoder version: decodeVariableLengthWord16 and decodeVariableLengthWord32 check the first byte's high bits only in the final group, and 0b10000000 passes that check. Slot bytes 80 00 decode to 0 upstream and are rejected here. fromCborBothAddr already answers the round-trip concern by setting cAddrNormalized = compactAddr addr for every StakeRefPtr at every version, so the reference normalizes rather than rejects. Drop ErrAddressPointerNotCanonical: a Babbage or later output carrying a padded pointer currently fails to decode.

ledger/common/address.go
```diff
@@ -438,7 +440,9 @@ func (a *Address) populateFromBytes(data []byte, allowTrailing bool) error {
 		payload = payload[AddressHashSize:]
 	case AddressTypeKeyPointer, AddressTypeScriptPointer:
 		var tmpPointer AddressPayloadPointer
-		n, err := tmpPointer.decode(payload)
+		// allowTrailing marks the pre-Babbage output path, the same eras for
+		// which cardano-ledger keeps the pointer decoder lenient.
+		n, err := tmpPointer.decode(payload, !allowTrailing)
```
arepala-uml: fromCborBothAddr gates strict decodePtr at decoder version 9; versions 7 and 8 take fromCborRigorousBothAddr True, so Babbage decodes pointers leniently. Keying strictness to allowTrailing also splits one era by output shape: a Babbage map-form output rejects an out-of-range pointer while the legacy array form reaches NewAddressFromBytesLenient via AlonzoTransactionOutput and normalizes it. Record the out-of-range pointer as populateFromBytes records trailingBytes, and reject it from the Conway and Dijkstra decoders as CheckAddressFullyConsumed does.

ledger/common/address.go
```diff
@@ -1074,34 +1078,148 @@ type AddressPayloadPointer struct {
...
+	// at most five and three groups respectively and fail with "too many
+	// bytes." on a sixth or fourth
+	// (libs/cardano-ledger-core/src/Cardano/Ledger/Address.hs).
+	addressPointerSlotGroups  = 5
+	addressPointerIndexGroups = 3
+
+	// A uint64 accumulator holds ten 7-bit groups. Past that the leading bits
+	// are shifted away, which is the silent narrowing this bound exists to
+	// detect.
+	addressPointerUint64Groups = 10
```
arepala-uml: decodeVariableLengthWord64 is fix (decode7BitVarLength), so it has no group cap and wraps the Word64 accumulator silently, and a wrapped value can land back in range. An 11-group slot wrapping to 5 yields Ptr 5 1 2 upstream against Ptr 0 0 0 here, so the stated rationale that the pointer is dangling under either rule does not hold for that case.

arepala-uml (changes requested):
Bounds, group caps and mkPtrNormalized semantics match cardano-ledger; the strict path does not. It rejects padded in-range encodings the reference accepts at every decoder version, applies Conway's decodePtr from Babbage, and splits a single era by output encoding shape. Each rejects an address the node accepts on the block decode path. Merged-tree build, go test ./ledger/..., gofmt and golangci-lint are clean. CodeRabbit is rate limited and cubic is over its monthly limit, so no bot reviewed this head.

ledger/common/address.go
wolf31o2: Fixed, and the declared deviation is withdrawn. ErrAddressPointerNotCanonical is deleted: the first-byte check sits inside d7last, so 80 00 is never checked at all and 0b10000000 passes it even at the group cap. The Bytes() justification collapsed on fromCborBothAddr setting cAddrNormalized = compactAddr addr for every StakeRefPtr at every version, so the reference re-encodes as well. Pinned by the padded cases in TestAddressPointerComponentBounds and by TestConwayOnwardOutputsAcceptPaddedPointers.

ledger/common/address.go
wolf31o2: Fixed, and the declared deviation is withdrawn. Strictness no longer follows allowTrailing: decoding reaches decodePtrLenient in every era and records what decodePtr would refuse, and CheckAddressPointerInRange applies that record from ConwayTransactionBody.UnmarshalCBOR and DijkstraTransactionOutput.UnmarshalCBOR. Reverting it reproduced the split by output shape exactly: babbage_map_form failed while babbage_legacy_array_form passed. Pinned by TestPreConwayOutputsNormalizeOutOfRangePointers and TestConwayOnwardOutputsRejectOutOfRangePointers.

ledger/common/address.go
wolf31o2: Confirmed and fixed. The lenient accumulator now wraps with no group cap, reproducing decodeVariableLengthWord64 = fix (decode7BitVarLength name buf) 0; the previous ten-group cap diverged by normalizing instead. The 11-group slot yields Ptr 5 1 2, and decodePtr's five-group cap still rejects the same bytes from Conway onward. Pinned by TestAddressPointerWrapsPastUint64 and TestAddressPointerWrapsToZero.

arepala-uml (approved):

Issue #2031, opened 2026-08-24
kes: bound SumKES input size and recursion


wolf31o2:
`kes/kes.go` `NewSumKesFromBytes` checks a recursively computed size with arithmetic that can overflow uint64 and has no recursion-depth bound. Crafted nested SumKES bytes can bypass the size guard or exhaust stack/work before returning an error.

Acceptance criteria:

- Check size arithmetic before addition/multiplication and enforce a recursion/depth limit.
- Reject oversized and deeply nested inputs without panic or excessive work.
- Add boundary and malformed vectors.

PR #2460, opened 2026-09-20
fix(protocol): contain panics raised on mini-protocol goroutines

### LINKED ISSUE #2009: cbor: reject zero-denominator rationals without panicking
`cbor/tags.go` permits a zero denominator in `Rat` JSON/genesis construction. `PoolRegistrationCertificate.UnmarshalJSON`, `NewGenesisRat`, and Alonzo genesis conversion can then panic while constructing or using the rational.

Acceptance criteria:

- Reject zero denominators with a normal error at decode/construction time.
- Do not partially construct a certificate or genesis configuration.
- Cover JSON, genesis, and Alonzo parameter boundary cases.

### LINKED ISSUE #2075: protocol: contain handler panics
Message handlers and pipeline workers do not have a common panic boundary. An
unexpected panic can terminate a protocol goroutine without returning a stable
connection or handler error.

Acceptance criteria:

- Recover panics at the message-handler and pipeline-worker boundaries.
- Convert failures into the existing error and shutdown flow.
- Preserve cleanup and connection accounting.
- Cover handler and pipeline panic cases.

wolf31o2:
## Problem

Mini-protocol and pipeline goroutines are started by the library, so a panic in decoding, a handler, or a consumer callback has no caller-owned recovery frame above it. One peer message or block item can therefore terminate the process.

The ordered apply runner needs a stricter disposition than independent decode and validation workers. A panic can occur after apply bookkeeping has advanced; treating that input as an ordered singleton could move a completion fence across an unresolved sequence gap.

## Changes

- Recover panics at the read, receive, state, and send loop boundaries.
- Add `Protocol.RunLoop` and use it for the Leios notification/vote callback loops and the block-fetch shutdown watcher.
- Report a contained protocol panic on the protocol error channel and stop that protocol, including when a shutdown callback panics after `DoneChan` is closed.
- Recover stage, metrics-predicate, and metrics-recorder panics in the worker pool without losing the worker or item.
- Recover `ApplyFunc` panics at the item boundary so buffered items retain their ordering.
- Treat apply-runner bookkeeping and completion-callback panics as pipeline-fatal: report the error, do not advance the completion watermark across the unresolved item, and cancel the owning pipeline so fences return.
- Preserve the panic value, stack, and originating frame in the reported error.

## Compatibility

This is additive: it introduces `protocol.ErrHandlerPanic`, `pipeline.ErrStagePanic`, and `Protocol.RunLoop`. Existing exported signatures and ordinary error paths are unchanged.

Consumers that already close a connection on any protocol error keep the same behavior; the difference is that the process survives and the affected protocol is stopped explicitly.

## Verification

- `make test`
- `go test -race ./pipeline ./protocol ./protocol/leiosnotify ./protocol/leiosvotes ./protocol/blockfetch`
- `go test ./...`
- `go vet ./...`
- `go build ./...`
- `make format`

`make lint` was not run locally because `golangci-lint` is not installed; hosted CI covers it.

Does not close #2009 or #2031, which track specific panic sources rather than boundary containment.

Fixes #2075


pipeline/apply_stage.go
```diff
@@ -328,7 +343,7 @@ func (r *ApplyStageRunner) run(ctx context.Context) {
 				return
 			}
 
-			processed, err := r.stage.ProcessWithStatus(ctx, item)
+			processed, err := r.process(ctx, item)
```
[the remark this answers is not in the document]
wolf31o2: Resolved in e23c9db with a pipeline-fatal disposition rather than singleton forwarding. `ProcessWithStatus` can panic after advancing `nextSequence` or mutating the pending buffer, so forwarding only the current item and invoking `processedFunc` could move a `Fence` across an unresolved predecessor or omit buffered items already released internally. The runner now records the failure as `ApplyError`, reports `ErrStagePanic`, cancels the owning pipeline, and exits without forwarding or advancing completion. `TestApplyStageRunnerContainsStagePanic` queues sequence 1 before 0 and asserts no output or completion advance plus fatal cancellation, so `Fence`/`WaitForDrain` return through pipeline cancellation instead of wedging or crossing the gap.

protocol/leiosvotes/client.go
```diff
@@ -168,7 +168,9 @@ func (c *Client) Sync() error {
 	}
 	c.syncRunning = true
 	c.sendMutex.Unlock()
-	go c.voteLoop(uint64(pipelineLimit) * c.config.RequestNextCount)
+	go c.Protocol.RunLoop("vote loop", func() {
+		c.voteLoop(uint64(pipelineLimit) * c.config.RequestNextCount)
```
arepala-uml: golangci-lint reports G115 here at the merged head: the int to uint64 conversion is newly flagged because it moved inside the new closure. One binary at 2.12.2 and one .golangci.yml over the same package set give 0 issues on origin/main and 3 at this head, exit 1, and AGENTS.md makes make lint a submit gate. Nothing reports this today because no lint or go-test check run exists at this head at all, only CodeQL, Analyze and DCO, although both workflows trigger on pull_request.

protocol/leiosnotify/client.go
```diff
@@ -123,7 +123,7 @@ func (c *Client) Sync() error {
 	c.pipelinedRequestNext = 0
 	c.notificationRunning = true
 	// Start notification loop
-	go c.notificationLoop()
+	go c.Protocol.RunLoop("notification loop", c.notificationLoop)
```
arepala-uml: staticcheck QF1008 fires on this line and on leiosvotes/client.go:171 at the merged head: the embedded Protocol field should be elided from the selector. These are the other two issues in the clean-to-failing lint differential.

protocol/blockfetch/client.go
```diff
@@ -359,7 +359,9 @@ func (c *Client) Start() {
 			c.lifecycleState = clientStateRunning
 			// Resolve any request left outstanding when the protocol shuts
 			// down, so no caller and no callback consumer is left waiting.
-			go c.failOutstandingOnProtocolDone(proto)
+			go proto.RunLoop("shutdown watcher", func() {
```
arepala-uml: Reverting this call site to go c.failOutstandingOnProtocolDone(proto) leaves the package green, and the same holds for leiosnotify/client.go:126 and leiosvotes/client.go:171. TestShutdownWatcherContainsRangeDonePanic and its two siblings build RunLoop in the test body instead of going through Start or Sync, so they re-assert protocol.RunLoop, which TestRunLoopReportsAndStops already covers, and not this wiring. Dropping the wrapper in a later refactor or merge resolution restores the process-killing panic with no test failing.

protocol/leiosnotify/client.go
arepala-uml: notificationLoop holds c.busyMutex across c.SendMessage and its own defer re-locks busyMutex, so a panic in that region self-deadlocks the goroutine before recoverLoop can run and the containment becomes a silent hang with no error reported. The shape pre-dates this change and the new guard does not reach it.

protocol/protocol.go
```diff
@@ -579,6 +580,50 @@ func (p *Protocol) SendError(err error) {
...
+func (p *Protocol) recoverLoop(where string) {
+	err := panics.New(ErrHandlerPanic, p.config.Name+": "+where, recover())
+	if err != nil {
+		// Report directly instead of using SendError: a protocol-owned
+		// shutdown watcher can execute a consumer callback after DoneChan is
+		// closed, and SendError deliberately ignores ordinary errors once
+		// shutdown has begun. A panic still needs to remain diagnosable.
+		select {
+		case p.config.ErrorChan <- err:
+		default:
```
arepala-uml, line 614: This default drops the panic error when ErrorChan, buffered at 10, is full, while protocol/panic.go documents the panic as reported on the error channel. Consistent with the existing SendError behaviour, but the stack trace then becomes the only record of the fault.

arepala-uml (changes requested):
The design is sound and the class audit came back complete: exactly three library-owned goroutines invoke a consumer callback and all three are covered, and the three earlier CodeRabbit comments are addressed at head. Merged against current main the tree builds and passes race tests on seven packages. Blocking on three items: golangci-lint goes from 0 issues on main to 3 at the merged head on lines this change wrote, and make lint is a submit gate; the three production RunLoop call sites are pinned by no test; and closes #2009 is false, so merging would close a live panic defect.

wolf31o2:
Addressed the current review findings in the pushed commit. Added callback startup-path panic tests for blockfetch, leiosnotify, and leiosvotes; fixed the two lint findings; removed the notification-loop lock-held send that could deadlock panic recovery; and log panic details when the error channel is full. Focused race validation passed: GOWORK=off go test -race ./protocol ./protocol/blockfetch ./protocol/leiosnotify ./protocol/leiosvotes. `make lint` remains unavailable locally because golangci-lint is not installed.

arepala-uml (approved):

PR #2461, opened 2026-09-20
fix(ledger): bound governance anchor URLs at decode

wolf31o2:
`GovAnchor.UnmarshalCBOR` decoded `anchor_url` as a bare string with no length
check, so governance anchors the reference ledger rejects were accepted. The
CDDL rule is `url = text .size (0 .. 128)`
(`ledger/dijkstra/testdata/dijkstra.cddl:514`), used by
`anchor = [anchor_url : url, anchor_data_hash : hash32]` at `:549`.
cardano-ledger reaches the same bound through the `DecCBOR` instance of its
shared `Url` type.

- Reject an anchor URL longer than 128 UTF-8 bytes at decode with
  `ErrGovAnchorURLTooLong`. A 128-byte URL still decodes unchanged.
- The pool metadata URL constants name the same CDDL rule, so they become
  `urlMaxLength` / `urlMaxLengthLegacy` and are shared. The legacy 64-byte
  bound cannot reach an anchor: anchors first exist in Conway, where the bound
  is already 128.
- `GovAnchor.UnmarshalJSON` stays unbounded, matching the reference ledger's
  `Url` `FromJSON` instance and the pool metadata decision in gouroboros#2209.
  Conway genesis carries a constitution anchor through that path.
- DRep registration and update certificates, committee resignation
  certificates, voting procedures, proposal procedures and the constitution all
  hold `common.GovAnchor`, so one decoder covers them. Tests assert rejection
  through four of those containers and assert the JSON path still accepts a
  longer URL.

Same class as gouroboros#2443 and gouroboros#2450. Related to gouroboros#2209,
which bounded pool metadata URLs and left anchors untouched. Pool relay IP
widths are a separate unenforced CDDL size rule tracked in gouroboros#2425 and
are not addressed here.

---
## Summary by cubic
Rejects governance anchor URLs longer than 128 UTF-8 bytes at CBOR decode, so anchors the reference ledger rejects are no longer accepted.

**Bug Fixes**
- Adds `ErrGovAnchorURLTooLong`; a 128-byte URL still decodes unchanged.
- Renames the pool metadata URL constants to `urlMaxLength` / `urlMaxLengthLegacy` and shares them across both rules.
- Keeps `GovAnchor.UnmarshalJSON` unbounded to match the reference ledger and existing pool metadata behavior.
- Covers all anchor-bearing certificate and governance types through the shared `GovAnchor` decoder, with tests for four containers.

<sup>Written for commit 3466dc39e226dae6a32e7a3de5c262bcdbbbecc8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2461?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ledger/common/strict_decode.go
```diff
@@ -259,6 +259,17 @@ func (a *GovAnchor) UnmarshalCBOR(cborData []byte) error {
 	if _, err := cbor.Decode(cborData, &decoded); err != nil {
 		return fmt.Errorf("decode governance anchor: %w", err)
 	}
+	// anchor_url is the CDDL url rule, bounded at 128 bytes. The 64-byte
+	// legacy bound cannot apply here: anchors first exist in Conway, which
+	// is the protocol version at which the bound is already 128.
+	if len(decoded.Url) > urlMaxLength {
```
arepala-uml: The 128-byte bound is applied at decode only. GovAnchor has no MarshalCBOR and NewGovAnchor accepts any length, so the type no longer round-trips: a constitution anchor decoded from JSON unbounded reaches localstatequery.ConstitutionResult.Anchor, encodes, and then fails to decode on the receiving side. PoolMetadata guards both directions for the same CDDL url rule, at certs.go:641 and certs.go:649. Probe at this head: NewGovAnchor with a 129-byte URL returns no error and encodes, while UnmarshalCBOR of those exact bytes rejects them.

ledger/common/gov_anchor_url_test.go
```diff
@@ -0,0 +1,149 @@
...
+		hex.EncodeToString(govAnchorTestHash()) + `"}`
+
+	var anchor common.GovAnchor
+	require.NoError(t, anchor.UnmarshalJSON([]byte(body)))
+	require.Equal(t, overLimit, anchor.Url)
+}
+
+// TestGovAnchorURLBoundAtEveryDecodeSite checks that the containers carrying
+// an anchor reject an over-long URL through the shared GovAnchor decoder.
+func TestGovAnchorURLBoundAtEveryDecodeSite(t *testing.T) {
```
arepala-uml, line 125: This table covers 4 of the 6 CDDL sites carrying anchor; proposal_procedure and constitution are absent from it. Both enforce the bound in the code, so the class itself is complete and it is the coverage claimed by the test name that is short.

ledger/common/gov_anchor_url_test.go
```diff
@@ -0,0 +1,149 @@
...
+	require.NoError(t, accepted.UnmarshalCBOR(govAnchorCBOR(t, atLimit)))
+	require.Equal(t, atLimit, accepted.Url)
+	require.Equal(t, govAnchorTestHash(), accepted.DataHash[:])
+
+	overLimit := govAnchorURL("https://example.com/", govAnchorURLMaxBytes+1)
+	require.Len(t, overLimit, govAnchorURLMaxBytes+1)
+
+	var rejected common.GovAnchor
+	err := rejected.UnmarshalCBOR(govAnchorCBOR(t, overLimit))
+	require.NotNil(t, err)
```
arepala-uml: require.NotNil(t, err) should be require.Error(t, err), and the following require.Contains on the text got 129 couples the assertion to the error-string format rather than to ErrGovAnchorURLTooLong.

arepala-uml (changes requested):
Bound value, boundary and bytes-not-runes semantics all check out against the CDDL rule url = text .size (0 .. 128), and the class audit came back complete: every anchor-bearing container enforces 128. Blocking on one gap, that the bound is applied at decode only, so GovAnchor no longer round-trips. Merged against current main the tree builds clean and passes ledger tests, race and lint; removing the bound block fails 6 of the 7 new tests, so causality is proven. Neither review bot covered this head: CodeRabbit declined this exact range at its limit and cubic is at its monthly cap.

wolf31o2:
Addressed review feedback in 83c6b714. GovAnchor now applies the same 128-byte UTF-8 bound when constructed, CBOR encoded, and decoded; tests cover round-trip behavior and all six anchor-bearing CBOR containers. Focused validation passed: GOWORK=off go test ./ledger/common ./ledger/conway.

ledger/common/gov.go
```diff
@@ -530,6 +546,18 @@ func NewGovAnchor(url string, dataHash []byte) (GovAnchor, error) {
 	}, nil
 }
 
+func validateGovAnchorURL(url string) error {
```
arepala-uml, line 550: This PR's validateGovAnchorURL and #2443's MaxGovActionIdx insert independent declarations at the same point in gov.go, producing a genuine content conflict against current main (mergeable=CONFLICTING). Resolving needs only a keep-both merge, no design change, but the branch must be rebased or merged before this can land.

arepala-uml (changes requested):
Blocking item from the prior review (decode-only anchor URL bound) is fixed and verified with a causal test; new blocker is a real merge conflict against main that leaves this PR unmergeable as-is.

wolf31o2:
Resolved the confirmed conflict with current main in a signed merge commit (a2b84c6c), keeping both the anchor URL validator and MaxGovActionIdx declaration. Focused ledger/common and ledger/conway tests pass; CI is running.

arepala-uml (approved):

PR #2463, opened 2026-09-20
fix(localstatequery): bound pool metadata URLs in query results

wolf31o2:
`StakePoolParamsResult` and `PoolStateParams` decode pool metadata through
inline unnamed structs rather than `common.PoolMetadata`, so
`PoolMetadata.UnmarshalCBOR` never runs and the CDDL rule
`url = text .size (0 .. 128)` (`ledger/dijkstra/testdata/dijkstra.cddl`) was
not enforced on the `GetStakePoolParams` or `GetPoolState` result paths.
cardano-ledger reaches the same bound through the `DecCBOR` instance of its
shared `Url` type, which `PoolMetadata` and `Anchor` both use.

- Both results now validate at decode through the existing exported
  `common.ValidatePoolMetadata`, returning `ErrPoolMetadataURLTooLong`. A
  128-byte URL still decodes unchanged; 129 bytes is refused, measured in
  UTF-8 bytes.
- The exported field sets are unchanged. The inline metadata hash is
  `ledger.Blake2b256`, a fixed 32-byte array, while `common.PoolMetadata.Hash`
  is `PoolMetadataHash`, deliberately unbounded since gouroboros#2236.
  Substituting the shared type would rename an exported field, change its
  type, and relax the hash length check on these paths in one step, so the
  bound is applied in place instead.
- `StakePoolParamsResult.Results` has an unnamed struct as its map value, which
  cannot carry a decoder, so its bound runs in the containing result's
  decoder. `PoolStateParams` is a named type and carries its own.
- Tests drive each site separately: reverting either check leaves the other
  site's tests passing. Results within the bound, and results carrying no
  metadata, are asserted field for field.

These are local-node query results, so the bytes come from the node an
operator chose to connect to rather than from an arbitrary peer. gouroboros is
a library and a caller may point it at a node it does not control, but the
case for the bound here is consistency with the rest of the decode domain, not
an imminent attack.

Same rule as gouroboros#2209, which bounded pool metadata URLs for
gouroboros#2183 and left these two query paths untouched, and as
gouroboros#2461, which bounds governance anchor URLs. No dependency on
gouroboros#2461: this change calls the exported validator rather than the
constant that gouroboros#2461 renames, so the two do not conflict in either
merge order. Pool relay IP widths are a separate unenforced CDDL size rule
tracked in gouroboros#2425 and are not addressed here.

---
## Summary by cubic
Bounds pool metadata URLs in `localstatequery` `GetStakePoolParams` and `GetPoolState` results, so both query paths now enforce the 128-byte `url` rule used elsewhere in pool metadata decoding. URLs up to 128 UTF-8 bytes decode as before; longer URLs are rejected with `ErrPoolMetadataURLTooLong`.

- The limit is measured in UTF-8 bytes, matching cardano-ledger's `Url` type.
- Exported result fields are unchanged; validation runs in the result decoders.
- Results without pool metadata are unaffected.

<sup>Written for commit 23f42457589fb9e1fbfe2748465832ad8b48660f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/gouroboros/pull/2463?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


protocol/localstatequery/queries.go
```diff
@@ -958,6 +958,38 @@ type StakePoolParamsResult struct {
...
+// Those structs are unnamed, so they cannot carry a decoder and
+// common.PoolMetadata.UnmarshalCBOR never runs on a query result path.
+func validateQueryPoolMetadataURL(url string) error {
+	return lcommon.ValidatePoolMetadata(&lcommon.PoolMetadata{Url: url})
+}
+
+// UnmarshalCBOR decodes a stake pool parameters query result and bounds each
+// pool metadata URL. The map value is an unnamed struct type, so the bound
+// runs here rather than in a decoder of its own.
+func (r *StakePoolParamsResult) UnmarshalCBOR(data []byte) error {
```
arepala-uml, line 972: Neither StakePoolParamsResult nor PoolStateParams gained a MarshalCBOR, so both encode a URL longer than 128 bytes without error and then fail to decode their own output. server.go:205 encodes whatever a QueryFunc returns, so a gouroboros server can emit CBOR violating the CDDL rule this bound cites, and every gouroboros client then fails the whole query. Probe: cbor.Encode of PoolStateParams with a 220-byte URL succeeds, and cbor.Decode of those bytes rejects them. lcommon.PoolMetadata.MarshalCBOR at certs.go:610 validates on encode; this reuses its validator but not its symmetry.

protocol/localstatequery/pool_metadata_url_test.go
```diff
@@ -0,0 +1,298 @@
...
+	var accepted StakePoolParamsResult
+	require.NoError(
+		t,
+		accepted.UnmarshalCBOR(stakePoolParamsCBOR(t, atLimit)),
+	)
+	params, ok := accepted.Results[testPoolId()]
+	require.True(t, ok)
+	require.NotNil(t, params.PoolMetadata)
+	require.Equal(t, atLimit, params.PoolMetadata.Url)
+	require.Equal(t, testPoolMetadataHash(), params.PoolMetadata.MetadataHash)
```
arepala-uml: This asserts require.Equal against ledger.Blake2b256 while the open #2469 widens PoolMetadata.MetadataHash to lcommon.PoolMetadataHash. Merging the two branches gives no textual conflict and compiles, then fails this test and TestPoolStateResultURLBound at line 198 on Not equal. Neither branch's checks can see it, so whichever merges second breaks main unless branches are required to be up to date. The fixture and testPoolMetadataHash() need to follow the widened type.

protocol/localstatequery/queries.go
```diff
@@ -958,6 +958,38 @@ type StakePoolParamsResult struct {
...
+	if r == nil {
+		return errors.New("nil StakePoolParamsResult receiver")
+	}
+	type stakePoolParamsResult StakePoolParamsResult
+	var tmp stakePoolParamsResult
+	if _, err := cbor.Decode(data, &tmp); err != nil {
+		return err
+	}
+	for poolId, params := range tmp.Results {
```
arepala-uml, line 981: One over-long URL rejects the entire multi-pool reply, so GetStakePoolParams returns nothing at all rather than dropping the single bad entry.

protocol/localstatequery/queries.go
```diff
@@ -1044,6 +1076,27 @@ type PoolStateParams struct {
...
+		return errors.New("nil PoolStateParams receiver")
+	}
+	type poolStateParams PoolStateParams
+	var tmp poolStateParams
+	if _, err := cbor.Decode(data, &tmp); err != nil {
+		return err
+	}
+	if tmp.PoolMetadata != nil {
+		if err := validateQueryPoolMetadataURL(tmp.PoolMetadata.Url); err != nil {
+			return err
```
arepala-uml, line 1106: This returns the bare error while the StakePoolParamsResult path wraps with pool %s, so a GetPoolState failure names no pool.

protocol/localstatequery/pool_metadata_url_test.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	atLimit := testPoolURL(queryPoolMetadataURLMaxBytes)
+	var accepted PoolStateParams
+	require.NoError(t, accepted.UnmarshalCBOR(encode(atLimit)))
+	require.NotNil(t, accepted.PoolMetadata)
+	require.Equal(t, atLimit, accepted.PoolMetadata.Url)
+
+	var rejected PoolStateParams
+	err := rejected.UnmarshalCBOR(encode(testPoolURL(queryPoolMetadataURLMaxBytes + 1)))
```
arepala-uml, line 258: This line is 85 columns against the repo's 80-column limit, and golines --max-len=80 reformats exactly this line and nothing else in the diff. golangci-lint does not catch it; make golines does.

arepala-uml (changes requested):
The bound is correct - dijkstra.cddl:514 gives url = text .size (0 .. 128), byte-measured, and the value, the boundary and the measure all match - and three fail-before reverts show the two sites are independently covered. Blocking on two items: the bound is applied at decode only, so both types encode a URL they then cannot decode back, and this branch merged with the open #2469 compiles and fails two of this PR's own tests, which neither branch's checks can see. Merged against current main the tree builds and passes race tests and lint. Neither review bot covered this head.

wolf31o2:
Addressed review feedback in 486ea6d7. Pool metadata URL bounds now apply on CBOR encode and decode; GetStakePoolParams filters only the offending pool, while GetPoolState errors include the pool id. Tests also use byte-slice comparisons for metadata hashes so they remain compatible with the widened hash type in #2469. Focused validation passed: GOWORK=off go test ./protocol/localstatequery.

wolf31o2:
Fixed the CI findings: compare metadata hashes using the same byte-slice type and explicitly guard the pool params pointer so NilAway can prove the dereference safe. Focused package tests pass locally.

protocol/localstatequery/pool_metadata_url_test.go
```diff
@@ -0,0 +1,356 @@
...
+	require.ErrorIs(t, err, lcommon.ErrPoolMetadataURLTooLong)
+}
+
+func TestQueryPoolMetadataURLEncodingBound(t *testing.T) {
+	tooLong := testPoolURL(queryPoolMetadataURLMaxBytes + 1)
+	decoded, err := decodePoolStateResult(
+		poolStateCBOR(t, "https://pool.example/"),
+	)
+	require.NoError(t, err)
+	params := *decoded.PState[testPoolOperator()]
```
arepala-uml: nilaway fails here (exit 3, currently red on this head): params := *decoded.PState[testPoolOperator()] dereferences a map lookup with no ok or nil guard, unlike the same pattern already fixed in TestPoolStateResultURLBound. A missing or nil PState entry panics.

arepala-uml (changes requested):
Both blocking items from the prior review are fixed and verified on the tree merged with main; new blocker is a currently failing nilaway check caused by this PR's own test code.

wolf31o2:
Fixed the new NilAway finding by checking the pool map lookup and guarding its pointer before dereference (e45c5546). Focused local package test passes; CI is running.

arepala-uml (approved):

PR #2492, opened 2026-09-21
fix(conway): evaluate GOV predicates against post-CERTS state

### LINKED ISSUE #2386: ledger/conway: evaluate GOV predicates against post-CERTS state
## Summary

Conway GOV validation evaluates several predicates against the pre-transaction ledger state. The reference applies CERTS first and passes the resulting cert state to GOV.

Without a transaction-scoped post-CERTS state, same-transaction certificate changes are invisible to governance validation.

## Concrete divergences

- RegisterDRep(D) then vote as D: reference accepts; pre-state lookup rejects.
- DeregisterDRep(D) then vote as D: reference rejects; pre-state lookup can accept.
- RegisterPool(P) then vote as P: reference accepts; pre-state lookup rejects.
- Authorize a committee hot key then vote: reference can accept; pre-state lookup can reject.
- Resign a committee member then vote with the old hot key: reference rejects; pre-state lookup can accept.
- Register a stake account then use it as a proposal or treasury-return account: reference accepts; pre-state lookup rejects.
- Deregister then use the account: reference rejects; pre-state lookup can accept.

## Acceptance criteria

- Apply certificates sequentially to a transaction-scoped overlay before GOV predicates.
- Use the overlay for DRep, pool, committee-hot, stake-account, proposal-return, and treasury-return queries.
- Preserve withdrawals versus CERTS ordering.
- Preserve full key/script identity.
- Add all listed accept/reject regressions through the complete Conway pipeline.
- Ensure proposal-and-vote behavior remains correct.
- Share one overlay across GOV predicates.
- Do not mutate persisted state during validation.
- Ensure rejected transactions leave no side effects.
- Apply equivalent behavior in Dijkstra.
- Keep individual certificate legality issues #2384, #4377, #4433, and elected-voter issue #2381 separate.

chrisguiney:
Fixes gouroboros#2386.

UtxoValidateProposalReturnAccounts and UtxoValidateUnknownVoters checked
DRep, pool, committee-hot, and stake-account registration only against
pre-transaction ledger state. The reference LEDGER rule
(conwayLedgerTransitionTRC) runs CERTS before GOV and passes GOV the
resulting CertState, so a certificate earlier in a transaction's own
certificate list is visible to that transaction's own governance checks.

Adds conwayCertsOverlay, a read-only, transaction-scoped view built by
folding a transaction's own certificates over ledger state, shared by
both predicates. It never mutates the underlying ledger state.

Fixes a live divergence: a Preview transaction registering a stake
account and using it as its own proposal's return account was rejected
locally while the network accepted it - confirmed via direct CBOR
decode of the transaction plus independent Koios confirmation of its
on-chain canonicity.

Withdrawal validation is unaffected: it already runs before CERTS
against unmodified ledger state, matching the reference's
pre-CERTS-drain ordering, verified against current cardano-ledger
source directly.

UtxoValidateProposalReturnAccounts returns before constructing the
overlay when a transaction carries no proposal procedures, keeping the
certificate fold and its map allocations off the path of every
transaction that proposes nothing. UtxoValidateUnknownVoters already
guards the same way on an empty vote list. The two predicates build
independent overlays, so neither guard changes what the other checks.

Individual certificate legality (UtxoValidateDelegation,
UtxoValidateCertificateDeposits, UtxoValidateCommitteeCertificates) and
elected-committee-voter validation are unchanged; those are separate
issues: gouroboros#2384, gouroboros#2381, blinklabs-io/dingo#4377, and
blinklabs-io/dingo#4433.

Dijkstra reuses these two predicates directly from conway, so no
separate Dijkstra change is needed.

Nine new tests cover all 7 divergence scenarios from gouroboros#2386
plus a no-certificate control, each proven to fail on current main and
pass with the fix, including explicit ledger-state-unmutated-on-reject
assertions.


cryptodj413 (dismissed):

ledger/conway/rules.go
```diff
@@ -1343,6 +1343,12 @@ func UtxoValidateProposalReturnAccounts(
 	if isInConwayBootstrapPhase(pp) {
 		return nil
 	}
+	// The reference (conwayGovTransition, processProposal) checks the
+	// return address and treasury withdrawal addresses against
+	// certStateAfterCERTS, not against pre-transaction ledger state: a
+	// stake account this same transaction registers via an earlier
+	// certificate is a valid return account (gouroboros#2386).
+	overlay := newConwayCertsOverlay(tx, ls)
```
wolf31o2, line 1355: This folds the certificates and allocates five maps for every transaction, before the code knows whether tx.ProposalProcedures() is empty; measured at 1063 ns/op and 6 allocs/op for a transaction with no certificates and no proposals. UtxoValidateUnknownVoters already guards the same construction behind its empty-votes early return at line 4277. Hoisting proposals := tx.ProposalProcedures() above this line and returning nil when it is empty removes the cost from the common case and keeps the two predicates symmetric.

wolf31o2 (changes requested):
The overlay and its post-CERTS semantics match conwayLedgerTransitionTRC, and all eight divergence tests fail against the parent commit. One hot-path cost to address before merge.

wolf31o2 (approved):

